PR #55, opened 2018-12-04
Feature/marine profiles

aerorahul:
Process marine profiles and convert to IODA netcdf format.


src/marine/profile/profile2ioda.py
```diff
@@ -0,0 +1,260 @@
...
+            'insitu_temperature@ObsValue', 'f4', ('nlocs',))
+        tmps.units = 'degree_celcius'
+        tmps.description = ''
+
+        tmp_errs = ncid.createVariable(
+            'insitu_temperature@ObsError', 'f4', ('nlocs',))
+        tmp_errs.units = '(degree_celcius)^2'
+        tmp_errs.description = 'Standard deviation of observation error'
+
+        ncid.date_time = int(datetime.strftime(self.date, '%Y%m%d%H%M'))
```
guillaumevernieres: Something is fishy with the global attribute (getting date_time = -59312912). We need yyyymmddhh (no mn) as an integer, how about 
ncid.date_time = int(datetime.strftime(self.date, '%Y%m%d%H%M')[0:10]) <-- That does the job.

aerorahul:
@guillaumevernieres When one does `int(YYYYMMDDHHMM)`, it is treated as int32 and you get the fishy value.
If you infact do `np.int64(YYYYMMDDHHMM)`, you will get the integer `YYYYMMDDHHMM`.

So, when you truncate the string YYYYMMDDHHMM to [0:10], you can represent that integer as int32 and thus the seemingly correct value is returned.

This raises the question, what about observations that are going to contain higher frequency of sampling (minutes and seconds)?

There are 2 solutions for getting rid of fishy value; the 1st one you pointed out. The second is: `ncid.date_time = int(datetime.strftime(self.date, '%Y%m%d%H'))`

aerorahul:
addressed @guillaumevernieres changes.

src/marine/profile/profile2ioda.py
aerorahul: @guillaumevernieres When one does `int(YYYYMMDDHHMM)`, it is treated as `int32` and you get the fishy value.
If you infact do `np.int64(YYYYMMDDHHMM)`, you will get the integer`YYYYMMDDHHMM`.

So, when you truncate the string YYYYMMDDHHMM to `[0:10]`, you can represent that integer as int32 and thus the seemingly correct value is returned.

This raises the question, what about observations that are going to contain higher frequency of sampling (minutes and seconds)?

There are 2 solutions for getting rid of fishy value; the 1st one you pointed out. The second is: `ncid.date_time = int(datetime.strftime(self.date, '%Y%m%d%H'))`

aerorahul: addressed @guillaumevernieres changes.

guillaumevernieres:
That works. 
date_time is just the center time of the window, the obs time is a real (in hours) referenced to the center of the window.

PR #63, opened 2018-12-14
Feature/gsi many obs

srherbener:
This pull request adds in the netcdf writer that works with the met office odb converter. With this code, the obd converter will dump out a netcdf file in the current format (list of vectors, nlocs long, using the @ symbol to attach the group name, etc.)

The netcdf writer supports both the current and newly designed netcdf formats. There is a global variable in ioda_conv_ncio.py script called OUT_NC_VERSION that controls which format is used. Version 1 is the current format and version 2 is the newly designed format.


srherbener:
Hi Steve,

Thanks for the review! I must have gotten mixed up when I merged in develop and didn't notice the extra fetchone call when fixing conflicts from the merge. I'll take care of that now.

PR #65, opened 2018-12-19
Add the path to the ioda_conv_ncio module in code

svahl991:
Before this change, the file ioda_conv_ncio.py had to be copied into the same directory as odbapi2nc.py in order for the latter to be able to import and use it. Now odbapi2nc.py programmatically adds the path to ioda_conv_ncio.py using the relative path from the ioda-converters file structure.

### DIFF
--- a/src/met-office/odbapi2nc.py
+++ b/src/met-office/odbapi2nc.py
@@ -5,7 +5,6 @@ import sys
 import os
 import argparse
 from math import exp
-import ioda_conv_ncio as iconv
 
 #This function takes two integers like date=20180415 time=61532 and converts them to a
 #ISO 8601 standard date/time string like "2018-04-15T06:15:32Z"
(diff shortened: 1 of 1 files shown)


svahl991:
Thanks for this, Steve. I didn't realize we were doing the ecbuild/cmake process on this repo. I thought it was just a collection of python scripts.

However, I tried the build/make instructions you provided and I got an error when I tried the 'make install' command. I don't have permissions to /usr/local inside the Singularity image and it didn't fall back to $HOME/tools, even though I had specified the CMAKE_INSTALL_PATH in the ecbuild options.

```
Singularity:~/tools>make install
Install the project...
-- Install configuration: "Release"
-- Up-to-date: /usr/local/include/.
CMake Error at cmake_install.cmake:40 (file):
  file INSTALL cannot set permissions on "/usr/local/include/."


Makefile:93: recipe for target 'install' failed
make: *** [install] Error 1
```

srherbener:
There's a typo in my instructions. The variable is CMAKE_INSTALL_PREFIX instead of CMAKE_INSTALL_PATH.  Sorry about that.

svahl991:
Yes. It works now. This pull request and corresponding branch can be ignored and removed.

PR #73, opened 2019-01-30
Add script to convert MO ODB API to JSON

svahl991:


svahl991:
Hi Steve,

Yes, I agree with all your comments about the obvious code duplication and refactoring that needs to be done. You are also right that the code was just something I put together "quick and dirty" for my MongoDB experiments. The reason I put it in a pull request "as is" was because I felt the code was useful enough that we should keep it around in case we end up using MongoDB and/or JSON more and we want to clean it up later, but not useful enough (since I think the chance of us actually using MongoDB is low) that I wanted to spend the time to do the obvious refactoring needed.

However, in the time since I put in the pull request, I decided I'm going to work on my conversion script more now anyway to get it ready for new Obs Types, so I'll go ahead and start by making the suggested changes.

PR #75, opened 2019-02-04
feature/marineobs

srherbener:
This pull request replaces the one @guillaumevernieres submitted recently that had a corrupted history. The 219 commits have been replaced with just one commit with the intended updates. 

I deleted the old pull request and the associated feature/marineobs branch. I simply created a new feature/marineobs branch with backups of the two scripts (profile2ioda.py and altimeter2ioda.py) from the old (deleted) branch. The scripts should be identical to those in the original pull request. Please test that the two scripts have the expected functionality.

### DIFF
--- a/src/marine/altimeter/altimeter2ioda.py
+++ b/src/marine/altimeter/altimeter2ioda.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python
+
+import scipy.io.netcdf as netcdf
+import numpy as np
+import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


srherbener:
Please note that any local clones of this repository should be deleted and replaced with a fresh clone in order to prevent this type of corruption in the future. Thanks!

srherbener:
That's a great idea. We've got it on the to-do list to add tests, and it sounds like one of these checkers could be one of the unit tests (like the oops_coding_norms test).

PR #79, opened 2019-03-05
Feature/fnmoc

aerorahul:
add surface observations and ship track converters from GODAE.


src/marine/godae/trak2ioda.py
```diff
@@ -0,0 +1,206 @@
...
+        times.units = 'hours'
+        times.description = 'why does this have to be with a reference to, why not absolute?'
+
+        sal = ncid.createVariable('sea_surface_salinity@ObsValue', 'f4', ('nlocs',))
+        sal.units = 'g/kg'
+        sal.description = ''
+
+        sal_errs = ncid.createVariable(
+            'sea_surface_salinity@ObsError', 'f4', ('nlocs',))
+        sal_errs.units = '(g/kg)^2'
```
guillaumevernieres: I think we need the standard deviation of the obs error (g/kg)

src/marine/godae/trak2ioda.py
```diff
@@ -0,0 +1,206 @@
...
+        u_errs.description = 'Standard deviation of observation error'
+
+        v = ncid.createVariable(
+            'sea_surface_meridional_wind@ObsValue', 'f4', ('nlocs',))
+        v.units = 'm/s'
+        v.description = ''
+
+        v_errs = ncid.createVariable(
+            'sea_surface_meridional_wind@ObsError', 'f4', ('nlocs',))
+        v_errs.units = '(m/s)^2'
```
guillaumevernieres: same as above, std deviation instead of variance.

aerorahul:
I will update the readme.

src/marine/godae/trak2ioda.py
aerorahul: I simply followed what is in profile2ioda.py.
I also noticed that in profile2ioda.py, someone added a floor value of 0.5 to the temperature obs. err.
The ob err. in track and surface obs. is currently arbitrary and needs to be addressed at some point.
At some point we will standardize these to use the netCDF API from @srherbener for writing to IODA format.

travissluka:
should the appropriate CMakeLists.txt files be added so that the marine scripts get placed in the install directory like the other converters?

aerorahul:
@travissluka That is a good idea.  I will update the PR, and ask for a re-review.

src/marine/godae/ship2ioda.py
```diff
@@ -0,0 +1,171 @@
...
+        idata is the IODA required data structure
+        '''
+
+        # idata is the dictionary containing IODA friendly data structure
+        idata = self.odata
+
+        # Diff. between ob. time and time in the file (hrs)!
+        for i in range(self.odata['n_obs']):
+            dtg = datetime.strptime(self.odata['ob_dtg'][i], '%Y%m%d%H%M')
+            idata['ob_dtg'][i] = (dtg - self.date).seconds/3600.
```
travissluka: the wrong values are being produced, this should use **.total_seconds()** instead

src/marine/godae/trak2ioda.py
```diff
@@ -0,0 +1,206 @@
...
+        idata is the IODA required data structure
+        '''
+
+        # idata is the dictionary containing IODA friendly data structure
+        idata = self.odata
+
+        # Diff. between ob. time and time in the file (hrs)!
+        for i in range(self.odata['n_obs']):
+            dtg = datetime.strptime(self.odata['ob_dtg'][i], '%Y%m%d%H%M')
+            idata['ob_dtg'][i] = (dtg - self.date).seconds/3600.
```
travissluka: use .total_seconds()

src/marine/godae/ship2ioda.py
travissluka: same thing for profile2ioda.py as well

travissluka: ...noticeable if you pass a date into the script that is different from the base date of the obs file

src/marine/godae/profile2ioda.py
```diff
@@ -158,7 +158,7 @@ def _to_nlocs(self):
             dtg = datetime.strptime(self.odata['ob_dtg'][n], '%Y%m%d%H%M')
 
             # Diff. between ob. time and time in the file (hrs)!
-            ddtg = (dtg - self.date).seconds/3600.
+            ddtg = (dtg - self.date).total_seconds/3600.
```
travissluka: **total_seconds** is a function, needs **()**
same with other 2 locations

aerorahul: thanks for the catch.

src/marine/godae/trak2ioda.py
aerorahul: done.

src/marine/godae/ship2ioda.py
aerorahul: indeed.

src/marine/godae/trak2ioda.py
aerorahul: fixed.

aerorahul:
merged.

PR #108, opened 2019-03-28
bugfix/py2-py3-compatibility

srherbener:
This PR introduces changes to enable using either of python2 or python3.


aerorahul:
@srherbener I merged this branch into feature/concat and ran one of my converters with python2 and python3. I notice non-zero differences in the output netCDF file; in ObsValue and MetaData for time, latitude and longitude, that I cannot explain.

I will send you a test case separately.

srherbener:
@aerorahul, thanks for reviewing this. I saw differences between python2 and python3 as well so I checked three locations. It turned out that I got the same data, but they were in a different order. Eg. python2 put the data in locations 0, 1 and 2; and python3 put them in other locations. I'm guessing this happens due to the dictionaries getting built in different orders. I didn't check the whole file, but it looked like I was getting the same output, but with different ordering. I'm not sure I have an easy way to check the entire files to make sure they match.

aerorahul:
@srherbener That was my impression too.  It appears that python dictionary is handled differently in the two versions.  Here is what I found about [dict](https://portingguide.readthedocs.io/en/latest/dicts.html) from googling it.  It also may be helpful to use `OrderedDict` from `collections` in `ioda_conv_dict` instead of `{ }` 
That will keep the keys in the order they are populated, and is generally preferred.

srherbener:
Thanks for checking the files @travissluka! I like @aerorahul 's  suggestion to use OrderedDict to make the output more reproducable. I'll look into that.

srherbener:
@aerorahul and @travissluka, I changed the dictionaries generated by ExtractObs to OrderedDict, and tested using rads2ioda.py. I got the same result as before where the ordering is different between python2 and python3. I tried dumping out the keys in the ExtractObs routine and that suggested that the ObsData dictionary being passed into BuildNetcdf is in a different order. rads2ioda.py is using defaultdict which I'm guessing doesn't guarantee the dictionary key order.

How much trouble would it be to fix the scripts using NcWriter? I'd like to get this fixed, but I also don't want to get bogged down and interfere with the code sprint.

Thanks!

travissluka:
@srherbener After the code sprint, when we are cleaning up everything we've done before doing a PR into develop, we can change to  ordereddict. Since python2/3 both work now, I say we can close and merge this pr

aerorahul:
@srherbener I agree with @travissluka. Merge this PR now and do clean-up after. The converter works now, so lets move on.

PR #112, opened 2019-03-29
Feature/odb2 yaml

svahl991:
This feature branch contains nearly a complete re-write of the ODB2 conversion script.

The pyhon script is now designed to be general and to be able to be configured to work on any ODB2 report file created by the Met Office or ECMWF. The configuration is done via a YAML definition file that is passed to script. YAML definition files have been created and tested for converting Met Office radiosonde, aircraft, and amsu-a data and are included in the repository.

Also, some changes were made to the source code directory structure. The name of the "met-office" directory was no longer appropriate since the script should work with ECMWF reports as well.


src/ODB2/Definitions/MO_aircraft_16026_Definition.yaml
```diff
@@ -0,0 +1,99 @@
...
+# Format: <key: value> = <variable name: vertco_type value>
+vertco_variables:
+    air_pressure: 11
+
+# varno_variables are read from the obsvalue, obs_error, and
+# QC columns when the varno column has the specified value.
+#
+# Format: <key: value> = <varno value: variable name>
+varno_variables:
+    2: temperature
```
srherbener: We've been using the CF name "air_temperature" for atmospheric sensible temperature. The website that I've been using (http://cfconventions.org/Data/cf-standard-names/64/build/cf-standard-name-table.html) doesn't list "temperature" as a CF name.

I would like to change this back to "air_temperature" here and in the other YAML files. However, I realize that you might have received a request to change it, and if that's the case we should discuss this with whoever made the request.

Thanks!

svahl991: That was totally an oversight on my part when I created the YAML. Good catch. "temperature has now been changed to "air_temperature" in all the relevant YAML files.

PR #115, opened 2019-03-29
EMC ice fraction converter

guillaumevernieres:
Converts l2 ice fraction to ioda


travissluka:
actually... emcice directory needs to be added to the marine/CMakeLists.txt

PR #120, opened 2019-04-01
Change var name to match ufo

srherbener:
Hi @svahl991,

I think our paths crossed this morning. I merged in your pull request (#112) while you added these changes to the feature/odb2-yaml branch. The merge didn't get these changes and I'm guessing you need them so I created this pull request on your behalf. If I'm wrong, and you don't need these changes I'll close this pull request. Otherwise, I'll merge it in.


svahl991:
I noticed that you had already merged the feature branch after I had pushed my changes to it. So then I created a new bugfix branch for the same change before I noticed that you had created this pull request. So our paths crossed twice!

I guess you can pick which PR you want to merge with develop, and delete the other one. They should be identical.

PR #125, opened 2019-04-04
Thinning of emc ice fraction

guillaumevernieres:
Added thinning to emc ice fraction.

### DIFF
--- a/src/marine/emcice/emcice2ioda.py
+++ b/src/marine/emcice/emcice2ioda.py
@@ -7,6 +7,7 @@ import netCDF4 as nc
 from datetime import datetime, timedelta
 import dateutil.parser
 from collections import defaultdict
+import numpy as np
 
 
 class Observation(object):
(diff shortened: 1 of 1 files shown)


src/marine/emcice/emcice2ioda.py
```diff
@@ -72,9 +78,12 @@ def _read(self):
                         type=str, required=True)
     parser.add_argument('-d', '--date',
                         help="base date", type=str, required=True)
```
aerorahul: add metavar="YYYYMMDDHH", to explicitly make sure the user provides that.
I have been bitten in the past.

src/marine/emcice/emcice2ioda.py
```diff
@@ -72,9 +78,12 @@ def _read(self):
...
+    parser.add_argument('-t', '--thin',
+                        help="amount of random thinning, from 0.0 to 1.0",
+                        type=float, default=0.0)
```
aerorahul: Provide a valid range, e.g. choices=range(0,101), metavar="[0-100]"

travissluka:
I'll deal with Rahul's suggestions in the cleanup for develop

PR #126, opened 2019-04-04
Feature/marine cs sst net cdf

ShastriPaturi:
1. I have the code in Fortran to extract the NESDIS Satellite SST data for 5 satellites with time window of +/1 12 hours from analysis time, concatenates the files and writes the output in IODA netcdf format.
It has a couple preQC filters already:
 a. Writes only quality flag =5 SST data.
 b. Writes only SST data excluding snow & ice regions.

2. The make file is for theia - very generic, two lines.
3. The shell script grabs the data and writes it into a separate file - filenclist.
USAGE: sh IODA_SST_prep.sh <analysis_time> 12 (time window in hrs)


aerorahul:
apologies for the accidental closing of the ticket. I meant to simply comment.

src/ncep/SST-NetCDF/make_theia.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+module load netcdf
+
+ifort -c  read_nc_combine_sst_IODA.f  -I/apps/w3lib/2.0.2-intel/lib `nc-config --fflags --flibs --libs`
```
aerorahul: Including -I/apps/w3lib/2.0.2-intel/lib a library? 
And Linking with a different actual library -L/scratch3/NCEPDEV/nwprod/lib

Issue #119, opened 2019-04-01
Rewrite GSI ncdiag -> IODA/UFO converter scripts


CoryMartin-NOAA:
Take Anna's existing scripts to convert GSI diag files to IODA/UFO but rewrite them as part of a Python module so that each obstype can be called concurrently/independently. As more satellites become available for example, it is better on Theia to use Python multiprocessing to convert each one simultaneously rather than serially. Additionally will allow for flexibility to call from command line or another python script.

Issue #107, opened 2019-03-27
Fix GSI ncdiag -> ioda/ufo converter scripts to use NcWriter


shlyaeva:
Currently GSI ncdiag -> ioda/ufo obs/geovals files converter scripts do not use NcWriter. 
It would be beneficial to use NcWriter (in that case if the ioda file format changes we won't have to change the scripts).

ADCollard:
OK, help me out here.   What is NCWriter and where do I find it?

I might re-assign to Cory!

ADC

CoryMartin-NOAA:
I believe it's part of the IODA-converters python library, which I guess since I commented means I'm volunteering myself? haha

ADCollard:
Yes, well volunteered Cory.

CoryMartin-NOAA:
@srherbener 

I'm working on these modified Python scripts to generate IODA observation files from GSI diag files, and while I am able to get the ObsValue, PrepQC, and ObsError to write out successfully, I have not yet figured out how to output additional information that is used for the UFO tests (GSIHofX for example). As far as I know the code runs with no errors but the extra variables are not there when I run ncdump.

Does the NcWriter class not currently support this, or is there an extra step I need to do to get it to save these extra data?

If you'd like to look at my code so far, it's in ioda-converters/feature/gsi_ncwriter:
https://github.com/JCSDA/ioda-converters/blob/feature/gsi_ncwriter/src/gsi-ncdiag/conv.py#L119-L123

Thanks!

CoryMartin-NOAA:
Never mind, I think I've figured out what I need to modify in ioda_conv_ncio to get this to work more generally. I'll test it and see if it matches what the original scripts produce.

srherbener:
Hi @CoryMartin-NOAA, Sorry about the slow response. The NcWriter class in the develop branch needs a simple modification to accommodate additional obs data such as the H(x) values from GSI. Lines 345-350 in ioda_conv_ncio.py are, unfortunately, redundant and they can be replaced with:
* ObsVars[VarKey[1]][VarNum, LocNum-1] = VarVal
Once that is done, you should be able to add ObsValue, ObsError, PreQC as before to the ObsData dictionary, and then add in the other values. Sorry about this code fault, I think it evolved from a state where the if statements were actually doing something and I didn't notice that the if statements became redundant.

srherbener:
[...]

PR #128, opened 2019-04-09
Feature/gsi ncwriter

CoryMartin-NOAA:
#107 #119 

This looks like a lot but most of the changes are moving/adding/renaming/deleting some files and replacing them with this new structure and putting most of the old files in old/.

I've tested and I know they run for all obs types on Theia (well at least they did) but I have not confirmed 100% that all of the output observations/GeoVaLs are correct (but a cursory look shows nothing out of the ordinary)


travissluka:
I'm about to do a big PR from the feature/marine-cs-03-2019 branch. There is something about the  changes made to ioda_conv_ncio.py that break break my converters, I'll look into it... 

Also, delete commented out lines of code.

CoryMartin-NOAA:
@travissluka okay then it's probably a good idea for me to wait and see what is pulled in from the code sprint and adjust accordingly. And yes, I forgot about the wall of commented lines in conv_dicts.py, I'll address that after feature/marine-cs-03-2019 is merged.

srherbener:
@CoryMartin-NOAA and @travissluka, my preference is for you to attempt to resolve the issue with this pull request that Travis raised now so that this pull request doesn't get caught up in a series of delays. Of course, if the fix is going to take a significant amount of time it might make sense to wait until after the marine code sprint pull request is merged so I'm absolutely open to taking that tactic. I just want to keep things streamlined as much as possible. Thanks!

CoryMartin-NOAA:
Fixed the commented out lines in conv_dicts.py. I merged with develop (sorry forgot to do that!) and the only conflicts were in ioda_conv_ncio but I left the old lines still in there commented out so that @travissluka and I can figure out how to get a version of the module that works for everyone.

travissluka:
@CoryMartin-NOAA  I've issued PR #130 for the marine converters, which should be straightforward to get merged since 95% of it has already been approved in PRs during the code sprint... once that's in I can get you a test file to make sure the marine converters still work while you track down the issue with your changes to ioda_conv_ncio.py (it has something to do with the way ObsVarList is created and ObsVars is initialized, PreQc fields are lost somewhere in the process)

srherbener:
I'm resetting my review. Let's go ahead and get the marine code sprint pull request in first. Then get the test case to Cory so he can debug. Thanks!

CoryMartin-NOAA:
@travissluka let me know when you have a sample file for me to test this out as I've merged develop into this branch. @srherbener performance is very slow on some of the satellites when using NcWriter (almost an hour for IASI), I'm not sure if it's due to NcWriter itself or the way I'm looping through the channels in toIODAobs() in radiances.py.

travissluka:
@CoryMartin-NOAA 
use the `marine/rads_adt2ioda.py` converter and grab any file from ftp://ftp.star.nesdis.noaa.gov/pub/sod/lsa/rads/adt

srherbener:
@CoryMartin-NOAA, I'm not surprised that the performance is poor for IASI. I think that one factor is the large tuple for the LocKey of which the refactoring I am proposing should help with. Another factor could be having a large number of locations where the obs data dictionary will have that same large number of keys. If this is correct, the refactoring should help with this as well. 

I think it would be good to get this pull request merged in first, then do the NcWriter refactoring. The refactoring will require some changes in your scripts but it should be small changes. Does this sound okay?

srherbener:
I'm adding Steve Vahl as a reviewer since I want to make sure that the modifications done in ioda_conv_ncio.py are okay with his ODB2 converter.

svahl991:
>     NumpyDtype = ObsVarExamples[o].dtype
> AttributeError: 'float' object has no attribute 'dtype'
> ```
> I think the issue is that dtype is an attribute on a numpy array, but the object at line 352 in ExtractObsData is a list (not a numpy array).

This is the error I'm seeing with the ODB2 converter script as well.

CoryMartin-NOAA:
@svahl991 @srherbener  ok let me modify this to be more generic (does it work to force the list to be a numpy array just for the check?)

srherbener:
@CoryMartin-NOAA, I think you could use the type() function:
~~~~~~~~
VarType = type(a)
if (VarType is int):
...
elif (VarType is float):
...
elif (VarType is str):
~~~~~~~~
seems a lot cleaner than converting to a numpy array just to do the check.

Hope this helps!

CoryMartin-NOAA:
@srherbener yes your plan sounds good and I can change these converters once NcWriter is reconfigured. I tried it with converting to numpy and it seems to work for my test case as well as the file that @travissluka suggested. Let me try it with type instead of numpy type because I agree it's probably best to not use Numpy unless needed.

srherbener:
Sorry, I didn't noticed the commit you just made until right now. I tried this on the rads_adt2ioda.py command and it works now.

CoryMartin-NOAA:
@srherbener I changed it slightly again to use either type or numpy type because they are not the same and type returns numpy.float32, etc. and thus I have to do if VarType in [float,numpy.float32,numpy.float64] to get it to work. However, @travissluka 's original concern I believe was with the output values of PreQC. I just tested a file and they were all the same value, not sure if this is correct or not.

CoryMartin-NOAA:
Nevermind, I see in the script that it explicitly assigns them all the value of 0.1 for error and 0 for QC so I think it's working properly now.

srherbener:
rads_adt2ioda.py is working with the latest commit. @svahl991, has this fixed the odb converter? Thanks!

srherbener:
@travissluka, if it isn't too much trouble, could you try @CoryMartin-NOAA's latest commit on some more of the marine scripts (other than rads_adt2ioda.py). Thanks!

svahl991:
> rads_adt2ioda.py is working with the latest commit. @svahl991, has this fixed the odb converter? Thanks!

Yep, the ODB converter is working again.

src/lib-python/ioda_conv_ncio.py
```diff
@@ -328,11 +347,19 @@ def ExtractObsData(self, ObsData):
...
-            self._oqc_name   : np.full((self._nvars, self._nlocs), self._defaultI4),
-            }
+	ObsVars = {}
+	for o in range(len(ObsVarList)):
+            VarType = type(ObsVarExamples[o])
+            if (VarType in [float,np.float32,np.float64]):
+                defaultval = self._defaultF4
+            elif (VarType in [int,np.int64,np.int32]):
+                defaultval = self._defaultI4    # convert long to int
+            elif (VarType in [str,np.s1]):
```
travissluka: np.s1 does not exist

travissluka:
```suggestion
            elif (VarType in [str,np.dtype('S1')]):
```

src/lib-python/ioda_conv_ncio.py
```diff
@@ -328,11 +347,19 @@ def ExtractObsData(self, ObsData):
...
-            self._oval_name  : np.full((self._nvars, self._nlocs), self._defaultF4),
-            self._oerr_name  : np.full((self._nvars, self._nlocs), self._defaultF4),
-            self._oqc_name   : np.full((self._nvars, self._nlocs), self._defaultI4),
-            }
+	ObsVars = {}
+	for o in range(len(ObsVarList)):
+            VarType = type(ObsVarExamples[o])
+            if (VarType in [float,np.float32,np.float64]):
+                defaultval = self._defaultF4
+            elif (VarType in [int,np.int64,np.int32]):
```
travissluka:
```suggestion
            elif (VarType in [int,np.int64,np.int32,np.int8]):
```

src/lib-python/ioda_conv_ncio.py
```diff
@@ -381,12 +428,13 @@ def ExtractObsData(self, ObsData):
...
-                        ObsVars[self._oerr_name][VarNum, LocNum-1] = VarVal
-                    elif (VarKey[1] == self._oqc_name):
-                        ObsVars[self._oqc_name][VarNum, LocNum-1] = VarVal
+                    #if (VarKey[1] == self._oval_name):
+                    #    ObsVars[self._oval_name][VarNum, LocNum-1] = VarVal
+                    #elif (VarKey[1] == self._oerr_name):
+                    #    ObsVars[self._oerr_name][VarNum, LocNum-1] = VarVal
+                    #elif (VarKey[1] == self._oqc_name):
+                    #    ObsVars[self._oqc_name][VarNum, LocNum-1] = VarVal
+		    ObsVars[VarKey[1]][VarNum, LocNum-1] = VarVal
```
travissluka: remove commented out code

src/lib-python/ioda_conv_ncio.py
```diff
@@ -311,15 +311,34 @@ def ExtractObsData(self, ObsData):
 
         # Walk through the structure and get counts so arrays
         # can be preallocated, and variable numbers can be assigned
+	ObsVarList = []
+        ObsVarExamples = []
+        VMName = []
+        VMExamples = []
```
travissluka: this var is not used, remove it

src/lib-python/ioda_conv_ncio.py
CoryMartin-NOAA: yeah sorry it should be np.str_ I didn't realize this one didn't follow the convention of the others

CoryMartin-NOAA:
@travissluka the error I got when running marine/hgodas_insitu2ioda.py with one of the files from your tarfile suggests that it's an issue with the variable names and not the changes to this file. Can you verify what error you were getting? I saw :     
...
time = ncd.variables['time'][:]
KeyError: 'time'

travissluka:
haha, oops, i gave you the wrong file. should have been
https://www.atmos.umd.edu/~tsluka/tmp/insitu.tgz

regardless, the latest changes seem to fix everything

srherbener:
I'm going to merge now. Thanks everyone for your hard work in getting all of these new scripts merged in this week!

PR #131, opened 2019-04-11
Update repo Readme file

svahl991:
After seeing that Anna had the same trouble running one of the scripts in this repo that I had a while back, I thought it would be good to add some clarification on how to install the scripts to the Readme file.

I also updated the Readme section regarding the odbapi2nc.py script, which should have been done in the same pull request that added the new ECMWF YAML files.


srherbener:
This is a pretty innocuous change since it's only the README file. I'm going to go ahead and merge despite not having two approving reviews.

PR #132, opened 2019-04-15
Feature/gsi ncwriter

CoryMartin-NOAA:
Found a few small issues in the new conversion scripts:

1 - LocKeys need to be within channel loop or else it does not write out the proper number of obs for radiances 
     - previous "fix" for this had the correct # of obs but the incorrect values/locations when written to the file. Output files look good now.

2 - changed the output filename to "geoval" (no s) and because of a NF90_INT issue in the UFO test, forced it to read the input GSI diag value for date_time to the output geoval file

I'll wait for any further changes/additions until NcWriter is rewritten.

### DIFF
--- a/src/gsi-ncdiag/composition.py
+++ b/src/gsi-ncdiag/composition.py
@@ -52,7 +52,7 @@ class AOD:
     import netCDF4 as nc
 
     # set up output file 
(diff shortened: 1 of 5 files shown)


srherbener:
@CoryMartin-NOAA, when you say you are waiting for NcWriter to be rewritten for further changes, do you mean further changes that will appear in subsequent pull requests? Thanks!

CoryMartin-NOAA:
@srherbener yes I meant for anything beyond what these changes are.

PR #134, opened 2019-04-17
Add option to copy bias-corrected obs value

svahl991:
We discovered that Met Office ODB2 reports contain a column called 'corvalue' which sometimes contains a bias-corrected observation value. This PR adds an option to the odbapi2nc import script that will allow users to import this bias-corrected value ('corvalue' column) instead of the un-corrected obs value ('obsvalue' column) if desired.

Also, while working on this enhancement, a problem was discovered with the ODB2 directory CMakeLists.txt file that was causing the ODB2 scripts to not get installed. (Not sure when this got broken since it used to work.) This is also fixed.

### DIFF
--- a/src/ODB2/CMakeLists.txt
+++ b/src/ODB2/CMakeLists.txt
@@ -1,8 +1,7 @@
 list( APPEND programs
   odbapi2nc.py
   odbapi2json.py
-  var_convert.py
-  DESTINATION bin)
+  var_convert.py)
 
 COPY_FILES( "${programs}" ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_BINARY_DIR}/bin )
 
--- a/src/ODB2/CMakeLists.txt
+++ b/src/ODB2/CMakeLists.txt
@@ -4,6 +4,14 @@ list( APPEND programs
   var_convert.py
   DESTINATION bin)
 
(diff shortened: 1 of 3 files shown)


src/ODB2/CMakeLists.txt
```diff
@@ -4,6 +4,14 @@ list( APPEND programs
   var_convert.py
   DESTINATION bin)
```
srherbener: Should get rid of the "DESTINATION bin" keywords, since they are extraneous.

svahl991: @srherbener , maybe I misunderstood, but when I tried removing the "DESTINATION bin" line I got an error when I executed the "ecbuild" command for the build directory. Restoring the removed line caused the error to go away. And all the other comparable "CMakeLists.txt" files in the repo also have the "DESTINATION bin" line in that place. Are you sure these words are extraneous?

srherbener: Sorry, I meant only on the list ( APPEND programs .... ) command, you need "DESTINATION bin" on the install (...) command. Check out the CMakeLists.txt file in src/marine for example.

svahl991: Ah. I did misunderstand. Sorry about that. I have committed the requested change now.

PR #138, opened 2019-04-18
Feature/ctests marine

travissluka:
ctests using nccmp are added for the marine ioda-converters. To be used as a template for other people to add their own ctests. nccmp is assumed to already be available on the system.

I've shrunk the marine input files down a lot, each file is on the order of 10-50Kb

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 *.nc filter=lfs diff=lfs merge=lfs -text
 *.nc4 filter=lfs diff=lfs merge=lfs -text
+*.bin filter=lfs diff=lfs merge=lfs -text
 src/ncep/xx120 filter=lfs diff=lfs merge=lfs -text
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,7 @@ include_directories( ${IODACONV_INCLUDE_DIRS} ${IODACONV_EXTRA_INCLUDE_DIRS} )
 include( iodaconv_extra_macros )
 add_subdirectory( src )
 add_subdirectory( tools )
+add_subdirectory( test )
 
 if(ECBUILD_INSTALL_FORTRAN_MODULES)
   install(DIRECTORY ${CMAKE_Fortran_MODULE_DIRECTORY}/${CMAKE_CFG_INTDIR} DESTINATION ${INSTALL_INCLUDE_DIR} )
--- a/src/lib-python/ioda_conv_ncio.py
+++ b/src/lib-python/ioda_conv_ncio.py
@@ -83,8 +83,8 @@ class NcWriter(object):
         self._ndatetime = 20
 
         # default fill values
-        self._defaultF4 = np.float32(np.abs(netCDF4.default_fillvals['f4']))
-        self._defaultI4 = np.int32(np.abs(netCDF4.default_fillvals['i4']))
+        self._defaultF4 = np.abs(netCDF4.default_fillvals['f4'])
+        self._defaultI4 = np.abs(netCDF4.default_fillvals['i4'])
 
         # Names assigned to record number (location metadata)
         self._rec_num_name = "record_number"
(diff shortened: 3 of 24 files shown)


CoryMartin-NOAA:
It's going to take me a minute to get nccmp installed on Theia to test this... they block access to gitlab for a direct clone.

CoryMartin-NOAA:
@travissluka 2 things: 9 of the tests fail for me. Some of these are because I don't have the pycodingnorms on my path, where do I find this? Second, 4 of the tests for the converters fail, I tried one and it seems to be a float vs int issue?:
8: DIFFER : VARIABLE : sea_surface_salinity@PreQc : TYPE : FLOAT <> INT
1/1 Test #8: test_iodaconv_smap_sss ...........***Failed    0.27 sec

srherbener:
4 out of 5 of the coding norms tests are currently expected to fail. We need help from the script writers to get these fixed. I've fixed the lib-python coding norm test in my feature/ncwriter-refact branch.

I'm seeing all of the marine convert tests pass. I'm running outside the container on my iMac. Not sure why Theia has this error come up, when it works on the mac.

travissluka:
@CoryMartin-NOAA it looks like you have a stale ioda_conv_ncio.py, you did the change to int32 a couple days ago. Make sure you're running "make install"

@aerorahul setup the coding norms, he might know what it takes to get it working on theia

CoryMartin-NOAA:
@travissluka the change I made was for the metadata in the netCDF file (things like nlocs, etc.) but this is the PreQc which is float in the file I generate and int (probably the correct type) in the file it's comparing to.

The input files are type "byte" for QC, and I wonder if this is not being handled correctly to int but instead float? It is still possible I have a file/git issue but I can't see how since I checked out your branch and created a new build directory.

CoryMartin-NOAA:
https://github.com/JCSDA/ioda-converters/blob/65600b054a23efd85ecd5aa9ac7c9cdd310c1977/src/lib-python/ioda_conv_ncio.py#L245
When I print Vvals.dtype here the type is float64 but in the converter script you wrote as well as the part of the code I modified, the type is int8. I don't yet know why this is the case.

travissluka:
@CoryMartin-NOAA hmm, i wonder if the version of python/netCDF4 on Theia is handling things differently than the version steve and I are using. I'll poke around some more on my end to see if I can find anything that makes sense

CoryMartin-NOAA:
@travissluka I've found the problem:
https://github.com/JCSDA/ioda-converters/blob/65600b054a23efd85ecd5aa9ac7c9cdd310c1977/src/lib-python/ioda_conv_ncio.py#L360

np.full has an option for dtype, otherwise it uses the value of fill_value which is defaultval which comes from the netCDF default fill value. It's using both as floats instead of integers for the integer values. 

@srherbener in ioda_conv_ncio, I think the default for F4 and I4 either need to be explicitly made as float32 and int32 or do it in the np.full line with logic similar to how the default fill value is chosen in the if statements.

CoryMartin-NOAA:
The test passes if I change https://github.com/JCSDA/ioda-converters/blob/65600b054a23efd85ecd5aa9ac7c9cdd310c1977/src/lib-python/ioda_conv_ncio.py#L87 to be
self._defaultI4 = np.int32(np.abs(netCDF4.default_fillvals['i4']))

srherbener:
Thanks for debugging. I'll fix this. Amazing that they use a float value for a default integer value.

CoryMartin-NOAA:
I spoke too soon, this didn't fix it. Instead of the np.int32 I guess it was the explicit np.fill type that worked correctly. Sorry, been a long couple of days and my brain is fried.

CoryMartin-NOAA:
@srherbener @travissluka All of the tests except for the coding norms now pass for me with the change I just pushed.

travissluka:
looks good on my end as well

srherbener:
Tests still pass for me too. I'll merge now

aerorahul:
> @CoryMartin-NOAA it looks like you have a stale ioda_conv_ncio.py, you did the change to int32 a couple days ago. Make sure you're running "make install"
> 
> @aerorahul setup the coding norms, he might know what it takes to get it working on theia

@CoryMartin-NOAA Catching up on all the github chatter from yesterday. For running the coding norms (python) you will need pycodestyle. It can be installed via conda or pip.  You could ask Ben.Blake to do it for you.  You may also ask for autopep8, to do automatic fixing of coding norm issues, so you don't have to do it manually line-by-line.

CoryMartin-NOAA:
Thanks @aerorahul ! I installed pycodestyle manually yesterday afternoon but I suspect it'll be better for others to use if it's installed as part of the conda installation most people use on Theia. I'll ask Ben.

PR #142, opened 2019-04-19
Feature/ncwriter refact

srherbener:
This pull request introduces a refactoring of the BuildNetcdf method in the NcWriter class. The ExtractObsData call inside BuildNetcdf has been pulled outside of BuildNetcdf. This enables the client script to do two things:
* append to the meta data dictionaries that ExtractObsData builds
* skip the ExtractObsData step if you have an obs type with its data already organized by locations
    * For example, radiance data that is organized as vectors (nlocs) or arrays (nchans X nlocs) in the input file

The original intent of the BuildNetcdf method was to process obs types that present their data as a stream of points such as radiosonde. The ExtractObsData call is used to reorganize the stream of points into vectors (one for each variable) that are nlocs long. For radiance data this was wasteful since the input data already has the obs data organized as vectors or arrays. Before the refactoring in this pull request, for radiance data the arrays or vectors would be pulled apart into a stream of points and then reorganized back into vectors.

You may want to revisit your scripts and take advantage of the "skip ExtractObsData step" if appropriate. I did this for the GSI (AOD in composition.py and radiances.py) as examples. 

I got all the marine tests to pass, and verified that the ODB2 and GSI scripts are working with test cases that I have. There are several marine scripts that remain untested, but I made the changes in these to accommodate the NcWriter refactoring (so hopefully, these scripts will at least run to completion).

Please test these changes with more of your test cases since I did a minimal job of testing.


srherbener:
@travissluka, I'm seeing the coding norms test pass on my iMac. What are the errors you are seeing?

travissluka:
that's odd, I they were failing, then I go home and now they're passing. Obviously the weekend needs to start

srherbener:
I second the motion to start the weekend!

I'm going to merge this now.

PR #143, opened 2019-04-19
bugfix/build-bin-install

srherbener:
This pull request fixes the issue where after editing a script it would not get update in the bin directory of the build area. This issue caused the older version of the script to be run during ctest.


travissluka:
non-executable python scripts are now being made executable when copied into the bin directory.

Also, it would be nice if support libraries (aod_dicts, radiances.py, etc) were placed in a separate directory within bin, as bin is starting to get a bit chaotic.

CoryMartin-NOAA:
@travissluka I agree and was already thinking about moving the GSI_Diag files that generate .pyc into a subdirectory.

aerorahul:
@srherbener @travissluka Rather than chmod-ing every script, let the developer commit the script with the correct permissions;  only those expected to run as executable need to be so, not ones under lib-python; for e.g. A simple copy then should suffice.
I agree with the need for copying lib in lib, and not under bin. The scripts then need to be updated to append the path to lib.

It may also be time to start thinking of an "setup.py" type approach customized for our needs.

srherbener:
@aerorahul, where is the COPY_FILES macro defined? I'm not finding it.

I ran through to this solution too quickly, sorry about that. I was trying to get a quick fix in before more people started getting tripped up.

I agree with all of the feedback. There should be a macro so we don't repeat code, not all scripts need execute permissions, scripts shouldn't all be in one directory, etc.

It gets problematic with paths if the scripts that are being imported are located in directories other than where the importing script is located. Is this why you are suggesting a setup.py approach @aerorahul?

Would it be acceptable for now to:
* create a macro for copying files
* get rid of the chmod +x
* place everything in the bin directory

Thanks!

srherbener:
I found the COPY_FILES macro.

aerorahul:
@srherbener the COPY_FILES macro is in [cmake/iodaconv_extra_macros.cmake ](https://github.com/JCSDA/ioda-converters/blob/bugfix/build-bin-install/cmake/iodaconv_extra_macros.cmake)

One can use `sys.path.insert` in the python script that has dependencies in locations other than ./ and `configure_file` in cmake to dynamically generate the final python script.  We would not have to use `setup.py` in that case.

I can create a demo branch with a few of the marine operators and test out after this work is integrated.

srherbener:
@aerorahul, I've got the macro built. Please review again.

 I decided to create a new macro since I thought it might be handy to keep the COPY_FILES macro. I got rid of the chmod +x, and I'm still placing everything in the bin directory for now. We can move over to your scheme when you get the demo branch working.

Thanks!

srherbener:
Sorry, I've got to merge in the other PR that I just merged into develop.

srherbener:
I'm seeing all marine script tests pass, and the marine and lib-python coding norms pass.

I'm going to go ahead and merge.

PR #147, opened 2019-04-22
Add tests for GSI Diag conversion and update to pass coding norms

CoryMartin-NOAA:
- GSI ncdiag conversion scripts have been updated to pass the PEP8 coding norms tests.

- src/gsi-ncdiag/old has been removed as these scripts should no longer be needed and they make the coding norm tests fail.
- Two tests added to check if obs are created properly from sample subsetted GSI diag files for surface T and a sample brightness temperature sensor. (NOTE: GeoVaLs are produced in the test but I am not comparing them currently to anything, nor am I processing AOD or Ozone diag files to 1) save space and 2) add extra tests, if you think they should be included, let me know)


shlyaeva:
@CoryMartin-NOAA: thank you for adding those! 
Cory and all: I won't be able to review this promptly, I won't have time until next Wednesday.

src/gsi-ncdiag/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ list( APPEND programs
   aod_dicts.py
   oz_dicts.py
   proc_gsi_ncdiag.py 
+  test_gsidiag.py
```
travissluka: this file is not in the repo

CoryMartin-NOAA: Oops. Done.

src/gsi-ncdiag/radiances.py
```diff
@@ -1,222 +1,245 @@
...
+            value = "brightness_temperature_{:d}".format(chanlist[c])
+            var_names.append(value)
+            print(self.obstype, value)
+            idx = chan_indx == chanlist[c]
+            if (np.sum(idx) == 0):
+                print("No matching observations for:")
+                print(value)
+                continue
+            obsdatasub = obsdata[idx]
+            obsdatasub[obsdatasub>1e6] = np.nan
```
jjguerrette: Presently the large IODA NC obs files for amsua have BT values of exactly 1e6.  Will this catch those?

CoryMartin-NOAA: No, I've changed it to be 9e5 as that should still be sufficiently large.

jjguerrette: Thank you Cory.

CoryMartin-NOAA: No problem, thanks for double checking this.

CoryMartin-NOAA: @jjguerrette using np.nan does not work with the ctest, as it can't properly handle NaN or Inf. I'm using the default netCDF fill value now of 9.96921e+36, hopefully this is fine.

jjguerrette: That should work since @svahl991 told me that any ObsValue larger than some predefined value (1e+10 maybe, please give the correct value) is treated as missing in IODA.  I assume that is why we already have values of 1e+11 in the radiance files.

svahl991: I just did some grepping and confirmed that the threshold value is 1e+8, which is what I thought. So any value larger than that is treated as a missing value. One thing to note, though, is that this threshold value is defined in the NetcdfIO source code. What this means is that it's defined in the code that reads the IODA data from IODA netCDF files. Right now we only use netCDF files to store IODA data so the distinction isn't important, but if we ever used a different storage medium (like ODC), we would use a different I/O class, and there's no mechanism to enforce that the same value is used by other I/O classes, or even that any threshold is set at all. We should probably set this threshold at a higher, public, level, so that it's not quite so hidden, and enforce somehow that all IODA I/O classes use the same value. I think I'll enter a Zenhub issue regarding this.

travissluka:
@CoryMartin-NOAA can you move `orddicts.py` to the `lib-python` dir, I'll be using it for the marine converters to clear up the python2/python3 ordering issues that are making my ctests fail on Theia

junmeiban:
Hi @CoryMartin-NOAA ,  

I cannot get u and v obs  when I use your branch to convert  the file “diag_conv_uv_ges.2018041500_ensmean.nc4” in http://data.jcsda.org/gsi_diags_2018041500.tar. Is this file ok to use? (I try to check this data use ncdump, but it shows empty even though the file size is almost 3G)

Many thanks,

Junmei

CoryMartin-NOAA:
@junmeiban  have you tried it recently? Since I initiated this pull request I added a new commit https://github.com/JCSDA/ioda-converters/pull/147/commits/75d9ce245c535af7a62e29733373f1db2415f7b1 which now I've tested and it works for at least aircraft and sondes. Previously it wasn't working because the GSI files have u_Observation and v_Observation and not Observation like the others do.

junmeiban:
@CoryMartin-NOAA ,I just try it today. I will double check the run procedure. Many thanks.
Here is my output information when I convert diag_conv_uv_ges.2018041500_ensmean.nc4:
('idx=', array([], dtype=bool))
No matching observations for:
Platform:aircraft Var:uv
('idx=', array([], dtype=bool))
No matching observations for:
Platform:sondes Var:uv
('idx=', array([], dtype=bool))
No matching observations for:
Platform:satwind Var:uv
('idx=', array([], dtype=bool))
No matching observations for:
Platform:vadwind Var:uv
('idx=', array([], dtype=bool))
No matching observations for:
Platform:windprof Var:uv
('idx=', array([], dtype=bool))
No matching observations for:
Platform:sfcship Var:uv
('idx=', array([], dtype=bool))
No matching observations for:
Platform:sfc Var:uv
('idx=', array([], dtype=bool))
No matching observations for:
Platform:scatwind Var:uv

shlyaeva:
@junmeiban, @CoryMartin-NOAA, might be a problem with the file. I'll double-check and try to run the scripts on theia.

shlyaeva:
It is the problem with the diag files. I think this might be an issue with nc_diag_cat (GSI utility to cat multiple diag files). I am working on this.

junmeiban:
@shlyaeva Many thanks!

shlyaeva:
@junmeiban: I've regenerated GSI diags and uploaded to the same place; please download again and try with this branch.

junmeiban:
@shlyaeva, @CoryMartin-NOAA,
 I can get geovals for the variable u and v successfully using your new GSI diags,
but I cannot get ioda obs. It failed to “import ioda_conv_ncio as iconv”. I am checking my setting now.

CoryMartin-NOAA:
@junmeiban are you running it in the build/bin directory? Cmake will copy the library ioda_conv_ncio from lib-python to the bin directory for Python to import it.

junmeiban:
@CoryMartin-NOAA,
I am not running it in the build/bin directory. Many thanks for reminding me. Now it works. I already got sondes and aircraft uv files.
It is still running… it looks that writing satwind to output is time-consuming (I use 128 processors).

BTW, I got a warning,  I think it’s okay, do you agree?
ioda-converters/build/bin/conv.py:213: RuntimeWarning: overflow encountered in divide

CoryMartin-NOAA:
@junmeiban  it probably won't work with more processors than you have on one node (not MPI only python subprocesses are called.
Many of the files, particularly satellites with many channels, do take some time. Probably a limitation somewhat of using Python vs a compiled language.

And yes that warning is okay, it just means you divided by zero somewhere, probably for one of the observation errors.

junmeiban:
@CoryMartin-NOAA Many thanks for your explaining. I just got satellite wind files.

junmeiban:
@shlyaeva , I have a question:
I checked the file: sondes_tsen_obs_2018041500.nc4 
Here is sensible T vertical distribution. Sensible T is mainly above 400hpa, is it normal? Many thanks.
pro_sondes_tsen_obs_2018041500_air_temperature

CoryMartin-NOAA:
@srherbener that test passes for me on Theia, but I see the infinity values in ncdump for the test run and not the missing values. This must be a different python/netCDF version issue. Let me add code to force divide by zero to be missing and see if that works on both of our environments.

CoryMartin-NOAA:
@srherbener try the test now, I think this should be set up properly now to force the inverse error divide by zero values to be 9e36 instead of INF.

srherbener:
I'm going to merge now

jjguerrette:
Is there any reply to Junmei's comment on vertical distribution?  We have seen this for radiosonde obs for quite a while.  The bottoms of the soundings appear to be filtered out.

srherbener:
@jjguerrette and @junmeiban, Sorry about merging this in without a reply to Junmei's concern about the vertical distribution. I think it would be best to submit a ZenHub issue and assign Anna, and others you feel appropriate, about this so that it doesn't get dropped. This way we can continue the discussion about the Temperature distribution while moving forward with code development.

Thanks!

shlyaeva:
@junmeiban, @jjguerrette: printing out some stats from the original gsi ncdiag file, it looks like 2/3 of all obs in the temperature file are below 400 hPa, overall almost 190000 obs. It's split between tv and tsen, with one of those being ~85K obs and another ~105K obs. I'll have to run ioda converters scripts and look into this more to understand why that's not what you see. I'm off work today, I'll let you know when I figure it out.

CoryMartin-NOAA:
@junmeiban @jjguerrette  @shlyaeva  I notice the same thing for sondes_tsen, but for sondes_tv the obs look like they are more evenly distributed throughout the troposphere. Perhaps it's an issue of the way these two quantities are reported in the BUFR files or the way GSI handles them (Tsen~Tv at upper levels where moisture content is small could be a theory?)

shlyaeva:
@junmeiban @jjguerrette @CoryMartin-NOAA: in my previous comment, I looked at all temperature obs in the original files (including aircrafts, surface stations, etc). When I subset the original GSI diag files to only look at radiosondes, I see the same picture as you all: there are only 50ish observations under 400hPa for tsen, and tv looks different. So this issue is not in the ioda-converter scripts, but in the way GSI uses radiosonde data (or maybe in the way prepbufr data are prepared).
@ADCollard: do you know why there are only few observations of radiosonde tsen below 400hPa in GSI? Is tv used mostly under 400hPa for some reason?

PR #148, opened 2019-04-22
Separate bin and lib for the python scripts to install

aerorahul:
@srherbener 
Can you please review this draft PR demonstrating the split between executable python scripts and libraries, and dynamically pointing to the correct location of the library.

Please do not merge into develop, as I have yet to update in many locations, pending your review of the change demonstrating above capability.

Change-Id: I1c9fdc25c0ee9a84709d236b848961956250c2e1


src/lib-python/CMakeLists.txt
```diff
@@ -1,22 +1,49 @@
...
+list( APPEND programs
   run-pyflakes.py
   run-mccabe.py
 )
 
-SET_TARGETS_DEPS( "${programs}"
+SET_TARGETS_DEPS( "${libs}"
                    ${CMAKE_CURRENT_SOURCE_DIR}
-                   ${CMAKE_BINARY_DIR}
+                   ${CMAKE_BINARY_DIR}/lib
```
CoryMartin-NOAA: What are everyone's thoughts on having subdirectories in lib/? Would it be better to have marine/, gsi-diag/, etc. in here or everything in one lib/ top level directory?

ShastriPaturi: My opinion is to have subdirectories with relevant names as you mentioned ... marine/ etc

travissluka:
seems like a good strategy to me

srherbener:
@aerorahul, thanks for coming up with this scheme! I think it looks good so I'm okay moving forward with this.

If I understand correctly, if we make subdirectories in the install "lib" directory for each of marine/, gsi-ncdiag/, etc. then we would have to specify all  of those paths in the "bin" scripts with sys.path.extend commands. For example, if we are installing in /usr/local and we have libs for marine, lib-python and gsi-ncdiag, then we would have to do something like the following in our bin scripts
~~~~~~~
sys.path.extend([ "/usr/local/lib/marine", "/usr/local/lib/lib-python", "/usr/local/lib/gsi-ncdiag" ])
~~~~~~~
If I've got this right, and at least for now, it seems easier to skip the subdirectories. We may want to reconsider however if placing all "lib" scripts in one directory gets cluttered later on.

src/lib-python/CMakeLists.txt
srherbener: Would it make sense to have a single subdirectory (py-iodaconv for example) in which we place all of the library type scripts? It doesn't seem that we have that many scripts, and this would keep them contained away from all of the .a and other files.

srherbener: @aerorahul, I agree that ultimately the library scripts should go in the standard python directory (lib/python2.7/site-packages/pyiodaconv). I also think it's a good idea to approach this incrementally, and that your scheme using lib/pyiodaconv is a good first step.

Thanks for doing this work and let's move forward with this plan.

aerorahul:
I am getting a bunch of errors, and it will take time to resolve them, since there are a few PR's in the queue.

The best approach is to make the developers of the individual converters make changes in their directories.
I made the changes to marine converters.
gsi-ncdiag has a bunch of libraries and executables. It would be good to combine the libs into a lib-gsincdiag.py and then move it to lib.
This is just an example and a suggestion.
@CoryMartin-NOAA what do you think?

CoryMartin-NOAA:
@aerorahul I'm fine with putting all of the non executable files into one long python file and then just have the high-level execution scripts remain in bin/ . It would be a good exercise/opportunity anyway to see what can be shared among the different observation types and remove some redundant code.

aerorahul:
@srherbener @travissluka @CoryMartin-NOAA 
I have updated this PR with all converters and is ready for your review.

Currently all tests (except ODB2 and NCEP coding norms) are passing with python 2.7.
GSI ncdiag tests for conventional and radiances fail with python 3.6 (as is also the case in develop)

srherbener:
I merged @CoryMartin-NOAA's PR which created a merge conflict with this PR. I went ahead and resolved the conflict and pushed that commit. I'm going to merge now since everything is working.

aerorahul:
@srherbener combine_conv.py will not work just by resolving the conflict. It will no longer be able to find the necessary modules from lib-python etc. I’ll push a bug fix later tonight.

PR #162, opened 2019-05-08
bugfix/utils-code-norms

srherbener:
This PR fixes code norm errors in the src/utilities scripts.


aerorahul:
Can this be part of `lib-python`?

srherbener:
Sure! The utilities subdirectory only has this one script so I think it makes sense to merge it with lib-python. I'll make that part of this PR.

PR #164, opened 2019-05-09
Concatenate conventional data with combine_conv.py

CoryMartin-NOAA:
combine_conv.py in gsi-ncdiag will take an input of N netCDF files in IODA format and output a concatenated file containing multiple locations and variables.

If a location is exactly the same for two files/variables, the location will not be duplicated and the value for each variable should be present for var(iloc). If the locations are independent of the other variables, the other variables will have undefined fill values in them.

Tested with trying to concatenate a subset of the original file with the original file results in only the # of locations for the original file (not orig + subset). Also concatenating all sfc_obs files results in a value of nlocs that is ~1/2 of sum(nlocs) of the separate files so I think this works correctly but I'm not 100% sure yet.

Using np.unique(axis=) which is not in older version of Numpy (i.e. on Theia it will only work with the Anaconda3 install as of now), but this is not a Python2 vs Python3 issue but rather the version of the Numpy library.

This pull requests also fixes the issues with GSI-ncdiag converters with Python3.


src/gsi-ncdiag/conv.py
```diff
@@ -227,7 +227,7 @@ def toIODAobs(self, OutDir, clobber=True):
                     for lvar in LocVars:
                         if lvar == 'Station_ID':
                             tmp = self.df[lvar][idx]
-                            locKeys.append([''.join(tmp[a])
+                            locKeys.append([b''.join(tmp[a])
```
aerorahul: List comprehension on L230 and 231 don't fit on 1 line? Makes it easy to read.

CoryMartin-NOAA: You know what, that is an artifact of using the automated pep8 script to fix the pycodestyle complaints, with the indentation the line length is too long for the default option. When I combine these into one lib-pygsidiag.py file, I'll try to sort back out these wacky line wrappings.

CoryMartin-NOAA:
> code looks Ok ~ is there a way to test these?

@aerorahul  I think the easiest way would be for me to make some dummy data by duplicating a file and replacing the name "air_temperature" for example with "virtual_temperature" and see if the values all are the same and match up. I can try this tomorrow.

PR #166, opened 2019-05-10
make combine_conv.py compatible with bin/lib structure

aerorahul:
Change-Id: Id204e317b16e456dd209369aac2c56ab75050607

### DIFF
--- a/src/gsi-ncdiag/combine_conv.py.in
+++ b/src/gsi-ncdiag/combine_conv.py.in
@@ -4,12 +4,15 @@
 # one output file with matching corresponding locations
 # and missing data where applicable
 
+import sys
 import netCDF4 as nc
-import ioda_conv_ncio as iconv
 import numpy as np
 import argparse
 from collections import defaultdict, OrderedDict
 import datetime as dt
+
+sys.path.append("@SCRIPT_LIB_PATH@")
+import ioda_conv_ncio as iconv
 from orddicts import DefaultOrderedDict
 
 vtypedict = {


srherbener:
My apologies for jumping the gun with the combine_conv.py script. 

It's strange however because I tested running combine_conv.py after running "make install" by simply running it without arguments to see if all the import statements work. And it did successfully import all modules. I even checked to make sure I didn't have any leftovers in my install directory from before the refactoring of the lib and bin scripts and there were none.

Not sure I understand how this worked, but I'm glad you caught the issue and fixed it. Thanks!

junmeiban:
Hi @CoryMartin-NOAA ,

I tried to combine two files into one. but I got an error:
Screen Shot 2019-05-10 at 9 30 01 AM

The files(aircraft_tsen_obs_2018041500.nc4 aircraft_q_obs_2018041500.nc4) are generated by proc_gsi_ncdiag.py  (#147) .  Do you have some idea about this error?

Thank you so much!

Junmei

CoryMartin-NOAA:
@junmeiban yes, this is an issue with your version of Numpy (needs to be I believe 1.13 or greater) because it adds the option axis= to np.unique().  Are you able to update Numpy on your machine (or try another installation?). This is not a Py2 vs Py3 issue but perhaps it will work if you use Py3 on your environment instead.

junmeiban:
@CoryMartin-NOAA , I load  numpy/1.13.3 and this script works! I get the combined obs now.
Many thanks.

junmeiban:
@CoryMartin-NOAA , I checked the variable from single file and combined file. The following figures are virtual temperature from single file(upper panel) and combined file(lower panel). It looks the observation number for Tv has reduced after combining.  
Screen Shot 2019-05-10 at 10 37 01 AM
Screen Shot 2019-05-10 at 10 36 42 AM

CoryMartin-NOAA:
@junmeiban ok yeah I definitely can see that in the lower atmosphere there are fewer points being plotted. Do you have a count of the difference (either total or percentage change)?  I'll look into this ASAP, I have an idea of a possible issue/bug but I'll need a baseline to compare it to.

srherbener:
@junmeiban, thank you for checking the data and bringing this issue to our attention!

Normally, issues like this should be reported by creating a ZenHub issue. @CoryMartin-NOAA, if you are on the verge of having this fixed, then let's not worry about a ZenHub issue for this. Otherwise, @junmeiban, could you cut and paste your plots and message into a new ZenHub issue. That would be good to have since this PR is closed and leaving the issue in the PR hides it from other people. Thanks!

CoryMartin-NOAA:
@junmeiban @srherbener yes open a issue and assign it to me, hopefully I can figure it out soon but no guarantees.

junmeiban:
Many thanks for reminding me. I will post a ZenHub issue for this.

PR #169, opened 2019-05-10
Add concatenation of GeoVaLs and Fix bug in combine_conv.py

CoryMartin-NOAA:
Fixes bug in combine_conv that filtered out too many Tv observations (among other issues) with incorrect indexing.

Also adds combine_geoval.py which when using -f will simply concatenate all of the input GeoVaLs files (unique locations not working 100% correctly in this one, it's more difficult because of less metadata)


CoryMartin-NOAA:
@srherbener oops, I forgot to run the test before I pushed the commit. Should be fixed now.

PR #180, opened 2019-05-16
Fix missing values in station_elevation and height for sfc obs

CoryMartin-NOAA:
Note: this should probably not be merged until my other pull request is finished (although the order doesn't matter much).

This fixes the issue where surface obs have a missing value for station_elevation and height to be 9999 (or 10009 for height for u/v, see below).
Now this should be replaced with the netCDF fill value.

Additionally, there is code in GSI to set u,v sfc obs to have a height of 10m + elevation, and to set height equal to station_elevation for surface T/q, but then the sfc_model operator has a hard coded assumption of 2m AGL. The converter class for conv has been modified to set height = 2 m + station_elevation for surface T/q obs and keep height = 10m + elevation for u/v.


srherbener:
@CoryMartin-NOAA, I tried this PR out and I'm getting the following coding norms error:
~~~~~~~~
+ pycodestyle --config=/Users/stephenh/projects/ioda-converters/.pycodestyle .
./gsi_ncdiag.py:468:83: E261 at least two spaces before inline comment
1       E261 at least two spaces before inline comment
<end of output>
Test time =   0.22 sec
----------------------------------------------------------
Test Failed.
"iodaconv_gsi-ncdiag_coding_norms" end time: May 20 09:52 MDT
"iodaconv_gsi-ncdiag_coding_norms" time elapsed: 00:00:00
----------------------------------------------------------
~~~~~~~~

srherbener:
No worries. Thanks for fixing it.

PR #186, opened 2019-05-24
Required 'feature/use_pressure' changes.

svahl991:
These changes to the ODB2-to-IODA converter are required due to the changes coming in PR's of the same branch name in IODA and [UFO](https://github.com/JCSDA/ufo/pull/327).

ODB2 files store pressure in Pa units, and the script no longer converts that to hPa when creating the corresponding IODA file.

Also, the CRTM variable names have been updated to reflect the change by UFO to using the CF names.

### DIFF
--- a/src/ODB2/Definitions/MO_amsua_Definition.yaml
+++ b/src/ODB2/Definitions/MO_amsua_Definition.yaml
@@ -38,16 +38,12 @@ column_variables:
     longitude: lon
     latitude: lat
     satellite_id: satellite_identifier
-    sat_zenith_angle: zenith
-    sat_azimuth_angle: azimuth
-    sol_zenith_angle: solar_zenith
-    sol_azimuth_angle: solar_azimuth
+    sensor_zenith_angle: zenith
+    sensor_azimuth_angle: azimuth
+    solar_zenith_angle: solar_zenith
+    solar_azimuth_angle: solar_azimuth
     scan_position: scanpos
     height: surface_height
-
-#TODO: change "sat_zenith_angle" to either "sensor_zenith_angle" or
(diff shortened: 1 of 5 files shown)


svahl991:
@srherbener 
I just realized that this PR is not ready for review yet because I need to update the output file for the odb ctest. However, the odb test isn't running for me in Singularity. I thought it would since I was inside the container. Do you have any ideas why it isn't running for me?

aerorahul:
@svahl991 is it possible that the python found via ecbuild is v3?

srherbener:
@svahl991, did you use the -DENABLE_ODB_API=1 option to ecbuild, and do this on a fresh build (ie, you may need to remove the current build directory and start over).

svahl991:
@srherbener @aerorahul 
I figured out the problem with the odb test not running. I went back and read Steve's instructions in his PR that said I need to ecbuild the repo with an additional flag in order for the odb ctest to run. Thanks for the help and sorry for the inconvenience.

svahl991:
@srherbener @shlyaeva 
I've updated the ctest output file now so this PR should now be fully ready for your review.

svahl991:
@srherbener , Won't every converter in this repository need to make modifications like this so that their output has pressure in units of Pa instead of hPa? If so, should we do all converters in this PR or in separate ones?

srherbener:
Yes, for the cases that are using log pressure. Between Cory's and your updates, we've probably hit most of the files (marine shouldn't be affected by this).

I think we should do the updates in smaller pieces in subsequent PR's. This one is ready to go and we should put it in "now" and fix anything we missed with subsequent PR's.

svahl991:
Thanks for the review @CoryMartin-NOAA !

PR #188, opened 2019-06-17
Creates individual IODA files for each obs type

aerorahul:
Separate out the ioda files, one-per-obstype.
This was found to be necessary since missing values were difficult to filter out (a filter will need to be implemented).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-build
+build*
 *DS_Store*
 ._*
 *.swp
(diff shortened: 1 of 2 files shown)


aerorahul:
@srherbener I don't think @guillaumevernieres has an opinion on these changes, and besides, I am the only one interested in keeping/using this capability.
I had to add 2 reviewers.

srherbener:
@aerorahul, sorry about the wait on this. I'm okay with only one reviewer since it doesn't impact the other flows. I'm still seeing all tests pass on my Mac, so I will merge it in now.

PR #192, opened 2019-06-21
bugfix/smap-sss-int-fails

### LINKED ISSUE #191: test_iodaconv_smap_sss now failing on Theia
I'm not sure who wrote the smap_sss test/converter but somehow now it's failing in the tests for ioda-converters (at least on Theia) both in develop and my pending PR branch (both @louiskouvaris  and I noticed this). Can someone (@srherbener perhaps) confirm that it still passes for them and I can try to figure out why it's now failing on this machine.

srherbener:
This PR fixes the intermittent SMAP SSS test failures. The problem was that in the input netcdf file, the 'time' variable is double precision while it's scale_factor and add_offset attributes are single precision. The fix is to change the attributes to double precision. I did this by using ncatted (netcdf attribute editor) which is part of the nco tool suite.

This works in my container on my iMac. Need to test this on Theia just to make sure.

We need to feed this issue back to the providers of the SMAP SSS netcdf files so this can get fixed at the source.

Closes #191


srherbener:
Great! Thanks for testing on Theia. It is weird. The order of automatic conversions must have changed, but I have no idea why that would happen.

guillaumevernieres:
@srherbener : Modifying the original file should not be the solution. We have zero control over the source, and even if they accepted, the fix may be weeks/months/years down the road ...

guillaumevernieres:
The problem is in the inconsistency across the netcdf modules, it looks like it's applying a scale of 1 and an offset of zero while reading the field.

srherbener:
@guillaumevernieres, I agree that we shouldn't have to modify the files. However, they are constructed incorrectly. It's okay to utilize the scale and offset, but the file came with single precision scale and offset for a double precision variable, and the result needs to be double precision. The script asks the netcdf interface (ie, the netcdf C code and python API from unidata) for the variable value and it is truncated to a single precision value when the interface returns it so there is nothing we can do about it in our scripts. The only way to fix this is to modify the file or modify the unidata netcdf interface.

The input file is clearly inconsistent and that is why I was suggesting that we have the supplier fix it.

PR #193, opened 2019-06-21
Revert "bugfix/smap-sss-int-fails"

srherbener:
Reverts JCSDA/ioda-converters#192


guillaumevernieres:
Thanks @srherbener, sorry for being picky, but that's a point of failure outside of ctest. @CoryMartin-NOAA , was it failing on Theia? It works on the container, which could also indicate that the netcdf module we use might not apply the scaling/offset byt default (?). We've had that issue in the past. 
I realized I'm late to the discussion, sorry again!

PR #196, opened 2019-06-28
Make sbuv_n19 obs files pass IODA nc checker

CoryMartin-NOAA:
Added in fixes that were done previously for conventional and radiance observations to have sbuv_n19 obs pass @srherbener 's IODA netCDF conventions checker.

This will also probably need to be done soon for the AOD class as well (I do not generate AOD files by default though but the capability is there)

@srherbener I also stopped using the ExtractObsData function as it seems that it did not correctly keep mssing values and instead assigned NaN to the missing values (or at least even if I set it to netCDF fill value when I later ran ncdump it displayed NaN instead of "-" like it does now. It may be worth looking into ExtractObsData and verifying it handles the missing values properly.


srherbener:
@CoryMartin-NOAA, thanks for the feedback about ExtractObsData. Do you have an example, or is it easy for me to create an example, of it creating nans in the output file. Thanks!

CoryMartin-NOAA:
@srherbener I noticed it with this sbuv change with using something like:

tmp = 1/0.
tmp[np.isinf(tmp)] = nc.default_fillval('f4')

print(tmp) showed [-,-] but was NaN after running it through ExtractObsData. Let me see if I have a commit that did it that you can use. Hold on.

CoryMartin-NOAA:
@srherbener  if you revert this branch back to commit # 96fc957 it should give that erroneous NaN in the netCDF files, if that's easy enough for you to do.

srherbener:
@CoryMartin-NOAA, I think I can just checkout the commit, so should be an easy thing to do. Thanks!

CoryMartin-NOAA:
@srherbener I just realized you'll need input data to process. 

./test_gsidiag.py -i ./diag_sbuv2_n19_ges.2018041500_ensmean.nc4 -o ./ -t oz

I'll email you the .nc4 file, it's under 1MB

CoryMartin-NOAA:
> @srherbener I just realized you'll need input data to process.
> 
> ./test_gsidiag.py -i ./diag_sbuv2_n19_ges.2018041500_ensmean.nc4 -o ./ -t oz
this is in bin/ of the build directory if that wasn't clear.

srherbener:
@CoryMartin-NOAA, I found the problem. I've got a fix on the bugfix/ext-obs-nan branch. If you find that helpful, go ahead and use it. I'm going to submit a PR for it since Hamideh might be running into this problem as well.

CoryMartin-NOAA:
@louiskouvaris can you review this soon and confirm that the new SBUV files generated still work with your operator and that they pass Steve's IODA netCDF conventions test? Thanks!

PR #199, opened 2019-07-11
Add capability for units metadata for netCDF observation files

CoryMartin-NOAA:
Adds VarUnits dictionary to NcWriter as an optional argument.

GSI nc diag file converter has a dictionary added units_values{} that should have all potential unit names (and new ones can be added/changed as anyone sees fit)

Line 307 in ioda_conv_ncio.py is commented out for now just so that all of the marine tests will still pass, but it adds the units: "hours since <analysis time>" to the time@MetaData variable

I don't want to get too thorough with this until I have more information on the role netCDF files will play with ODC, and if ODC can still use these units strings in the same way, etc.


CoryMartin-NOAA:
@srherbener I can change it to "units" no problem, probably best to do what you suggested in steps though, keep things simple for easier pull request reviews.

src/gsi-ncdiag/gsi_ncdiag.py
```diff
@@ -188,38 +188,41 @@
...
-    'Sfc_Wind_Speed': 'Sfc_Wind_Speed',
-    'Sfc_Wind_Direction': 'Sfc_Wind_Direction',
-    'Lai': 'Lai',
-    'Soil_Moisture': 'Soil_Moisture',
-    'Soil_Temperature': 'Soil_Temperature',
-    'Land_Type_Index': 'Land_Type_Index',
-    'Vegetation_Type': 'Vegetation_Type',
-    'Soil_Type': 'Soil_Type',
-    'Snow_Depth': 'Snow_Depth',
+    'Water_Fraction': 'water_area_fraction',
```
shlyaeva: thank you for fixing names as well! we'll have to update ufo to use those.

src/lib-python/ioda_conv_ncio.py
```diff
@@ -286,13 +292,19 @@ def WriteNcMetadata(self, MdataGroup, DimName, Mdata):
...
+            except KeyError:
+                pass
 
             # If we are writing out the datetime string, then we also
             # need to write out the time offset (from the reference date_time
             # attribute value).
             if (NcVname == "datetime@MetaData"):
                 ToffsetValues = self.ConvertToTimeOffset(Vvals)
                 self._fid[ToffsetName][:] = ToffsetValues
+                # self._fid[ToffsetName].setncattr_string("Units", "hours since "+self._ref_date_time.strftime("%Y-%m-%d %H:%M:%S")+" UTC")
```
shlyaeva: is this commented intentionally?

CoryMartin-NOAA: Just for now so that I didn't have to regenerate the test case data for the marine obs tests (because it would add it to them as well). But I thought it was an important placeholder to have for later (that the time@MetaData means hours since a reference time)

CoryMartin-NOAA:
@srherbener don't merge this until I have fixed the combine_conv.py script to also account for the units metadata for input variables. Should be a new commit for that by Thursday afternoon. Thanks!

srherbener:
@CoryMartin-NOAA, sounds good. I'll wait for the new commit.

src/gsi-ncdiag/gsi_ncdiag.py
jjguerrette: @CoryMartin-NOAA, as long as the Absorbers are updated, can we also update Clouds?
For example,
`atmosphere_mass_content_of_cloud_liquid_water` (total column quantity)
needs to be
`mass_content_of_cloud_liquid_water_in_atmosphere_layer` (layer quantity)

See 
https://github.com/JCSDA/ufo/issues/385
and
https://github.com/JCSDA/ufo/blob/420e2e1db46f9b1c2e656ee4436d990846711c46/src/ufo/ufo_variables_mod.F90#L39

CoryMartin-NOAA: @jjguerrette  sure, is it just the 6 you have in the linked ufo_variables_mod that need changed in the ioda-converters?

jjguerrette: Yes @CoryMartin-NOAA, only those 6.  Both ufo_variables_mod and ufo_crtm_utils_mod will need changes after the geovals files are updated.  I'm happy to make those code changes when it's time.  I have never created a geovals file though.

CoryMartin-NOAA: Perfect thanks, @jjguerrette . Once I've got the issues with the units in the combined conventional files as well as the subsetted ones, I'll regenerate the geovals files as well as the obs files with units and push them to the repositories. Concurrently, perhaps the easiest way is for you to make the code changes but don't create a pull request and we can merge the two (code changes and new geovals files) into one PR?

jjguerrette: The code changes for condensates are in UFO branch feature/CF_names.  Feel free to push other name changes there.

CoryMartin-NOAA:
@jjguerrette @srherbener @shlyaeva I think all of the changes should be made now to go ahead and review this pull request. I just submitted a job to re-generate the obs and GeoVaLs files using the most recent commit. If everything looks good once it completes (units, names, etc.) then I'll create a new branch and PR for updating the files in UFO and IODA.

srherbener:
No worries about the coding norms error - it was trivial to fix.

src/gsi-ncdiag/gsi_ncdiag.py
jjguerrette: Thank you Cory!

CoryMartin-NOAA:
One more commit, to change CO2/O3 from mass_fraction to mass_concentration to be consistent with UFO (and the concentration is probably the correct one, not sure if it's ppm(b)v or not in the model(s) though

jjguerrette:
> One more commit, to change CO2/O3 from mass_fraction to mass_concentration to be consistent with UFO (and the concentration is probably the correct one, not sure if it's ppm(b)v or not in the model(s) though

@CoryMartin-NOAA, I think the earlier change to `mass_fraction_` was correct since the units are normally ppm/ppb for gas-phase species.  That is a unitless quantity divided by a fixed multiplier.  The CF page states that the units of `mass_concentration_` are in mass per volume of air.  That would apply more often to aerosols, not gas-phase species.

jjguerrette:
@CoryMartin-NOAA going back to your previous question on ppm(b).  Are the those fractions by mass or volume in the ncdiag files?  UFO-CRTM is assuming volume.  Do we actually have volume_fraction_of_* instead of mass_fraction_of_* for CO2 and O3?  Those are not CF standard variable names, but fit the general templates.

CoryMartin-NOAA:
@jjguerrette they are from the GSI diag files and GSI-CRTM assumes mass mixing ratio for H2O and volume mixing ratio for the other absorbers as far as I can tell. So then I guess we want to use, mole_fraction_of_carbon_dioxide_in_air as the CF standard variable name for these absorbers? I think so since ppmv CO2 in the GHG community is the standard abbreviation for micromoles per mole and that's what is used generally.

jjguerrette:
@CoryMartin-NOAA, that sounds right.

srherbener:
I just tried with the latest commits. All tests pass, and the obs files pass the IODA checker. Looks great! I'm going to merge now. Thanks everyone!

PR #203, opened 2019-08-01
Minor build clarification

jmccreight:


jmccreight:
Should I open an issue for this? The issue would be more verbose than the change.

srherbener:
No need for an issue, I think it's covered with this PR.

I ran all the tests with this branch, and they all pass. Normally we need two approvals before merging, but since this isn't a code change, I'm going to go ahead and merge it in now.

Thanks for fixing the README instructions!

PR #209, opened 2019-08-21
Add capability to GSI nc diag converter to produce radar reflectivity and winds

CoryMartin-NOAA:


src/gsi-ncdiag/gsi_ncdiag.py
```diff
@@ -1314,3 +1332,211 @@ def toIODAobs(self, OutDir, clobber=True):
...
+        for ncv in self.df.variables:
+            if ncv in all_LocKeyList:
+                LocKeyList.append(all_LocKeyList[ncv])
+                LocVars.append(ncv)
+        # for now, record len is 1 and the list is empty?
+        recKey = 0
+        writer = iconv.NcWriter(outname, RecKeyList, LocKeyList)
+
+        nlocs = self.nobs
+        print self.obstype
```
srherbener: Please change to python 3 syntax. Thanks!

CoryMartin-NOAA: Oops, I'll just delete that print statement.

PR #212, opened 2019-08-26
Feature/gnssro fortran converter

### LINKED ISSUE #58: Integration of multiple conversion paths
Conversion goes from various file formats to IODA netcdf. We currently have conversion paths that go from the following input formats:
* BUFR
* GSI ncdiag
* Met Office ODB API
* Marine profiles
* GNSSRO

The GNSSRO path is a Fortran program using the libbufr.a library. All others are Python scripts. These paths have come from various contributors.

All of these need to write out a IODA netcdf file. They should, as much as possible, share the same netcdf writer routines so that code is not duplicated.

markjolah:
Created `CMakeLists.txt` for fortran NCEP bufr to NetCDF ioda converter and prefilter for GNSSRO data from COSMIC by @hailingz

Note: building this branch will require an up-to-date (as of today) version of [JCSDA/bufrlib](https://github.com/jcsda/bufrlib)

Closes #58


srherbener:
I tried building on my iMac and got the following ecbuild error:
~~~~~~~
CMake Error at src/gnssro/CMakeLists.txt:3 (find_package):
  By not providing "FindBufrlib.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Bufrlib", but
  CMake did not find one.

  Could not find a package configuration file provided by "Bufrlib" with any
  of the following names:

    BufrlibConfig.cmake
    bufrlib-config.cmake

  Add the installation prefix of "Bufrlib" to CMAKE_PREFIX_PATH or set
  "Bufrlib_DIR" to a directory containing one of the above files.  If
  "Bufrlib" provides a separate development package or SDK, be sure it has
  been installed.
~~~~~~~

I can get it to work be changing the find_package command in the gnssro CMakeLists.txt from `find_package(Bufrlib REQUIRED)` to `find_package(BUFRLIB REQUIRED)`. I thing the issue is that the module environment uses all upper case, the JCSDA/bufrlib repo builds the cmake config file with all lower case and the find_package command has mixed case. Just need to make the case consistent, and I'm okay however you prefer to do this.

Thanks!

markjolah:
Thanks Steve.  That didn't show up on my system because you are using modules, and the jedi-stack module I guess is the problem, using the wrong case...  If the bufrlib is installed to /usr or /usr/local, CMake will find it with either case format as it doesn't rely on the environment variable.  I will get a PR in to jedi-stack to address this.

markjolah:
Updated with CMake changes and also requires new jedi-stack branch [bugfix/jcsda-bufrlib-envvar](https://github.com/JCSDA/jedi-stack/tree/bugfix/jcsda-bufrlib-envvar) corresponding to [jedi-stack # 46](https://github.com/JCSDA/jedi-stack/pull/46)

srherbener:
Hi @CoryMartin-NOAA, I'm not sure how you get access to JEDI libraries on Theia/Hera. If you are using jedi-stack, then you need to use the feature/jcsda-bufrlib-envvar branch as noted by @markjolah above.

If you are not using jedi-stack, you need to build bufrlib from the JCSDA/bufrlib repo, install bufrlib, and the set the bufrlib_DIR environment variable to point to the directory containing the cmake configuration for bufrlib. For example:
~~~~~~~~
git clone https://github.com/JCSDA/bufrlib.git
cd bufrlib
git checkout master      # should be on the master branch by default

mkdir build
cd build
ecbuild -DCMAKE_INSTALL_PREFIX=$HOME/tools ..
make -j4
make install

export bufrlib_DIR=$HOME/tools/share/bufrlib/cmake
~~~~~~~~

After the above, you should be ready to try building this PR code.

Thanks for helping with the reviews!

CoryMartin-NOAA:
@srherbener Hera/Theia are down today because of some sort of electrical unforeseen circumstances so I am not able to compile/test/review yet.
@markjolah Is there a particular reason why this forked bufrlib needs to be compiled separately, and not just use the already compiled NCEPLIBS BUFR library available on NOAA machines? For example, CMake for GSI can link to the BUFR library just by setting the environment variable $BUFR_LIBd (Which on WCOSS-Dell-Venus is =/gpfs/dell1/nco/ops/nwprod/lib/bufr/v11.2.0/ips/18.0.1/libbufr_v11.2.0_d_64.a)

markjolah:
This is ready for re-review and hopefully merging.

Updates to gnssro_bufr2ioda:
 * Now creates NetCDF V4 file as output
 * Remove obs date filter and any references to window width or center
 * Add `datetime@MetaData` column in string format (keep `time@MetaData` also)
 * Cleanup spelling/indentation

Updates to build system:
 * Find bufrlib and NetCDF in top-level
 * Use variables `bufrlib_FOUND` and `NETCDF4_FOUND` to enable gnssro converter

@CoryMartin-NOAA To answer your question we would need to create a `FindBufrlib.cmake` to find versions of NCEP bufrlib that are not CMake enabled unlike `JCSDA/bufrlib`.  This will be done in a subsequent pull-request.  For now the only use cases of the gnssro converter programs will be for environments where the JEDI-stack is available in module or container form, in which case the JCSDA/bufrlib should be found correctly, causing the gnssro converter to be enabled.

srherbener:
@markjolah, the GNSSRO test was failing for two reasons: The reference file did not get updated after the updates to the output file, and the test was not being excluded when either of bufrlib or NETCDF4 were not found. Since I added that test in the first place, I thought I should fix it so I went ahead and did that.

CoryMartin-NOAA:
@markjolah @srherbener I've been in GSI land almost exclusively the last few weeks but I'll make sure that I prioritize time on this tomorrow for review. Should be relatively easy enough to use the FindBufrlib.cmake that is provided with GSI for use on NOAA HPC.

srherbener:
@CoryMartin-NOAA, no worries about being busy with GSI. Thanks for taking the time to review this PR!

hailingz:
@markjolah Thank you very much for having the gnssro converter ready in the repo!

I have a couple comments:
1) you might want to clear up the top lines/comments I put there since it is now not a draft and you can make it more general. I mean my name should be there any more as it goes to a public domain. 
2) why did you remove the time check? is that because the time window check would be in the DA?
Thanks

markjolah:
@hailingz 
1. I updated the comments, but this program is still temporary, so the comment is correct.  And as far as I know we will not be releasing these codes as "Public Domain", but likely under some open source licence.  I attributed the copyright to UCAR, which I believe is correct.
2. The time check was hardcoded to a time window of -3h and +3h from the reference time.  This did not work because we are now using the beginning of the window as the reference time, and we are using different width windows with the same program.  In general, I will only call this program with a .bufr file containing the correct observation set for the given time window, so there is no need to check each observation.

hailingz:
It is good!

CoryMartin-NOAA:
@markjolah @srherbener I'm unable to get the bufrlib to compile on Hera: 

```
bufrlib/src/restd.c(55): error: identifier "MAXNC" is undefined
      f77int desc, ncdesc, cdesc[MAXNC];
                                 ^
```

any suggestions, otherwise I don't know if I'll be able to test this converter unless I can link to the existing compiled BUFR libraries.

srherbener:
@CoryMartin-NOAA, perhaps the version of cmake is passing the requirement (3.9 or newer) in CMakeLists.txt, but in reality needs to be newer. I'm using cmake version 3.14.2 and maybe something got fixed/added to cmake between the version on Hera and 3.14.2.

CoryMartin-NOAA:
@srherbener could be since the version is exactly 3.9.0.

CoryMartin-NOAA:
@srherbener that fixed it, however I got another error when running ecbuild for the ioda-converters, seems the bufrlib:shared is hard coded in the CMakeLists.txt file, and I had the static library so I had to change that. Can the option for both be added to the CMakeLists.txt file before the PR is merged, or later, @markjolah ?

src/gnssro/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# src/gnssro/CMakeLists.txt
+#
+# Build gnssro ncepbufr to ioda NetCDF4 converter
+#
+
+if( HAVE_BUFRLIB_AND_NETCDF )
+    set(GNSSRO_SRC gnssro_bufr2ioda.f90)
+    add_executable(gnssro_bufr2ioda ${GNSSRO_SRC})
+    target_link_libraries(gnssro_bufr2ioda bufrlib::bufrlib_shared)
```
CoryMartin-NOAA: Can we also support static libraries? That is what was build following @srherbener 's directions in the conversation tab

srherbener: It looks like in jedi-stack we are building both the shared and static bufrlib libraries, but by default when building bufrlib you only get the static library. You have to add
`-DBUILD_SHARED_LIBS=1` to the ecbuild command line when building bufrlib to get the shared library.

I see two options:
* Change bufrlib CMake config so you get both shared and static libraries by default
* Change the line marked above to something like the following so that either static or shared can be used:
~~~~~~~~
target_link_libraries(gnssro_bufr2ioda bufrlib)
~~~~~~~~

I'm leaning toward the second option (and perhaps we should do both) since that might be necessary after getting the FindBufrlib.cmake script written which handles native installations on HPC systems. (The sys admins on the HPC systems may be inclined to only build one of static or shared.)

My vote is to put the change in to allow either static or shared (target_link_libraries command marked above) bufrlib libraries before merging this PR.

CoryMartin-NOAA: I agree it would be best if this change could be made before the PR is merged.

srherbener: I'd like to get this resolved and merged in this week. I realize that @markjolah is swamped with other high priority tasks. @markjolah, @CoryMartin-NOAA is it okay if I make the change I proposed (change the target_link_libraries command) and then merge this in. Thanks!

CoryMartin-NOAA: Works for me

srherbener:
@CoryMartin-NOAA, I've updated the target_link_libraries command with what I believe will select either static or shared bufrlib libraries for linking, with a preference for static if both versions exist. It works on my iMac. Could you test on Hera? Thanks!

@markjolah, it looks like the bufrlib CMake configuration sets bufrlib_LIBRARIES to either static or shared depending on which were compiled. It looks like this is a safe way to do this, but I want to double check with you on this. Thanks!

CoryMartin-NOAA:
@srherbener using the command:
rm -rf build && mkdir build && cd build && ecbuild -DCMAKE_PREFIX_PATH=$JEDI/tools ../
compiled and the test ran successfully on Hera. Note this was with using the JCSDA bufrlib compiled and installed to $JEDI/tools and not 'module load bufr'.

srherbener:
@CoryMartin-NOAA, thanks for testing on Hera. Is this acceptable for merging in this PR? Thanks!

CoryMartin-NOAA:
@srherbener it is in my opinion.

markjolah:
@srherbener Seems all good on my end.

srherbener:
Thanks @CoryMartin-NOAA and @markjolah for your reviews! I'll merge it in now.

PR #224, opened 2019-10-10
Fixed a bug in the n_vrsn variable to read 6 characters of the WMO si…

ShastriPaturi:
This PR addresses a bugfix in godae_ship2ioda.py.
The godae_ship2ioda.py converter was failing to convert the FNMOC ship data to ioda format and has bug in the n_vrsn variable related to reading the WMO sign number.
The original fortran code (ocn_obs.f) reads 6 characters of the WMO sign. 

The bug has been fixed and tested.


aerorahul:
Thanks for exploiting this untested feature.

PR #225, opened 2019-10-10
Read a window in time of the Argo climatology

aerorahul:
so we can keep a single file for the entire climatology and read from parts of the same file and target the time window of interest.

Change-Id: I86a0a4b5edc32a79b0da5bb9694db98878f17c95


srherbener:
Could someone please review since we need 2 approvals before merging. Thanks!

srherbener:
@guillaumevernieres, thanks for the review! I'll merge now.

PR #227, opened 2019-10-14
Two bugs with emc_ice

flampouris:
The code had one bug and there was an issue with the time convention of the data. 

More specific: 
1. Loading the ioda-converter modules was transferred after appending the path.
2. The time convention was a four-digit integer, which is impossible for 12-9AM, making the  DateTime fail when 00:00. 

The code tested by converting successfully 30 days of observations.


flampouris:
Thank you @*srherbener*

PR #229, opened 2019-10-22
Implement CI/automated testing 

mer-a-o:


mer-a-o:
@srherbener, the problem with intel CodeBuild is that not all Python dependencies are installed in the container. We can merge this for now and ignore the intel Codebuild failure until we update the intel container with the dependencies?

srherbener:
@mer-a-o, would it be okay to shut off the Intel test until the container is properly updated? If we merge and ignore the test failure, I'm concerned that people will get used to ignoring the checks. We could submit a ZenHub issue to remind us to add the Intel test in a future PR.

mer-a-o:
@srherbener : Good idea. I disabled CodeBuild intel for now but ```buildspec_intel.yml``` is there for when the container is ready.  The code is ready for review and merge.

mmiesch:
As we add testing to more repos, do we need both Travis and CodeBuild?  It was useful to compare the two but now it seems to me that CodeBuild might be sufficient.

Regarding the intel container, it would be easy to add the python modules to the existing intel 17 container.  But I was hoping to kill two birds with one stone and add the python container when we move to intel 19.  So, if it's not urgent, we can just disable the intel test for now until I can reduce the size the intel 19 container (the intel19 container does have the python tools).

PR #231, opened 2019-11-11
Add exising LD_LIBRARY_PATH to bash script

mmiesch:
So @travissluka and @mer-a-o - this is why the ioda-converters tests are failing in the intel containers.  Python wants to access the library ```libimf.so``` which is in the ```LD_LIBRARY_PATH```.  So, running the python scripts manually works fine but when the compare script runs them, it resets the paths and causes this error:
~~~~~
python: error while loading shared libraries: libimf.so: cannot open shared object file: No such file or directory
~~~~~
This bugfix fixes that so all ioda-converters tests now pass in the intel17 container.  I haven't tried yet with the intel 19 container.

### DIFF
--- a/test/iodaconv_comp.sh
+++ b/test/iodaconv_comp.sh
@@ -5,7 +5,7 @@
 # argument 1: the command to run the ioda converter
 # argument 2: the filename to test
 
-export LD_LIBRARY_PATH=${CTEST_LIBRARY_PATH}
+export LD_LIBRARY_PATH=${CTEST_LIBRARY_PATH}:${LD_LIBRARY_PATH}
 
 $1 && \
 nccmp testrun/$2 testoutput/$2 -d -m -g -f -S


mmiesch:
@flampouris - I just now uploaded new intel 17 containers that have the ```I_MPI_SHM_LMT``` environment variable set to ```shm``` when running bash (set in the system .bashrc file).  I believe @travissluka was finding that this was sufficient to get the soca tests passing.  I think the intel 19 container should be ok as is.  Please let me know if you have any other issues.

flampouris:
@mmiesch I had the issue on Hera, outside of the containers. 

Before approving the PR, I tested the code on Hera, and everything went as expected.

mer-a-o:
I activated CodeBuild Intel tests. The tests will run on aws after we merge this PR.

srherbener:
Anyone know why the Travis-CI test didn't pass? I'm re-running it now.

Looks like it's safe to merge this in since the others have tested on several different platforms. Does anyone object to merging this if the Travis-CI test fails again?

mer-a-o:
@srherbener I was able to reproduce the error in Travis-CI. It shows up when building in debug mode and it is related to gnssro_bufr2ioda. I think we can merge this and address this problem in a separate PR.

srherbener:
@mer-a-o, sounds good and thanks for debugging the Travis-CI test failure!

PR #237, opened 2019-11-21
Clang for CodeBuild

mer-a-o:


mer-a-o:
@srherbener: tests 6-18 fails in clang container. Looks like the error is related to ```nccmp``` as you guessed. 
```
6: nccmp: error while loading shared libraries: libnetcdf.so.15: cannot open shared object file: No such file or directory 
```
I use ```jcsda/docker-clang-mpich-dev:latest container``` in CodeBuild 
I cannot reproduce the error on my laptop inside the Docker container though.

srherbener:
@mer-a-o, this is a different error that what I saw before, but by coincidence with the same program (nccmp). It looks like the nccmp is loading the wrong netcdf library (ie, a version that is different than the one it was built with).

Python installs the netcdf library version that goes with the netCDF4 python package. It's possible that nccmp is accidentally finding the Python netcdf libraries and these could be different than the versions that nccmp was built with. Python should be installing these in /usr/local/lib/python3 or /usr/local/lib/python3/site-packages. If the Python path is on LD_LIBRARY_PATH before /usr/local/lib, then that would cause this situation.

mer-a-o:
@srherbener Thanks for the comment. I removed mpi setup steps I use for gnu/mpi container from clang yml and it fixed the problem! Now all 3 aws CodeBuild tests pass for Ioda-converters. I will add the clang badge to README.md and then it will be ready for merge.

mmiesch:
great - thank you both for figuring this out!

PR #240, opened 2019-12-20
QC GMAO obs.

aerorahul:
Remove observations from the GMAO S2S system that are far from the expected value.
This could be done in the QC filter as well, but then one would have to remember to implement that everytime.  This way, the QC is done before we even get to assimilating these bad obs.

Change-Id: I3abbfd790440db559e1677779eaa6fbc1175d197

### DIFF
--- a/src/marine/gmao_obs2ioda.py.in
+++ b/src/marine/gmao_obs2ioda.py.in
@@ -155,6 +155,12 @@ class IODA(object):
 
             for n in range(obs.data['nobs']):
 
+                oval = obs.data['value'][n]
+                oerr = obs.data['oerr'][n]
+
+                if discardOb(varName, oval):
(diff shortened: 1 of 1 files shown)


srherbener:
I'm going to merge this in now so it can be used now, but I agree with @ytremolet that in the long run we should separate pre-processing and format conversion.

aerorahul:
@ytremolet I agree with the use of generic filters. In the future, we won’t be using this converter, and instead use the individual marine converters.  This particular script allows for a format conversion from current GMAO ODAS observation format to IODA. The GMAO ODAS already goes through some QC, but I found some odd values slipped through. This PR takes care of those bad observations.  This is meant to be a one stop shop for all things GMAO ocean observations.

PR #247, opened 2020-02-07
Add a netcdf to odb converter

srherbener:
This PR introduces a python script, nc2odb.py, that converts a ioda netcdf file to a ioda odb file.

In addition a shell script is included, run_nc2odb.sh, that will run nc2odb.py on all netcdf files under a given directory. run_nc2odb.sh will name the output odb files by substituting the .nc4 or .nc suffix with .odb in the name of the input netcdf files. Eg, `my_obs.nc4` will get converted to `my_obs.odb`.

The initial purpose of this script is to create duplicate odb files from the ioda netcdf files for testing and comparison.


src/odc/nc2odb.py
```diff
@@ -0,0 +1,245 @@
...
+
+    RowNum += 1
+
+os.close(CsvFile)
+
+# Run odc import
+OdcCmd = "odc import {} {}".format(CsvFname, OdbFname)
+print("Running: ", OdcCmd)
+print("")
+os.system(OdcCmd)
```
aerorahul: please use subprocess.

CMakeLists.txt
```diff
@@ -64,6 +64,18 @@ else()
   message( STATUS "Fortran components disabled since both bufrlib and NETCDF are not found")
 endif()
 
+# ODC and odyssey are optional. If found, the following variables are set
+#  odc_FOUND
+#  odyssey_FOUND
+find_package(odc)
+find_package(odyssey)
+if (odc_FOUND)
```
aerorahul: are these prints necessary?

src/CMakeLists.txt
```diff
@@ -2,5 +2,6 @@ add_subdirectory(lib-python)
 add_subdirectory(gsi-ncdiag)
 add_subdirectory(marine)
 add_subdirectory(ODB2)
+add_subdirectory(odc)
```
aerorahul: Only add this if ODC/ODB are found.

src/odc/nc2odb.py
```diff
@@ -0,0 +1,245 @@
...
+
+# Arguments are okay. Read in the odb file using the xarray interface. Then covert to
+# a pandas dataframe, write out the pandas into the odb file.
+
+# Select the variables that are to be transferred to the output which are all the ones
+# with nlocs as the first dimension. Also change datetime@MetatData to two integer
+# variables: date@MetaData and time@MetaData (with formats YYYYMMDD, hhmmss respectively).
+NcDs = nc.Dataset(NetcdfFname)
+IntMissingValue = 2147483647
+DoubleMissingValue = -2147483647
```
aerorahul: shouldn't this be -2147483647. 0 or float(IntMissingValue) ?

src/odc/nc2odb.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+###################################################################################
+# MAIN
+###################################################################################
+ScriptName = os.path.basename(sys.argv[0])
+
+# Parse command line
+ap = argparse.ArgumentParser()
+ap.add_argument("input_netcdf", help="path to input netCDF4 file")
```
aerorahul: If we passed a list of netcdf files, we could thread/parallelize it.  This could speed up when processing a directory of files.
May be another time, or if you want to wait, I can provide a parallelized/threaded patch on Monday.

src/odc/nc2odb.py
srherbener: Currently, the script is using the `odc import` command to build the output ODB file. The import command wants a CSV file for input and the integer style missing value for a variable of type double is working. Plus, using an integer value makes it simpler to get all the digits to come out correctly compared to using a float value.

It would probably be good to eventually replace the `odc import` method with methods in odyssey (odc python api) so we can avoid creating giant CSV files. However, I would like to defer this to a subsequent PR.

src/odc/nc2odb.py
srherbener: The run_nc2odb.sh companion script creates a list of netcdf files and runs nc2odb.py on individual files. This way, the list of netcdf files can be gigantic and not overflow the capacity of wildcard expansion in the shell. Would it make more sense to parallelize the run_nc2odb.sh script instead of the nc2odb.py script?

src/odc/nc2odb.py
aerorahul: Ok. I defer to the inner workings of `odc import`.

src/odc/nc2odb.py
aerorahul: Either way is fine.  Parallelize internally, such that each thread/process handles one file, or externally through the shell script.  The latter implies, all threads/processes need to be completed till the next batch of threads/processes start. 

To avoid overflow, you have to cap the number of parallel processing; be it shell or in the python.

We can leave this as is and re-visit if we find that we are waiting for a while on this converter.  How does that sound?

markjolah: I view this as something a higher-level workflow should be doing.  The `obs-convert` app in JEDI-Rapids is currently using process-based parallelism with to run this script when converting.

srherbener: I think we are good for now, and we can address this in the future if the need arises.

I'm going to go ahead and merge now. Thanks for the reviews!

PR #248, opened 2020-02-24
Feature/cryosat freeboard

YanHao-NOAA:
Added Cryosat-.DBL file converter and test case.


srherbener:
@YanHao-NOAA, thanks for submitting this PR. It looks like there is something wrong in the test fixtures which is likely not related to your changes. I'll look into this.

srherbener:
@YanHao-NOAA, we have a fix for the test failures in another PR. Once that gets merged in, please merge develop into this PR to pick up that fix.

In the meantime, I noticed that the cryosat DBL test was not running the conversion yet was passing. I made a fix in the test configuration for cryosat DBL and now the test fails because a python package is missing:

```
362 Traceback (most recent call last):
363   File "/Users/steveherbener/projects/ioda-converters/build/bin/cryosat_ice2ioda_DBL.py", line 41, in     <module>
364     import cryosat_toolkit.read_cryosat_L2 as read_cryosat_L2
365 ModuleNotFoundError: No module named 'cryosat_toolkit'
```

If the cryosat_toolkit is code you wrote for this converter, we should check it into ioda-converters (in the src/lib-python directory). If it's a package that needs to be installed into our python installation, where do we install it from?

Thanks!

YanHao-NOAA:
Hi Steve,
The  cryosat_toolkit was originally installed in anaconda/1.0.0, if you are
not using module anaconda/1.0.0 anymore, then it won't show up.

Thanks
Yan

srherbener:
Thanks @YanHao-NOAA. When I try ``conda search cryosat_toolkit``, it cannot find it. Is there a special site that I have to download from?

Thanks

YanHao-NOAA:
Hi Steve,
The webpage to download the toolkit is:
https://github.com/tsutterley/read-cryosat-2
At the bottom of that page:
A zip archive of the latest version is available directly at:
https://github.com/tsutterley/read-cryosat-2/archive/master.zip

Thanks
Yan

test/CMakeLists.txt
```diff
@@ -186,9 +188,18 @@ ecbuild_add_test( TARGET test_iodaconv_cryosat2
                        "${CMAKE_BINARY_DIR}/bin/cryosat_ice2ioda.py
                        -i testinput/cryosat2_L2_test.nc
                        -o testrun/cryosat2_L2.nc
-                       -d 2019092100"
+                       -d 2019092112"
                       cryosat2_L2.nc)
 
+ecbuild_add_test( TARGET test_iodaconv_cryosat2_DBL
```
srherbener: This test is taking 6-8 minutes to run on my iMac. Not sure why it is so slow - I see that it runs in ~50 seconds during the automated testing.

The 6-8 minute runtime is prohibitive. Would it be acceptable with you make it so that this test doesn't run by default (and perhaps set a variable to make the test run). If this sounds okay with you, you might want to create another script (alongside test/iodaconv_comp.sh) that checks an environment variable and if not set returns true (pass), and if set then runs iodaconv_comp.sh as is done here. That way you can control the test without having to re-run cmake. Once we get the nightly test scheme going then we can place this test into the nightly tests.

Thanks!

YanHao-NOAA:
Hi Steve,
Yes, I am okay with the solution of the test does not run by default. Would
you please help with setting up this?

Many thanks
Yan

srherbener:
I agree that the best option would be to thin the test file. Who can help us figure out how to do that?

Perhaps the package that contains the read_cryosat_L2.py library has other utilities that could be used for doing the thinning.

flampouris:
@guillaumevernieres  , @YanHao-NOAA has expressed the same concerns about the file sizes and she has been coordinating with @srherbener from the moment that she realized the issue. 

We do not have a way to write DBL files and has no meaning to develop a dbl writer for creating the test. 

@srherbener , I created last night the optional ctest, but this defeats the purpose of the ctest. So, I do not support this solution.

My suggestion is to identify a small dbl file and update the ctest input/out dataset. 

@YanHao-NOAA 
By scanning 2016, it appears that this file is the smallest: https://science-pds.cryosat.esa.int/?do=download&file=Cry0Sat2_data%2FIce_Baseline_C%2FSIR_GDR%2F2016%2F05%2FCS_OFFL_SIR_GDR_2__20160510T110838_20160510T124752_C001.DBL
 
Please, download the file, rename it to a cute short name, recreate the test and push it to GitHub (please do not forget to remove the old test files first). 

So, we can see how long it takes for the default test machines to process it.

YanHao-NOAA:
@flampouris Great Idea. I tested the smaller file you found and it took about 16 seconds on Hera. I will find another smaller size file and do the test again.

YanHao-NOAA:
I updated feature/cryosat_freeboard with a much smaller sized test case. It took about 12 seconds on Hera.

srherbener:
@guillaumevernieres, @aerorahul I tried the smaller test file on my iMac (outside the container) and the test still takes 50-90 seconds. Much better, but not great. I then tried the same inside Singularity on the same iMac and the test runs in 7 seconds!

I'm inclined to think that my iMac configuration may be problematic causing perhaps a file io performance issue, the test time is going to okay on most systems (5 - 15s) and we should merge this PR in. Do you agree?

guillaumevernieres:
> @guillaumevernieres, @aerorahul I tried the smaller test file on my iMac (outside the container) and the test still takes 50-90 seconds. Much better, but not great. I then tried the same inside Singularity on the same iMac and the test runs in 7 seconds!
> 
> I'm inclined to think that my iMac configuration may be problematic causing perhaps a file io performance issue, the test time is going to okay on most systems (5 - 15s) and we should merge this PR in. Do you agree?

@srherbener : Seems reasonable. Thanks @YanHao-NOAA and @flampouris .

Issue #244, opened 2020-02-05
gsi-ncdiag Converter: Eliminate nrecs


markjolah:
Currently the `gsi-ncdiag` converter package is producing improperly formatted NetCDF IODA files.  These files are retaining a `nrecs` dimension and some are also retaining variables indexed by `nrecs`.  This is a problem, as these invalid `nrecs` indexed variables are preserved in the HofX output IODA files, and this can lead to problems concatenating the output HofX files from each processor.  Additionally, it makes it impossible to concatenate GSI observation IODA files to form longer time windows, as each window may disagree on the `nrecs` dimension.

To solve this issue, the `gsi-ncdiag` python package will need to be updated and tested with the GSI input files.  I don't currently have these original files, so if someone could provide them, I maybe able to help debug this issue.

Here is a list of GSI output files from the July 2019 run @emilyhcliu has completed, and for each file any references to `nrecs` dimension or attributes should be removed and especially any variables indexed by `nrecs` should be removed.  In particular some of the conventional obs are producing two different variables for Station_ID with different capitalization and indexing dimensions.

```
F: aircraft_obs_2019070100.nc4
	nrecs = 4333 ;
		:nrecs = 4333 ;
F: aircraft_q_obs_2019070100.nc4
	nrecs = 276 ;
	char Station_ID@RecMetaData(nrecs, nstring) ;
		:nrecs = 276 ;
F: aircraft_tsen_obs_2019070100.nc4
	nrecs = 4333 ;
	char Station_ID@RecMetaData(nrecs, nstring) ;
		:nrecs = 4333 ;
F: aircraft_uv_obs_2019070100.nc4
	nrecs = 4135 ;
	char Station_ID@RecMetaData(nrecs, nstring) ;
		:nrecs = 4135 ;
F: airs_aqua_obs_2019070100.nc4
	nrecs = 1 ;
	int rec_id@RecMetaData(nrecs) ;
		:nrecs = 1 ;
F: amsua_aqua_obs_2019070100.nc4
	nrecs = 1 ;
	int rec_id@RecMetaData(nrecs) ;
		:nrecs = 1 ;
F: amsua_metop-a_obs_2019070100.nc4
	nrecs = 1 ;
	int rec_id@RecMetaData(nrecs) ;
		:nrecs = 1 ;
F: amsua_metop-b_obs_2019070100.nc4
	nrecs = 1 ;
	int rec_id@RecMetaData(nrecs) ;
		:nrecs = 1 ;
[...]

PR #249, opened 2020-02-25
WRFDA radiance ncdiag converter

jjguerrette:
Adds a converter for WRFDA radiance ncdiag files.  Have tested on 23MB all-sky ABI diag files with 10 IR channels.  Currently the test is not exercised, because I need to add input/output conversion files with a lower number of observations.

The code is modeled after the GSI NCDIAG converter Radiances class, with some modifications for this special case.


jjguerrette:
Coding norms are failing.  I will have a look soon.

srherbener:
@jjguerrette, thanks for submitting this PR. It looks like there is something wrong in the test fixtures which is likely not related to your changes. I'll look into this.

srherbener:
@jjguerrette, @mer-a-o helped get the test issues fixed and my latest commit has installed that fix into this PR. Looks good now.

src/wrfda-ncdiag/wrfda_ncdiag.py
```diff
@@ -0,0 +1,407 @@
...
+        i = False
+        for s in rad_platform_sensor_combos:
+            if s in splitfname:
+                self.platform_sensor = s
+                i = splitfname.index(s)
+        if not i:
+            raise ValueError("Observation is not a radiance type...")
+
+    def read(self):
+        import netCDF4 as nc
```
markjolah: Why do imports here as opposed to global scope?  It seems NetCDF4 will always be available.

src/wrfda-ncdiag/wrfda_ncdiag.py
```diff
@@ -0,0 +1,407 @@
...
+        self.validtime = dt.datetime.strptime(tstr, "%Y%m%d%H")
+        # sensor and satellite
+        self.sensor = self.platform_sensor.split('-')[-1]
+        satellite = "-".join(self.platform_sensor.split('-')[0:-1])
+        if satellite in wrfda2crtm_satellite_map:
+            self.satellite = wrfda2crtm_satellite_map[satellite]
+        else:
+            print("ERROR: Satellite not found in wrfda2crtm_satellite_map:")
+            print(satellite)
+            os._exit(1)
```
markjolah: This should be `sys.exit(1)`

src/wrfda-ncdiag/wrfda_ncdiag.py
```diff
@@ -0,0 +1,407 @@
...
+     to the JEDI/IODA observation format
+        """
+        import ioda_conv_ncio as iconv
+        import os
+        import errno
+        from collections import defaultdict, OrderedDict
+        from orddicts import DefaultOrderedDict
+        import numpy as np
+        import datetime as dt
+        import netCDF4 as nc
```
markjolah: Move imports to global scope.

src/wrfda-ncdiag/wrfda_ncdiag.py
```diff
@@ -0,0 +1,407 @@
...
+
+        # global attributes
+
+        AttrData["date_time_string"] = self.validtime.strftime("%Y-%m-%dT%H:%M:%SZ")
+        AttrData["satellite"] = self.satellite
+        AttrData["sensor"] = self.sensor
+
+        # set dimension lengths in the writer since we are bypassing
+        # ExtractObsData
+        writer._nrecs = nrecs
```
markjolah: I think we are trying to eliminate the nrecs variable.  See #244.

srherbener: There will need to be a number of changes to the ioda-converters code before nrecs can be eliminated. It's possible that @emilyhcliu already made these changes in order to address #244. In that case we should make a PR from her updates.

Regardless, I think the code changes for eliminating nrecs is outside the scope of this PR and we should do that work in a separate PR (whether it comes from Emily's work or we create a new PR). We can leave in this statement for now, and update accordingly.

Thanks!

emilyhcliu: @srherbener I will submit the PR for #244 today.

srherbener: @emilyhcliu, sounds great. Thanks for taking care of this!

jjguerrette:
All, for the most part I was following the existing interfaces to the writer in the gsi-ncdiag Radiances class.  If those are changing, then this will need to change.  I will have time to update this branch next week some time, but the PR will have to wait until then.  Thank you @markjolah for the review.

srherbener:
@jjguerrette, I merged in a PR (#256) today that removed the nrecs and record metadata variables from  the NcWriter class since these data are being ignored by the IODA reader. I merged in develop and removed the corresponding  items (setting of nrecs and rec_mdata, and the record_number@MetaData variable) from the wrfda_ncdiag.py script. Check out the recent commit 9ef40ba.

Since there isn't a test yet, I'm not sure if I've broken anything. My apologies if something is broken, and I can help with fixing it.

jjguerrette:
@srherbener, after fixing a small bug (pushed to this branch), the program runs.  However, the output netcdf file is empty.  I'm trying to figure out why.

srherbener:
@jjguerrette, do you want me to help debug? If so, can you send me sample input so I can reproduce the error. Thanks!

jjguerrette:
Nevermind Steve.  It was a simple fix.  The challenge is that the multiprocessing Pool somehow eats up all the error messages.  Once I circumvented the Pool, the error messages showed up and it was easy to fix.  Hopefully I have time in the coming weeks to add a small test file.  Then this converter will become easier to maintain.

src/wrfda-ncdiag/wrfda_ncdiag.py
jjguerrette: I moved the imports to global.  If that is the preferred convention, then other converters could be updated too.

jjguerrette:
Ready for review again @markjolah.

srherbener:
@jjguerrette, the input test file is a netcdf file, correct? If so, ncks (part of nco package) has an option (-d) that can select along a dimension. For example, the first 100 entries along a dimension (nlocs or nobs) could be selected to trim down the current test file. Would doing something simple like that work to create a valid small test file?

jjguerrette:
In general, I think it would be preferable to pick a random set of locations.  However, the first 100 would be a good start at least.

jjguerrette:
Alright I see how nimble ncks is.  How can I upload input/output files for the test?

jjguerrette:
Now there is a ctest in place.  I added the input/output files, which seem to have been added to LFS properly.  Please check that I did this correctly as I have not used LFS very much.

srherbener:
@jjguerrette, I think you are in good shape with git lfs. In the top directory of ioda-converters, there exists a file ``.gitattributes`` which contains:

```
*.nc filter=lfs diff=lfs merge=lfs -text
*.nc4 filter=lfs diff=lfs merge=lfs -text
*.bin filter=lfs diff=lfs merge=lfs -text
*.bufr filter=lfs diff=lfs merge=lfs -text
*.DBL filter=lfs diff=lfs merge=lfs -text
*.odb filter=lfs diff=lfs merge=lfs -text
src/ncep/xx120 filter=lfs diff=lfs merge=lfs -text
```
These show file name patterns that will automatically be stored in git lfs. The patterns contain .nc and .nc4 (as suffixes) so your files are covered. Another thing you can do if you are not sure is use the command: ``git lfs ls-files``. Running this inside ioda-converters reveals that the two files you added are being manages by git lfs.

jjguerrette:
No problem!  It's great to have the code exercised automatically.

PR #252, opened 2020-03-03
Add converter for GIIRS longwave radiance

srherbener:
This PR introduces a script for converting longwave radiance obs type from a GIIRS netcdf file into a IODA netcdf file. The script is called "giirs_lw2ioda.py". I created a new source directory (src/nsmc) in anticipation that more conversion scripts will be needed in the future (perhaps for the MW and VIS radiance obs types).

I was able to successfully read the file created by giirs_lw2ioda.py with the IODA reader.

I created a test from the sample GIIRS file: SSEC_V0.1.3_FY4A-_GIIRS-_N_REGX_0995E_L1-_IRD-_MULT_NUL_20170302080000_20170302081049_016KM_002V1.nc which has the first 10 locations and first 30 channels for LW and MW data, and the first 10 rows and first 10 columns of the VIS data. I did this so that the test could run fast.

The script successfully completes when run on the original sample file.

@weihanucar, I would like for someone on your end to eventually take over development of the NSMC converter scripts. Hopefully, there is enough in the script now that it can show someone how to add more variables, and I am happy to help train someone. If you have someone who can do this, please add them as a reviewer. If you don't have a person to do this, we can discuss further and work something out. Thanks!


srherbener:
@weihanucar, I moved your new versions into the src/nsmc/  directory so that they remain associated with the GIIRS converters.

I put .new suffixes on the files so that the syntax checker would not run on them. They currently fail the syntax check, plus they have hard coded paths to the ioda_conv_ncio.py library. I figure that you can fix these issues in a subsequent PR.

If you are happy with this, can you please approve so I can merge this into the develop branch.

Thanks!

PR #264, opened 2020-04-03
Removed unused imported package (codyssey)

srherbener:
This PR cleans up the nc2odb.py script which had a leftover ``import codyssey`` command but was not using that package. This breaks when we go to the release versions of odc and pyodc since the package names in pyodc changed.

### DIFF
--- a/src/odc/nc2odb.py
+++ b/src/odc/nc2odb.py
@@ -14,7 +14,6 @@ import tempfile
 import argparse
 import datetime as dt
 import netCDF4 as nc
-import codyssey as odc
 
 ###################################################################################
 # SUBROUTINES


srherbener:
Just to be clear, if the ``import codyssey`` command were to remain in the nc2odb.py script, then that will break when we go to the released versions of odc and pyodc.

srherbener:
Please review soon. This is needed before we switch over to the release versions of odc and pyodc and we are rapidly approaching building new containers with the released odc and pyodc. Thanks!

srherbener:
@mmiesch, could you review this soon. It's just a simple one-liner. Thanks!

srherbener:
No worries, we are all pretty busy these days! The nc2odb.py script doesn't use the python odc interface at all. Instead it uses an application called 'odc' that is built with the odc libraries. This script works with both the pre-release and release versions of the odc application, so we are good to go with this PR.

Thanks!

PR #265, opened 2020-04-03
Add simple odb to netcdf converter

srherbener:
This PR introduces a python script (odb2nc.py) to help convert a ioda formatted odb file into a ioda formatted netcdf file for the jedi-rapids flow. This is a functional equivalent of an existing script, which is being used in jedi-rapids, ported over to use the ECMWF release versions of odc and pyodc. The currently used script is based on the pre-release versions of odc and odyssey.


srherbener:
Note the test failures are because the automated testing is using a container with the pre-release versions of odc and odyssey.

src/odc/odb2nc.py
```diff
@@ -0,0 +1,88 @@
...
+
+# Find python runtime dependency paths.
+import sys
+import os
+from pathlib import Path
+prefix_lib_path = Path(__file__).absolute().parent.parent/'lib'
+prefix_bin_path = prefix_lib_path.parent/'bin'
+sys.path.append(str(prefix_lib_path))
+sys.path.append(str(prefix_lib_path/'python{}.{}'.format(*sys.version_info)/'site-packages'))
+os.environ['PATH'] = "%s:%s" % (os.environ['PATH'], str(prefix_bin_path))
```
markjolah, line 11: I think we can remove all this now that pyodc and codc are installed on the system.

srherbener:
This works on my MacBook (with the new odc and pyodc installed). I believe that the tests will continue to fail until the test containers are updated. I've put the "waiting for other repos" label on to denote this.

mmiesch:
I just tested this in the `jedi-gnu-openmpi-dev_v0.2` singularity container and with the `jedi/intel19-impi-v0.2` module on S4.  All tests pass.  As I noted before, neither of these is the default yet but we can make them so soon.  I think we're mainly waiting on the upstream atlas PR, which should be in soon.

srherbener:
@mmiesch, thanks for testing this and for the info on the containers, modules and PRs. Hopefully, the atlas PR will get merged soon.

buildspec_clang.yml
```diff
@@ -41,6 +41,8 @@ phases:
   build:
     commands:
       - cd /build_container
+      - echo $PYTHONPATH
+      - export PYTHONPATH=/usr/local/lib
```
mmiesch, line 45: This is already set when I run the docker containers manually but CodeBuild seems to run things a little differently.  Weird, since the whole idea of a container is portability.  Anyway, I'm ok with adding this if it fixes things.  But, maybe remove the preceding echo statement or move it to after it is set.

mer-a-o: I suspected that this could be the issue. But it looks like PYTHONPATH is set correctly as you mentioned. I cannot really reproduce these errors in containers outside CodeBuild...
I don't see xarray installed in clang and intel19 containers. Do you remember if you have installed it for these containers?

mer-a-o:
@srherbener and @mmiesch , I noticed that in Intel container ODC tests don't get built (there are 26 tests and 30 in the other two containers). ODC is installed in intel19 container so I am not sure why this is happening. I was able to reproduce this issue in the intel19 container outside CodeBuild environment.

srherbener:
@mer-a-o, @mmiesch, I just tried using the Intel19 Charliecloud container on my iMac and I'm getting all 30 tests to run and pass. I did, however, have to install Python xarray in my user area to get this to work. Looks like xarray is missing from the container.

mer-a-o:
@srherbener and @mmiesch, I think we can ignore the test failures related to xarray and finding shared libraries in ioda-conventers for now. @mmiesch will add xarray to the containers later this week and I will debug the issue in a separate PR. What do you think? 
Also @srherbener, I tried `pip install xarray` in intel19 docker container (not on codebuild) but still got 26 tests. How did you install xarray in the container?

srherbener:
@mer-a-o, for installing xarray, I used ``pip install --user xarray`` which installs in your home directory instead of the system directory (which is probably read only).

The ODC tests only get included when the ODC package is found. The top-level CMakeLists.txt has a ``find_package(odc)`` command. I'm guessing that the find_package is failing in tests that you are doing. On my iMac I've got odc_PATH and odc_ROOT set in my environment. I used jedi-stack to build and install odc libraries, and modules to set the environment. Perhaps these are not set, or set incorrectly when you are seeing only 26 tests.

I think it's okay to merge this in and debug the xarray/library issues in a subsequent PR. Should we go with that?

mer-a-o:
@srherbener, You are right, probably it cannot find odc. I will set odc_PATH and odc_ROOT in the container and test. I guess we can merge this now.

mmiesch:
I just tried running both the gnu and the intel19 docker containers.  After installing xarray, I got 30 tests to pass on both.   I'm not seeing the problem outside of CodeBuild.  Is `/usr/local/lib` in the PYTHONPATH in CodeBuild?

PR #268, opened 2020-04-09
Fix unicode encoding errors for station id

markjolah:
This hopefully fixes `UnicodeEncodeError` and `UnicodeDencodeError` for station ids in some conventional obs produced during the GSI NCDIAGS conversion process.

Honestly, this particular part of the code this patch addresses is a mess and whatever is happening is likely wrong or very inefficient or both, and will need to be eventually refactored to deal with strings in a sane manner.   For now, I'm crossing my fingers that this patch works and isn't papering over any more serious problems.  At least it is now not crashing.


srherbener:
@emilyhcliu, would you be able to review this soon? Thanks!

srherbener:
@emilyhcliu Thanks! I'll merge now.

PR #280, opened 2020-05-29
remove azimuth angle QC in convertor

hailingz:
The azimuth angle QC check is removed in GNSSRO IODA convertor because operators should check the azimuth angle when needed. Actually azimuth angles are only required in 2D operators.  RO data without azimuth angle should not be rejected in data processing.

https://github.com/jcsda/aop20/issues/58


hailingz:
@srherbener 
The Travis CI log shows iodaconv_marine_coding_norms failed. Not sure how to fix.  Thanks,

hailingz:
@srherbener 

Thank you so much for your review and explanation.

srherbener:
@hailingz, I merged in develop, which contains the fix for the marine coding norms test failure. The CI testing issues have been fixed. We should see the tests pass now.

markjolah:
OK with me.

PR #284, opened 2020-06-10
fix issue with test failures in auto tests

mer-a-o:


mer-a-o:
@mmiesch, @srherbener:
I updated intel19 container and it solved the issue with finding odc. After installing xarray in CodeBuild and Travis-CI all tests pass except `test_iodaconv_odb2nc_tool` in CodeBuild GNU container (not in Travis-CI GNU container). The paths look correct:
`PATH = /usr/local/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/codebuild/user/bin`
`PYTHONPATH  = /usr/local/lib`
`LD_LIBRARY_PATH = /usr/local/lib:`

But the test fails with this error: 
```
OSError: cannot load library 'libodccore.so': libodccore.so: cannot open shared object file: No such file or directory.  Additionally, ctypes.util.find_library() did not manage to locate a library called 'libodccore.so'
```
Any idea on how to solve this issue?

srherbener:
@mer-a-o, not sure what's going on. I tried this on my Mac using the charliecloud intel19 container (dated Jun 8) and it works. The file /usr/local/lib/libodccore.so exists, which is where it should be. Looks like pyodc 1.0.2 is installed, which is correct.

The charliecloud container has python3.6.  I tried importing ctypes and there is not a util subpackage. Perhaps the python version is not right, or the ctypes package is not right. My ctypes version is 1.1.0.

srherbener:
Sorry, tested the wrong container. Retried this using the Singularity gnu-openmpi container dated June 8. Same results. All tests pass, libodccore.so is in /usr/local/lib, pyodc is 1.0.2, python 3.6.

mer-a-o:
@srherbener, yes everything passes inside the container but in CodeBuild the paths are somehow reset and I need to specify them again. We don't have this problem in Travis-CI so everything passes there. Is there any other path that I need to set other than `PATH`, `LD_LIBRARY_PATH`, and `PYTHONPATH` ?

mer-a-o:
The issue is finally solved! I had to set a user (jcsdauser) to be able to run mpi jobs in the gnu container. In CodeBuild, I had to set the paths for jcsduser in the same command line as ctest.

mmiesch:
Great work @mer-a-o!

.travis.yml
```diff
@@ -56,6 +56,7 @@ script:
   - docker exec jcsda_container ls
   - docker exec jcsda_container ls /jcsda/src_repo
   - docker exec jcsda_container bash -c 'cp -r /jcsda/.openmpi/ ~/'
+  - docker exec jcsda_container bash -c 'pip install xarray'
```
mmiesch, line 59: Needed for now but we need to remember to remove this when we move to the new containers

buildspec_clang.yml
```diff
@@ -14,6 +14,10 @@ phases:
       - echo $CODEBUILD_WEBHOOK_ACTOR_ACCOUNT_ID
       - echo $CODEBUILD_WEBHOOK_EVENT
       - echo $CODEBUILD_RESOLVED_SOURCE_VERSION
+      - pip install xarray
+      - echo $PATH
+      - echo $LD_LIBRARY_PATH
+      - echo $PYTHONPATH
```
mmiesch, line 20: Do you want to leave these echos in or clean up?  I'm ok either way - leaving them in might be useful in future debugging.

buildspec_gnu.yml
```diff
@@ -96,6 +106,8 @@ phases:
       - if [ "$BUILD_STATUS" = "1" ] && [ "$CODEBUILD_BUILD_SUCCEEDING" = "0" ];
```
rickgrubin, line 106: CodeBuild defines:

`CODEBUILD_BUILD_SUCCEEDING: Whether the current build is succeeding. Set to 0 if the build is failing, or 1 if the build is succeeding.`

Suboptimal that it's opposite of what POSIX defines.

PR #293, opened 2020-06-25
Remove MetaVars which have empty data in ncdiags from lists in combine_conv.py

markjolah:
This fixes a bug where if the `MetaVarData` is empty for some meta variables, the concatenation process will break.  Instead of keeping around all the broken metavariables, I am deleting those that are empty/erronous.  This seems to fix the problem with the `2020061506_obs.tar` and later gsi ncdiags conversion.  I am having some difficulty fully testing due to unrelated updates with fv3-bundle which are preventing a jedi-rapids build from completing.  I am about 90% sure this should be working, but let's hold off on merge until I get the jedi-rapids build working again.

### DIFF
--- a/src/gsi-ncdiag/combine_conv.py
+++ b/src/gsi-ncdiag/combine_conv.py
@@ -89,7 +89,7 @@ def concat_ioda(FileList, OutFile, GeoDir):
             ncf.close()
     # extract metadata and generate a numpy array
     MetaVarData = []
-    MetaVarUnique = []
+    bad_idxs = []
     for v in MetaVars:
         tmpvardata = []
         if MetaInAll[v]:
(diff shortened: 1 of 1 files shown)


markjolah:
Apparently this is breaking the tests, so I will look into this.  Any insights are welcome.  The suspect data is from 20200615_06 window center and beyond.

srherbener:
@markjolah, the test failure was a one-liner coding norms issue so I went ahead and fixed it.

markjolah:
OK.  Hold off for now.  I've got my testing system working and there is still and issue.  Will push an update soon.

srherbener:
@markjolah, looks like the test failures are due to coding norms errors.

markjolah:
@srherbener OK. I fixed that coding norms, and it looks like the new obs conversion process is working and HofX will run on the output, so I am ready for merge now.

srherbener:
@markjolah, sounds good. I just tried on my iMac and it builds and all tests pass. I'll merge now. Thanks!

PR #300, opened 2020-07-24
Add Python IODA-converter for TROPOMI NO2 retrievals

CoryMartin-NOAA:
First version of IODA-converter for TROPOMI NO2 retrievals from NESDIS/KNMI.

Will be used to support FV3-LAM-CMAQ regional NO2 air quality assimilation using JEDI.

This will need to be revisited once 2D variables are supported as the 3D averaging kernel from the data provider is currently broken up into nlocs x 33 different variables for each TM5 level, instead of nlocs x nlevs.

Information on how to use this product and description of input netCDF files available:
https://sentinel.esa.int/documents/247904/2474726/Sentinel-5P-Level-2-Product-User-Manual-Nitrogen-Dioxide


YoulongXia-NOAA:
Could you please replace flatten( ) with ravel ()? The reason is that ravel is faster than flatten() as it does not occupy any memory.

YoulongXia-NOAA:
I went through your all 21 commits and did not find any existing problem.  You can wait for others and see if they find any issues.

srherbener:
Could some of the reviewers please take a look at this. I need two approvals before I can merge. Thanks!

markjolah:
I get these deprecation warnings:
```
./tropomi_no2_nc2ioda.py:59: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
  AttrData['date_time_string'] = ncd.getncattr('time_reference')[0:19]+'Z'
./tropomi_no2_nc2ioda.py:60: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
  AttrData['sensor'] = ncd.getncattr('sensor')
./tropomi_no2_nc2ioda.py:61: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
  AttrData['platform'] = ncd.getncattr('platform')
```

Also, I am curious if there is a larger dataset (on S3) to test performance with?

CoryMartin-NOAA:
@markjolah I've never seen this before from the netCDF4 module. I don't know if it's something we can fix or not without modifying the netCDF4 Python source.

There are large datasets available here to test: https://s5phub.copernicus.eu/dhus/#/home Any single file of NO2 retrievals should work in the converter.

CoryMartin-NOAA:
@markjolah great! Yes that was my experience too, only a few seconds on Hera to run a full sized file. It alleviated some of my concerns about needing to move away from Python towards compiled code for the time being.

flampouris:
@CoryMartin-NOAA <https://github.com/CoryMartin-NOAA> beautifully done.

PR #304, opened 2020-08-11
Performance improvement for NcWriter DateTime conversions

markjolah:
While testing performance of @CoryMartin-NOAA 's new `tropomi_no2_nc2ioda.py` in PR #300, I noticed that the majority of the time was spent in `NcWriter::ConvertToTimeOffset` which parses datetime strings in ioda format and converts to the decimal hour float32 offset used in the `Time@MetaData`.  Since this was bottlenecking the whole converter there was not much sense in trying to optimize the tropomi converter itself any further until this was addressed.  

Now I think this PR should remove the bottleneck and allow further performance profiling and improvements.

Percent of total runtime spent in `NcWriter::ConvertToTimeOffset` (with 45MB NetCDF datasource):
* Original: `2.67 s / 3.38 s = 79%` tropomi_orig_ConvertToTimeOffset
* Improved:  `0.284 s / 1.16 s = 24%` tropomi_improved_ConvertToTimeOffset

Overall runtime without profiler overhead 
* Original: 2.78 s 
* Improved: 0.996 s 
* **Speedup factor: 2.8** 

Strategy:
Knowing that the dates are almost always going to be all the same or at most two different dates, we compute and cache the date time offsets based on the bytes representation of the first 10 chars.  Next, we convert only the time part of the string to datetime.time and compute the offset from 0000Z on the given date.  Along with a pre-computed offset for the time part of the reference date, we can compute the overall offset.  

**Notes:** 
 * The `fromisoformat()` functions are much faster than `strptime` because they are less generic, but they match our format for the date and time parts individually.
 * It's important to round offsets to integer seconds and then convert to decimal hours, to match the implementation of `timedelta.total_seconds()` and exactly reproduce the reference behavior.
 * I also tried a few methods of computing the offset directly from converted digits of the time and all methods were less efficient than this version. [Edit: this method was actually faster and also python-3.6 compatible, so is implemented in the final version]
 * I have tested a few other converters and saw only modest speedup.  It seems bottlenecks in other scripts are elsewhere, but I will test with the whole ncdiags pipeline to see if there is any noticeable effect.

### DIFF
--- a/src/lib-python/ioda_conv_ncio.py
+++ b/src/lib-python/ioda_conv_ncio.py
@@ -187,14 +187,14 @@ class NcWriter(object):
         ref_offset = math.floor(self._ref_date_time.hour*3600. + self._ref_date_time.minute*60. + self._ref_date_time.second)/3600.  # decimal hours
         date_offset_cache = {}  # Cache a map from bytes[0:10] to the corresponding date's offset from reference date
         for i in range(len(datestr_vec)):
-            date_str = datestr_vec[i][0:10].tobytes()
-            date_offset = date_offset_cache.get(date_str)
+            date_bytes = datestr_vec[i].tobytes()
+            date_offset = date_offset_cache.get(date_bytes[0:10])
             if date_offset is None:
                 # There are at most 2 distinct dates per cycle for any window length < 24hrs
-                date_offset = (dt.date.fromisoformat(date_str.decode('ascii')) - self._ref_date_time.date()).total_seconds()/3600.
-                date_offset_cache[date_str] = date_offset
-            t = dt.time.fromisoformat(datestr_vec[i][11:19].tobytes().decode('ascii'))
-            t_offset = math.floor(t.hour*3600. + t.minute*60. + t.second)/3600.
+                delta = dt.date(int(date_bytes[0:4]), int(date_bytes[5:7]), int(date_bytes[8:10])) - self._ref_date_time.date()
+                date_offset = delta.total_seconds()/3600.
+                date_offset_cache[date_bytes[0:10]] = date_offset
+            t_offset = (int(date_bytes[11:13])*3600. + int(date_bytes[14:16])*60. + int(date_bytes[17:19]))/3600.
             offset[i] = date_offset + (t_offset - ref_offset)
         return offset
 
--- a/src/lib-python/ioda_conv_ncio.py
+++ b/src/lib-python/ioda_conv_ncio.py
@@ -175,21 +175,28 @@ class NcWriter(object):
 
         return IodaDtype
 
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA:
@srherbener am I wrong, or was there some discussion before about not even using the time@MetaData anymore in IODA and only using the string timestamps? If that's the case, can we remove this from the converter API?

markjolah:
Note.  Apparently there is no `.fromisoformat()` in Python 3.6.  I will have to figure out a workaround for the older python support.

Not sure about what the overall plans for the variables... I actually prefer the format of `time@MetaData`, as it is more efficient in both space and time.  Working with strings is always slower, and we we're running into that as a bottleneck with plotting HofX output.

CoryMartin-NOAA:
The problem is we don't want the time to be stored as reference to some value. GSI does this and it's fine for now, in reference to analysis time, but long term the actual time should be stored. However, the Unix epoch time might be a good solution to use in time@MetaData? But this is probably going to have to depend on IODA-engines.

markjolah:
Yeah, I definitely see there are +/- for each option.  From a scientific and user-practicality standpoint I think the textual representation is the best, and that's why it will end up being the final choice.   However, for performance, the floating point representation will always be way better and by about a factor of 10-1000 depending on the situation.   We'll just have to work to get that closer to the lower end of that range with tricks like employed here.

markjolah:
Ok. Fixed with an even more efficient version that has no python version problems.  Now the time in `NcWriter::ConvertToTimeOffset` is down to: `0.211 s / 0.973 s = 21.7%` of total and the overall speedup is a factor of 2.9 for the 45mb dataset.  Similar code might be useful in other areas as profiling determines.

markjolah:
Here's the new profile:
tropomi_improved2_ConvertToTimeOffset

srherbener:
@CoryMartin-NOAA, you are correct that the Reference/Offset style of datetime is going away. This code is a leftover from when we only could handle the Reference/Offset style in the IODA reader, and were transitioning to handling the ISO 8601 formatted strings in the IODA reader. That has been done, so this step in the converters can be deleted.

The plan is to move forward with absolute datetime information in ISO 8601 formatted strings. If we run into performance issues with strings, we may want to consider using something like two integers (YYYYMMDD, hhmmss for example) instead. The target is to get the datetime information into an oops DateTime object so we can add a new constructor if we decide to change the way we store datetime info. The nice thing about ISO 8601 strings is that they are an accepted standard, readable and easily understood.

This speedup modification is great work, but I would like to caution that we are in a time where the ioda converters are going to change (transition to the new interface) and the probability of optimizing code that is going to be thrown away is higher than usual. I recommend that we wait until after the migration to the new interface to continue with performance optimization.

PR #309, opened 2020-08-19
Bugfix/pyiodaconv installdir try2

markjolah:
@srherbener and @CoryMartin-NOAA, you guys are too fast.  Thanks for the quick merge on #308 but I just found some other locations I overlooked and wasn't able to push the updates before the PR got merged.  

These changes are on the same lines as before.  I also changed the installs to `bin/` to use `CMAKE_INSTALL_BINDIR` for symmetry and correctness, but this always evaluates to `bin/` on all the platforms we use anyways.

To check that it is working it is necessary to call ecbuild with an install prefix and do make install.
```
cd <ioda-converters>
ecbuild -H. -B_build -DCMAKE_INSTALL_PREFIX=_install
cd _build
make -j8 install
```
There should be a `bin` and either a `lib` or a `lib64` (but not both) depending on your platform under `_install` prefix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,8 +94,9 @@ endif()
 set( IODACONV_PYLINT_CFG_DIR ${CMAKE_CURRENT_SOURCE_DIR} )
 
 # Location of installed python iodaconv libraries
+include(GNUInstallDirs)
 set( PYIODACONV_BUILD_LIBDIR   ${CMAKE_BINARY_DIR}/lib/pyiodaconv )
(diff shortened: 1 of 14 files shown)


markjolah:
@srherbener Hold off on merging this until I can confirm.  I may have found another bug.

srherbener:
@markjolah okay, I'll wait for the merge. Is the "waiting for another PR" label signifying that you are still checking for bugs?

Thanks for doing these fixes!

markjolah:
Will close this as a more comprehensive approach was necessary.  See #310

PR #310, opened 2020-08-19
Feature/jedicmake

markjolah:
This is a comprehensive update that removes ecbuild as a required dependency, and allows either jedicmake or ecbuild to be used to provide FindNetCDF.cmake.  Overall the goal of this update is that everything should work the same way as before.  All the executables, libraries, tests, and files should be the same.  What has changed is the implementation.  Major notes:
* Replace the `@SCRIPT_LIB_DIR@` template logic for locating the libray with an implementation using relative paths.  
  * This allows the whole prefix to be relocated as no absolute paths are written
  * Scripts are templated to be executable from either, the source dir, the build dir, or the install prefix.  The correct relative paths are used in each scenario
  * This eliminates the need to name files with `.in` suffix which confuses python editors.
* Replace `add_custom_target()` logic with `execute_process()` logic, which has the same effect but simpler.
* Replace `ecbuild_add_test()` with standard `add_test()` function
* Remove obsolete and unused ecbuild features and options
* Installs `lib-python` under `${CMAKE_INSTALL_LIBDIR}` which allows for portable installs
* Provides a `iodaconv-config.cmake` which defines:
  * Variables:
    * iodaconv_VERSION - Version number
    * iodaconv_BINDIR - Path to bin dir (adapted for either build- or install-prefix)
    * iodaconv_PYTHON_PATH - Path to iodaconv python libraries `lib-python` (adapted for either build- or install-prefix)
  * Components: 
    * gnssro
     * odb2
     * odc

While this still falls short of the ultimate goal of a full python setuptools configuration with a `setup.py` based installation, this will give us a solid stepping-stong along the way.  It should solve most of the current portability problems, and removes unnecessary ecbuild baggage.

This version seems to be working for me so far.  As always it's difficult to test these build-system updates, so please let me know if something is no longer working.  I will test on a few more systems tomorrow to see if I can find any remaining issues, then ask for reviewers more widely.


markjolah:
Ok.  I think I've got all the bugs worked out and am ready for testing/review.  Again, everything should work the same as before, except the install will use the `lib` or `lib64` directory depending on the system.  Also you should now be able to build with either the `cmake` or `ecbuild` executables.  If you build with CMake and not ecbuild, ecbuild will only be added if `jedicmake` could not be found.  This latest version also has user friendly messages to indicate which optional components are enabled or disabled.

PR #311, opened 2020-08-25
Feature/ncep bufr

markjolah:
This PR must go along with [`jcsda/bufrlib` # 11](https://github.com/JCSDA/bufrlib/pull/11), and also [jedi-stack # 137](https://github.com/JCSDA/jedi-stack/pull/137).  Please hold off on merging until the those PRs go through and we have the new bufrlib version available on the HPCs / containers.

The net effect is to transparently allow either [`jcsda/bufrlib`](https://github.com/JCSDA/bufrlib) or [`NOAA-EMC/NCEPLIBS-bufr`](https://github.com/NOAA-EMC/NCEPLIBS-bufr) to be used to build the gnssro bufr converters.  If either package is available on the system the build should work the same.  

The idea is that it was easiest to match the specific target name `bufr::bufr_4_DA` which the NCEPLIBS CMake has defined.  Now the new version of `jcsda/bufrlib` will export an alias target with that name which will point to the compatible shared library if available or the static library otherwise.

To test this with either repository version it's necessary to install that package to a temporary install prefix directory `_install`.  Then provide that path as `-Dbufr_ROOT=<path-to-install-dir>`.  For example:

```
cd <jcsda-bufrlib>
cmake -H. -B_build -DCMAKE_INSTALL_PREFIX=_install
cd _build && make install
cd <ioda-converters>
cmake -H. -B_build -DCMAKE_INSTALL_PREFIX=_install -Dbufr_ROOT=<jcsda-bufrlib>/_install
cd _build && make install && ctest
```

The same procedure should work for `NCEPLIBS-bufr` also.


markjolah:
I thought probably this should fail CI tests as the library name has changed from `bufrlib`->`bufr` to match what NCEPLIBS-bufr uses.  Inside the current CI container, the build will not find the new name for the library.  However, `bufr` is an optional dependency and the build is set to disable the gnssro converter if it couldn't find bufr.  When this happens there is no actual error, so the CI passes, but the build is still broken until the installed bufr is updated.

markjolah:
This branch is now working on S4 with the `jedi/intel-impi` C++14 modules.  In fact it now won't build the gnssro converters without this change.  It should also be working with NECPLIBS-bufr if you have `bufr_ROOT` point to the right place or have a module built with that repo.

srherbener:
@markjolah, just want to double check. This has the ready for merge label, but it also says to wait for JCSDA/jedi-stack#137. Should I merge now, or wait for JCSDA/jedi-stack#137?

Thanks!

markjolah:
@srherbener Well this is now required to build on S4, but it seems that @mmiesch is still updating the orion stack, so I guess let's discuss the status tomorrow at the meeting.

PR #318, opened 2020-08-31
Remove obsoleted odb2 and odc converters and tools

### LINKED ISSUE #317: Remove obsolete odb/odc converters and tools
The current scripts in src/ODB2 and src/odc are based on an experimental odb file layout of which we have decided not to use. Instead we will be using the UKMO/ECMWF file layout, and the Met Office will be contributing a converter/reader/writer for their layout in the near future.

Because of this, we need to get rid of the converters and scripts based on the experimental layout.

srherbener:
This PR removes obsoleted converters and tools which were based on an experimental odb file layout. We've decided to not use the experimental file layout and instead use the UKMO/ECMWF file layout instead. The Met Office will be contributing a converter/reader/writer for their file layout.

Some developers have expressed the desire to refer back to these tools, so I created a tag called "exp-odb-tools" for easy retrieval.

Closes #317


smnewman (approved):
Thanks for tag "exp-odb-tools" so we can refer back to these tools if necessary

PR #319, opened 2020-08-31
Fix giirs_ssec install dir paths

markjolah:
Replace the `IODA_CONV_PATH` logic with the new version used in the other python files.  This will allow the script to locate the lib paths at install-time when run from pwd from outside the script's bin directory.


markjolah:
I think we can merge this.  I have another PR which will build depend on this one.

srherbener:
@weihanucar it looks like @markjolah has another PR that depends on this one so it would be good to get your review soon. Thanks!

markjolah:
Closed by #322 which supersceeds this PR.

PR #322, opened 2020-09-03
Feature giirs ssec performance

markjolah:
This is a performance analysis and optimization for the new `giirs_ssec2ioda.py` converter.  The GIIRS converter reads and collates the observations from a set of several thousand individual scan files for a typical 6Hr window.   

Initial testing showed a single 6-hr window took approximately 60+ mins to collate 1200 obs files and convert to IODA and displayed a non-linear scaling with the number of files processed.  Profiling short runs shows the overwhelming majority of time is spent processing the data read from the files in the `readInFiles` method.  Detailed line profiling identified the majority of the time was spent in the for loops converting radiance units to brightness temperature.

 giirs_ssec_orig

These calculations can be vectorized by refactoring the constant terms from the radiance conversion equation and expanding the conversion as an element-wise matrix manipulation rather than as nested for loops.  

After eliminating the radiance conversion overhead, the remaining time can be attributed mainly with the strategy of incrementally re-allocating an `NcVector` for each meta-variable and for each data variable for each input file.  This strategy is quadratic and working with `NcVectors` directly is often 100x slower than numpy arrays.  I modified the strategy with the knowledge that the maximum observations are `128*numfiles`, we can pre-allocate ordinary numpy arrays and matrices for the observations, datetimes, and metadata, etc.  Then any excess space is removed at the end.

Also note:
* Constant vectors `K1` and `K3` are for the radiance conversion and are independent of the obs files and so are moved outside the main loop to avoid re-computation of any terms not necessary.
* The brightness_temperatures for one entire scan file are computed directly into the corresponding slice of the `obs_vals` matrix which has shape `nChans X nTotalLocs`.
* We use a `mask` to mask out the unselected obs.  This controls the total number of obs per file (which is at most 128), and how many columns of `obs_vals` and other metavars are written for each file processed.
* For obs that pass the first `mask`, but fail to have a valid radiance there is a second `valid = (lwRadiance_apo > 0) & (lwRadiance_apo < 300)` mask which controls if the radiance conversion is applied to that loc or not.
* The radiance conversion uses a `log(1+x)` in the original formulation.   This was replaced with the numpy function `np.log1p()` which is faster and more numerically stable especially for small absolute values of `x` which may lead to numerical underflow and loss of precision in the original calculation.
* Radiance computation intermediates are now all at float32 precision.

After applying optimizations, the time for `readInFiles` over a 20 file test set has decreased from `99.0s -> 0.441s` and from 99% of runtime to 20%.  With the remainder of the time spent writing the ioda file in `ioda_conv_ncio.py`.  The resulting file is ~1.2Gb per 6hr period, so approximately 1.5s to write a file this big seems at least reasonable.  I did not explore any further optimization into those routines.

giirs_ssec_improved

Overall runtime for a 236 file test set (approx 1hr windows) with 27251 obs:
* **Original**: 370.451s
* **Improved**: 3.499s
* **Speedup**: 106x

Note: after re-arranging the radiance conversions some of the values changed by up to `5.0x10^-5` in relative magnitude, which while small, was enough to cause the tests to fail.   I had to update the reference file with the new values.  I expect these differences are because all the intermediate calculations are now done at 32-bit precisions, and the change to the `np.log1p()` function for the logarithm, which will change precision slightly.  

This is the version that is currently running on S4 and displayed on [nrt.jcsda.org](http://nrt.jcsda.org).

@weihanucar  I would appreciate if you could help review these changes and validate the radiance conversion is still valid and I didn't miss any details.


src/ssec/giirs_ssec2ioda.py
```diff
@@ -98,40 +110,57 @@ def readInFiles(self, filenames):
         qcKey = self.VAR_NAME, self.writer.OqcName()
 
         numFiles = 0
+        totalFiles = 0
         self.writer._nlocs = 0
 
         sat_zen_max = 74
         detector_number = 128
-        detector_array = np.arange(1, detector_number+1, dtype='i4')
+        detector_array = np.arange(1, detector_number+1, dtype=np.float32)
```
srherbener, line 118: Why the type change from int to float?  It looks like this is used to create the scan_position@MetaData variable, which I would have expected to be an integer. @weihanucar, is scan_position@MetaData supposed to be an integer or a float?

srherbener (commented):
Thanks for improving the performance of this converter. The results are impressive!

I just have the one comment.

markjolah:
Note these are the differences as seen in the output when using the optimized version.  All changes are about at the 9th sig fig, which is related to float32 rounding precision.

```
22: Test command: /home/molah/rapids-home/build/bundle/nrt/iodaconv/_build/bin/iodaconv_comp.sh "netcdf" "/home/molah/rapids-home/build/bundle/nrt/iodaconv/_build/bin/giirs_ssec2ioda.py
             -i testinput/giirs_ssec.nc
             -o testrun/giirs_ssec_ioda.nc4
             -d 2020080100" "giirs_ssec_ioda.nc4"
22: Test timeout computed to be: 1500
22: [GIIRS2IODA] Processing 1/1 #locs:128 -- giirs_ssec.nc
22: [GIIRS2IODA] Processed 128 total locs from 1/1 valid files.
22: DIFFER : VARIABLE : brightness_temperature_115@ObsValue : POSITION : [25] : VALUES : 279.799 <> 279.799
22: DIFFER : VARIABLE : brightness_temperature_118@ObsValue : POSITION : [66] : VALUES : 248.646 <> 248.646
22: DIFFER : VARIABLE : brightness_temperature_363@ObsValue : POSITION : [22] : VALUES : 287.556 <> 287.556
22: DIFFER : VARIABLE : brightness_temperature_373@ObsValue : POSITION : [16] : VALUES : 256.633 <> 256.633
22: DIFFER : VARIABLE : brightness_temperature_397@ObsValue : POSITION : [36] : VALUES : 261.598 <> 261.598
22: DIFFER : VARIABLE : brightness_temperature_467@ObsValue : POSITION : [43] : VALUES : 241.366 <> 241.366
22: DIFFER : VARIABLE : brightness_temperature_566@ObsValue : POSITION : [112] : VALUES : 255.3 <> 255.3
22: DIFFER : VARIABLE : brightness_temperature_656@ObsValue : POSITION : [105] : VALUES : 238.191 <> 238.191
22: DIFFER : VARIABLE : brightness_temperature_662@ObsValue : POSITION : [74] : VALUES : 242.429 <> 242.429
22: Variable                            Group Count          Sum      AbsSum          Min          Max Range         Mean StdDev
22: brightness_temperature_115@ObsValue /         1  3.05176e-05 3.05176e-05  3.05176e-05  3.05176e-05     0  3.05176e-05      0
22: brightness_temperature_118@ObsValue /         1 -3.05176e-05 3.05176e-05 -3.05176e-05 -3.05176e-05     0 -3.05176e-05      0
22: brightness_temperature_363@ObsValue /         1 -3.05176e-05 3.05176e-05 -3.05176e-05 -3.05176e-05     0 -3.05176e-05      0
22: brightness_temperature_373@ObsValue /         1 -3.05176e-05 3.05176e-05 -3.05176e-05 -3.05176e-05     0 -3.05176e-05      0
22: brightness_temperature_397@ObsValue /         1 -3.05176e-05 3.05176e-05 -3.05176e-05 -3.05176e-05     0 -3.05176e-05      0
22: brightness_temperature_467@ObsValue /         1 -1.52588e-05 1.52588e-05 -1.52588e-05 -1.52588e-05     0 -1.52588e-05      0
22: brightness_temperature_566@ObsValue /         1 -3.05176e-05 3.05176e-05 -3.05176e-05 -3.05176e-05     0 -3.05176e-05      0
22: brightness_temperature_656@ObsValue /         1  1.52588e-05 1.52588e-05  1.52588e-05  1.52588e-05     0  1.52588e-05      0
22: brightness_temperature_662@ObsValue /         1 -1.52588e-05 1.52588e-05 -1.52588e-05 -1.52588e-05     0 -1.52588e-05      0
```

src/ssec/giirs_ssec2ioda.py
markjolah: Yes, thanks for catching that.  Is addressed with new changes.

markjolah:
Since this has still not been reviewed, I cleaned up some more items to improve the testing and flexibility, and to fix several bugs I have encountered in production use.  Notable changes:
* Use original `giirs-fy4a-test.nc` file instead of new file I had added.  The original test file has fewer detectors and channels, which makes the test data significantly smaller.  This accomplishes one of my objectives for the testing-code sprint.
* New class name `Giirs2Ioda` is more informative.  Remove references to `longwave_rad` which is not used by the UFO/CRTM operator for GIIRS.
* Allow converter to determine: num detectors, num channels, and wavenumbers from input data files rather than hard-coding them.  This change necessitates moving the construction of the data structures inside the main for loop over data files.  On reading the first successful file, we initialize numpy data structures large enough to hold all remaining data and metadata.  Also, we pre-compute the radiance to brightness temp conversion constants which are identical for each file.
* Check that the assumption that all files in the sequence have the same number of channels and detectors, and throw a RuntimeError if violated.
* Allow for missing metadata names when the input files don't have them. This happens with the test data which is missing a `Cloud_Fraction` variable but this is optional, so we don't want to fail.
* Check for critical missing metadata name: 'IRLW_SatelliteZenith' and drop data files without this variable.  Normally, these are corrupted or empty files.
* Promote the `SAT_ZEN_MAX` filtering constant to a class-level constant.
* Ensure the `scan_position` is an integer and is set correctly for the number of detectors.

weihanucar (commented):
Very impressive on the optimization !

markjolah:
Very interesting.  Fixing the `scan_position@MetaData` to be an `int32` causes the HofX to crash which was the source of the recent failures on the NRT website.  I will revert to a `float32` in the feature/nrt branch until the issue can be resolved.  I will link that issue here when ready.

markjolah:
See linked issue: [ufo # 1285](https://github.com/JCSDA/ufo/issues/1285)

markjolah:
@weihanucar If you approve of the PR, you need to click "start a review" and click "approve" otherwise your response without approval causes github to think you've asked a question.

weihanucar:
There is a bug in the python converter script:
self.xferLocMdataVar_detector(ncd, numFiles, detector_array*1.0, 'scan_position', LocMdata, 'float', mask)
It should be :
self.xferLocMdataVar_detector(ncd, numFiles, detector_array, 'scan_position', LocMdata, 'integer', mask)

weihanucar:
Hi Mark,

There is a bug in the python converter script:
self.xferLocMdataVar_detector(ncd, numFiles, detector_array*1.0, 'scan_position', LocMdata, 'float', mask)
It should be :
self.xferLocMdataVar_detector(ncd, numFiles, detector_array, 'scan_position', LocMdata, 'integer', mask)




Wei
 
From: Mark J Olah
Date: 2020-09-17 09:21
To: JCSDA/ioda-converters
CC: weihanucar; Mention
Subject: Re: [JCSDA/ioda-converters] Feature giirs ssec performance (#322)
Very interesting. Fixing the scan_position@MetaData to be an int32 causes the HofX to crash which was the source of the recent failures on the NRT website. I will revert to a float32 in the feature/nrt branch until the issue can be resolved. I will link that issue here when ready.

src/ssec/giirs_ssec2ioda.py
markjolah: It turns out that this change leads to a runtime failure in ufo/crtm.  See [ufo # 1285](https://github.com/JCSDA/ufo/issues/1285).

markjolah:
@weihanucar Yes, as you saw we discovered that issue after I updated the PR based on reviewers' comments.  The UFO/CRTM interface somehow requires `scan_postion@MetaData` to be `float32` or it cannot be found.  I've reverted back to using a float32 for scan_position, which I think should solve the problem, and is working for the NRT website now.

weihanucar (approved):

srherbener:
The insistence of the `scan_position@MetaData` to be a float is a design flaw in the current ObsSpace implementation. The container for the data (ObsSpaceContainer) is templated and if you call get_db() to read data from the ObsSpace using the float signature (which is probably what is being done in UFO), it looks in the float version of ObsSpaceContainer. This works if you make `scan_position@MetaData` in the ioda file a float (so that is get loaded into the float version of ObsSpaceContainer).

Changing `scan_position@MetaData` to an int in the file caused it to get loaded into the int version of ObsSapceContainer of which the get_db() with the float signature cannot find. It would also work to change both the ioda file and the UFO get_db() call (to the int signature), but I think we should wait until the new re-factored ObsSpace is in place before doing this. The new ObsSpace will give the caller the ability to query the type of the data and then call the appropriate templated read function.

srherbener (approved):
Sorry about my review comment about the int vs float with `scan_position@MetaData` and the confusion it caused. Ultimately, I think this should be an int, but due to the issue with UFO I think it should remain a float until we get the new re-factored ObsSpace in place.

PR #324, opened 2020-09-12
Feature/deps

### LINKED ISSUE #313: Cmake update for `feature/bufr-converter`
This package no longer is ecbuild-based.   The `feature/bufr-converter` is a C++ package which was depending on ecbuild and will need to be refactored to work in standard CMake.

aerorahul:
## Description

This PR enables the import of ioda-engines, jedi_base, oops and eckit.
These components are required for the bufr converter.

ioda-bundle branch: `feature/bufr-converter` [here](https://github.com/JCSDA/ioda-bundle/tree/feature/bufr-converter)

What bug does it fix, or what feature does it add? No bugs. Simplifications and enhancements.
Is a change of answers expected from this PR? NO

### Issue(s) addressed
closes #313 

Waiting on the following PRs:
None

## Impact

Requires changes in the following repositories:
None

Requires updating AWS test data for the following repositories:
None


markjolah:
@aerorahul Thanks for this update.  It sounds like from your email that this PR is replaced by the feature/bufr-converter-2 branch?  Let me know which one I should review.

aerorahul:
> @aerorahul Thanks for this update. It sounds like from your email that this PR is replaced by the feature/bufr-converter-2 branch? Let me know which one I should review.

@markjolah This sets the stage to build with ioda-engines.
You should review this PR. 

This PR adds a lot more dependencies, For CI, we need to include these optional dependencies in the container/travis-CI framework.

When the bufr-converter-2 is ready, we will include that in another PR.

CMakeLists.txt
```diff
@@ -9,17 +9,13 @@
 
 cmake_minimum_required( VERSION 3.12 )
 
-set(iodaconv_major 0)
-set(iodaconv_minor 0)
-set(iodaconv_revision 1)
-set(iodaconv_version ${iodaconv_major}.${iodaconv_minor}.${iodaconv_revision})
-project( iodaconv VERSION ${iodaconv_version} LANGUAGES Fortran )
+project( iodaconv VERSION 0.0.1 LANGUAGES C CXX Fortran )
```
markjolah, line 12: I don't think we need `C` language compiler, just `CXX`, right?

CMakeLists.txt
```diff
@@ -9,17 +9,13 @@
...
 
 ## Configuration
 include(GNUInstallDirs)
 include(CTest)
 
-set( CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake;${CMAKE_MODULE_PATH} )
+list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
```
markjolah, line 18: This changes the order of the module search path.  I would either stick with the original code or do `list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_CURRENT_SOURCE_DIR}/cmake )`

CMakeLists.txt
```diff
@@ -50,8 +46,13 @@ else()
     include(ecbuild_log) #Required for ecbuild/FindNetCDF.cmake
 endif()
 
+find_package(eckit QUIET)
+find_package(oops QUIET)
```
markjolah, line 50: If I understood the purpose of `jedi-base` was to remove the need for `oops`?

CMakeLists.txt
```diff
@@ -68,23 +69,42 @@ else()
...
+    endif()
+    if(NOT bufr_FOUND)
+        message(STATUS "NOT-Found: bufr - BUFR converter disabled")
+    endif()
+    message(STATUS "Disabled Component: bufr")
+endif()
+
+
+include( iodaconv_compiler_flags )
+include_directories( ${IODACONV_INCLUDE_DIRS} ${IODACONV_EXTRA_INCLUDE_DIRS} )
```
markjolah, line 102: We are moving away from using this method of setting compiler flags, which is a holdover from ecbuild.  If you require a specific option use `target_compile_options()` or `target_compile_features()`, and if necessary use a conditional against `CMAKE_<LANG>_COMPILER_ID` and `CMAKE_<LANG>_COMPILER_Version`.  See for example how we do this for the GNSSRO converter regarding gfortran-10 specific flags.  If there are system specific flags required they should go into the system-specific toolchain under `jedicmake`.

src/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+add_subdirectory(wrfda-ncdiag)
+
+# Optional components
+if(iodaconv_gnssro_ENABLED)
+    add_subdirectory(gnssro)
+endif()
+
+if(iodaconv_bufr_ENABLED)
+  add_subdirectory(bufr)
+endif()
```
markjolah, line 16: Can we move these to the root CMakeLists.txt?  I don't see the advantages of adding another file to manage which doesn't actually define any variables or do anything other than include other files.  It just makes it more complicated to see what is being src'd.

src/bufr/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+#                    )
+
+#target_link_libraries(ingester PRIVATE oops::oops)
+#target_link_libraries(ingester PRIVATE jedi_base::jedi)
+#target_link_libraries(ingester PRIVATE ioda_engines::ioda_engines)
+
+#add_executable(test_ingester.x test_ingester.cpp)
+#target_link_libraries(test_ingester.x PRIVATE ingester)
+
+ecbuild_add_test( TARGET iodaconv_bufr_coding_norms
```
markjolah, line 31: This can be a normal CMake `add_test()`.

tools/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 list( APPEND programs
     ${PROJECT_NAME}_lint.sh
     ${PROJECT_NAME}_comp.sh
+    ${PROJECT_NAME}_cpplint.py
```
markjolah, line 4: Can we possibly add this to jedicmake.  At the moment, every repo is contributing its own version.  I can submit a PR to jedicmake to export this maybe?  I guess then it would require jedicmake to run this test?  Or we could just go ahead and make jedicmake a required dependency?

markjolah (changes requested):
I guess this doesn't build the actual executable for the bufr-converter yet, so I was unable to test.  You will want something with a normal `add_executable()` command and then add the dependencies with `target_link_libraries()`.

CMakeLists.txt
markjolah: Also there is no need for `include_directories()` commands.  Those should be handled by the [`INTERFACE_INCLUDE_DIRECTORIES` ]( `https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html`) target property of each lineked target dependency.

CMakeLists.txt
aerorahul: Ok.

CMakeLists.txt
aerorahul: Ok.

CMakeLists.txt
aerorahul: When jedi-base has those, we can remove oops.

CMakeLists.txt
aerorahul: I don't agree with removing `iodaconv_compiler_flags`.
Compiler flags pertaining to a repository are stored there. oops does this, soca, fv3jedi, ufo all do this.
Compiler flags pertaining to a platform/system are handled at the toolchain level e.g. AVX2 instruction flags, SIMD flags etc.

If an executable/library needs specific flags, they can be added at the level of the source code and not in the location of iodaconv_compiler_flags.
For e.g. debug, release, bit flags for the project should reside in iodaconv_compiler_flags.
executable specific flags. e.g. -fconvert=big-endian when dealing with IO of binary files belong with the application.

The gfortran-10 flags in gnssro should be in `iodaconv_compiler_flags` to avoid repetition of this code everywhere a fortran source code is compiled.

aerorahul: I am fine with removing `include_directories`

src/CMakeLists.txt
aerorahul: This keeps the top-level CMakeLists.txt clean.

Infact, the checks of enabling bufr or gnssro based on packages found or not in the top-level CMakelists belong at the level of src where those directories are going to be added.

tools/CMakeLists.txt
aerorahul: I copied this from oops. I believe UFO, IODA have their own copies.
We can add this to jedicmake and make it a required repo.  However this is not cmake related and thus not really a jedicmake thing.

src/bufr/CMakeLists.txt
aerorahul: Ok.

tools/CMakeLists.txt
rhoneyager: I forget the reason why we split cpplint per repository. It was pre-jedicmake. The versions have diverged: https://github.com/JCSDA/oops/commits/develop/tools/cpplint.py, https://github.com/JCSDA/ufo/commits/develop/tools/cpplint.py.

There was an old ecbuild issue where the wrong cpplint.py was being run, which is why we settled on different names.

CMakeLists.txt
rhoneyager: Looks like jedi-base was not selected for release. So, use oops for now.

aerorahul: Thanks for the confirmation @rhoneyager

CMakeLists.txt
markjolah: Actually we've been discussing this in the core JEDI meetings for some time, and it is a current goal to remove this way of setting CFLAGS.  Dan has already removed them from [fv3-jedi/cmake](https://github.com/JCSDA/fv3-jedi/tree/develop/cmake) and the rest of the repositories will eventually move that way too.  Also all the ecmwf packages have already removed them.  They preventing the use of the built-in CFLAGS handling of ecbuild and CMake.  

The documentation for how ecbuild is designed to use CFLAGS in the comments of [ecbuild_compiler_flags.cmake](https://github.com/JCSDA/ecbuild/blob/release-stable/cmake/ecbuild_compiler_flags.cmake).

>  The procedure is as follows:
> 
>  1.  ecBuild does **not** set ``CMAKE_<lang>_FLAGS`` i.e. the user can set
>      these via ``-D`` or the CMake cache and these will be the "base" flags.
> 
>  2.  ecBuild **overwrites** ``CMAKE_<lang>_FLAGS_<btype>`` in the CMake cache
>      for all build types with compiler specific defaults for the currently
>      loaded compiler i.e. any value set by the user via ``-D`` or the CMake
>      cache **has no effect**.
> 
>  3.  Any value the user provides via ``ECBUILD_<lang>_FLAGS`` or
>      ``ECBUILD_<lang>_FLAGS_<btype>`` **overrides** the corresponding
>      ``CMAKE_<lang>_FLAGS`` or ``CMAKE_<lang>_FLAGS_<btype>`` **without being
>      written to the CMake cache**.

We want to move to using toolchains for specifying the jedi-system-wide specialized debug flags.  The problem is that changing a single debug flag requires modifying every single repository with the old way of setting flags, and because it doesn't follow the ecbuild mechanisms for setting flags, we cannot override these compiler flags easily with a toolchain.
 
When there are target specific flags that must only be applied for debug builds, they should be added with a generator expression  `target_compile_options(tgt PRIVATE $<$<CONFIG:Debug>:-my-special-debug-flag>)`.  Setting flags like this allows the distinction between PRIVATE, PUBLIC, and INTERFACE inheritance of imported target flags.

src/CMakeLists.txt
markjolah: It's a matter of style, so whatever is prefered.  I would rather have everything I'm interested in a single file.  

The reason to not source a directory I'm not interested in is a general strategy to reduce CMake configuration overhead, which is already very significant for large bundles.  Every extra file sourced is actually multiple filesystem calls, which take forever on some systems (S4).

tools/CMakeLists.txt
markjolah: We'll what's most important is that the cpplint.py doesn't have any dependencies so adding a module to jedicmake to provide it is easy, and will reduce duplication and confusion.  We want to bump common stuff up to that level.  Eventually all the "extra macros" will be unified and added there too to lessen the maintenance burden on each repo to maintain that infrastructure.

rhoneyager (approved):

tools/CMakeLists.txt
aerorahul: > We'll what's most important is that the cpplint.py doesn't have any dependencies so adding a module to jedicmake to provide it is easy, and will reduce duplication and confusion. We want to bump common stuff up to that level. Eventually all the "extra macros" will be unified and added there too to lessen the maintenance burden on each repo to maintain that infrastructure.

Ok. So what do you want me to do here? 
Should I wait until this is added to `jedicmake`?

CMakeLists.txt
aerorahul: @markjolah 

I agree that target specific flags should be applied on the target itself.  However, where do you set general repo flags such as `-g -traceback`? Are these flags at toolchain level? If so, where is the toolchain repo, and I can add that to the repo.

There is no use of `ecbuild_compiler_flags` macro here. It is a simple include of `iodaconv_compiler_flags` that sets the compiler ID specific flags and for this repo alone.

I am following logic from the OOPS repository. OOPS still has `oops_compiler_flags`.
Where do you discuss these items? I don't think I heard anything to this tune in the JEDI weekly round table discussions.

markjolah (approved):

PR #331, opened 2020-09-22
Add BUFR Ingester.

rmclaren:
## Description

Adds code module that can create IODA data from BUFR data when given a YAML file as an input (see example in test/testinput/bufrtest_bufrdescription.yaml). 
Adds tests for MHS and AMSUA bufr files.
Other tests and test data will be added in a future PR.

### Issue(s) addressed
closes #350

## Dependencies

None

## Impact

None

PR #334, opened 2020-09-28
Feature/bufr ext

rmclaren:
## Description

- Adds support for transforms to BUFR exports.
- Fixes bug were in memory (ObsSpace backend) creation of numerical value would fail because of a type mismatch (used floats instead of doubles)
- Fixed bug were in memory (ObsSpace backend) creation of std::string variables would fail (worked around ioda-engines bug (zenhub::ioda-engines:140 )).
- Changed name of main executable to toIoda.x to accommodate the fact that it might do more than just BUFR.
- Added backend parameter to YAML file so that the user can explicitly set either "inmemory" or "netcdf" backends.

## Dependencies

- waiting on JCSDA/ioda-converter/pull/331

### DIFF
--- a/src/bufr/BufrParser/Exports/Transforms/OffsetTransform.cpp
+++ b/src/bufr/BufrParser/Exports/Transforms/OffsetTransform.cpp
@@ -12,9 +12,9 @@ namespace Ingester
     {
     }
 
-    void OffsetTransform::apply(std::shared_ptr<ArrayDataObject> array)
+    void OffsetTransform::apply(IngesterArray& array)
     {
-        array = array->get().get() + offset_;
+        array = array + offset_;
     }
 
 }  // namespace Ingester
--- a/src/bufr/BufrParser/Exports/Transforms/OffsetTransform.cpp
+++ b/src/bufr/BufrParser/Exports/Transforms/OffsetTransform.cpp
@@ -1,0 +1,20 @@
+//
+// Created by Ronald McLaren on 9/24/20.
+//
+
+#include "OffsetTransform.h"
+
+
+namespace Ingester
+{
+    OffsetTransform::OffsetTransform(double offset) :
+      offset_(offset)
+    {
+    }
+
+    void OffsetTransform::apply(std::shared_ptr<ArrayDataObject> array)
+    {
+        array = array->get().get() + offset_;
+    }
+
+}  // namespace Ingester
--- a/src/bufr/BufrParser/Exports/Transforms/OffsetTransform.h
+++ b/src/bufr/BufrParser/Exports/Transforms/OffsetTransform.h
@@ -1,0 +1,23 @@
+//
+// Created by Ronald McLaren on 9/24/20.
+//
+
+#pragma once
+
+#include "Transform.h"
+
+
+namespace Ingester
+{
+    class OffsetTransform : public Transform
+    {
+     public:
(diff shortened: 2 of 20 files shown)


aerorahul (approved):
I reviewed these changes with respect to PR #331 and they are good.

aerorahul:
when #331 is merged into `develop`,  please rebase so we don't see all the same commits in the merge log.

src/bufr/BufrParser/Exports/Transforms/OffsetTransform.cpp
```diff
@@ -0,0 +1,20 @@
+//
+// Created by Ronald McLaren on 9/24/20.
```
rhoneyager, line 2: Copyright statement needed.

src/bufr/BufrParser/Exports/Transforms/OffsetTransform.h
```diff
@@ -0,0 +1,23 @@
+//
+// Created by Ronald McLaren on 9/24/20.
```
rhoneyager, line 2: Same

rhoneyager (commented):
Copyright statements need to follow standards.

src/bufr/CMakeLists.txt
```diff
@@ -60,8 +67,8 @@ add_library( ${PROJECT_NAME}::ingester ALIAS ingester )
 add_executable(test_ingester.x test_ingester.cpp)
 target_link_libraries(test_ingester.x PRIVATE ${PROJECT_NAME}::ingester)
 
-add_executable(convertBufrToIoda.x main.cpp)
-target_link_libraries(convertBufrToIoda.x PRIVATE ${PROJECT_NAME}::ingester)
+add_executable(toIoda.x main.cpp)
+target_link_libraries(toIoda.x PRIVATE ${PROJECT_NAME}::ingester)
```
CoryMartin-NOAA, line 71: toIoda.x is not explicit enough IMO. Can we at least make it Convert2Ioda.x or something like that?

src/bufr/IodaEncoder/IodaDescription.cpp
```diff
@@ -89,4 +111,20 @@ namespace Ingester
     {
         variables_.push_back(variable);
     }
+
+    void IodaDescription::setBackend(std::string backend)
+    {
+        if (backend == "netcdf")
```
CoryMartin-NOAA, line 117: if the code supports "inmemory" and "in-memory", can we throw in here "netCDF" and "NetCDF" just for safety's sake?

CoryMartin-NOAA: Maybe any case variation of netCDF?

src/bufr/BufrParser/Exports/Transforms/OffsetTransform.h
rmclaren: OBE (Overcome by Events) in a later branch.

src/bufr/CMakeLists.txt
rmclaren: updated to convertToIoda.x

src/bufr/IodaEncoder/IodaDescription.cpp
rmclaren: Backend is now case insensitive.

PR #336, opened 2020-10-02
Feature/bufr yaml updates

rmclaren:
## Description

This PR includes requested modifications to YAML files. Yaml files now consist of a list of "observations" where each "observation" consists of an "obs space" "ioda" pair. "obs space" specifies the input parser and its properties and "ioda" specifies the output encoder and its properties. In this way one YAML file can specify many re-encodings. "obs space" can be extended to support many different kinds of input parsers (ex: marine and others), but the encoder and the general strategy for mapping fields from on data type to another stay the same.

The PR also adds more sophisticated management of Parsers within the framework. It adds a ParserFactory which is responsible for instantiating new Parser objects by their name. 

## Dependencies

Waiting on the following PRs:
- waiting on JCSDA/ioda-converter/331
- waiting on JCSDA/ioda-converter/334


aerorahul (approved):
I reviewed this PR with respect to #331 and #334 and the changes are good.

src/bufr/main.cpp
```diff
@@ -21,62 +21,56 @@
...
+    void registerParsers()
+    {
+        ParserFactory::registerParser<BufrParser>("bufr");
+    }
 }  // namespace Ingester
 
+
 int main(int argc, char **argv)
 {
     if (argc < 1)
```
rhoneyager, line 67: See #331

src/bufr/ParserFactory.h
```diff
@@ -0,0 +1,83 @@
...
+#include <memory>
+#include <string>
+#include <ostream>
+
+#include "eckit/config/LocalConfiguration.h"
+#include "eckit/exception/Exceptions.h"
+
+#include "Parser.h"
+
+static const char* PARSER_NAME = "name";
```
rhoneyager, line 20: Move inside the functions. Otherwise, all code that includes this header file adds in the PARSER_NAME symbol.

rhoneyager (commented):
Small changes. Feel free to merge into the base PR once addressed.

CoryMartin-NOAA (approved):

PR #341, opened 2020-10-13
marineglider_converter

lliu1220:
## Description

This PR reads in marine glider data (temperature and salinity profile) from AOML and convert to ioda format, output data were tested in JEDI/SOCA
Note Glider data and in-situ data are of different format , it might make sense to have them separate pre-processed

Command to run (testing data included in branch):

python glider_S2ioda.py.in -i SG547_0001_dn_AOML.nc -o test_glider_S.nc -d 2018041500
python glider_T2ioda.py.in -i SG547_0001_dn_AOML.nc -o test_glider_T.nc -d 2018041500

working on data tanking and quality control for next step, since this is just one data sample.

Here is a plot of example (above) glider observation and 3dvar (glider DA only) result:
image

Ling

### DIFF
--- a/src/marine/glider_S2ioda.py
+++ b/src/marine/glider_S2ioda.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+
+#
+# (C) Copyright 2019 UCAR
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# 
+
+from __future__ import print_function
+import sys
+import argparse
+import netCDF4 as nc
+from datetime import datetime, timedelta
+import numpy as np
+import numpy.matlib
+sys.path.append("@SCRIPT_LIB_PATH@")
+import ioda_conv_ncio as iconv
+from orddicts import DefaultOrderedDict
+
+vName = "sea_water_salinity"
+
+locationKeyList = [
+    ("latitude", "float"),
+    ("longitude", "float"),
+    ("depth", "float"),
+    ("datetime", "string")
+]
+
+AttrData = {
+    'odb_version': 1,
+}
+
+class Profile(object):
+
+    def __init__(self, filename, date, writer):
+        self.filename = filename
+        self.date = date
+        self.data = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
+        self.writer = writer
+        self._read()
+
+    def _read(self):
+        ncd = nc.Dataset(self.filename)
+        dpth = (ncd.variables['ctd_depth'][:])
+        lons = np.float32(ncd.variables['longitude'][:])
+        lats = np.float32(ncd.variables['latitude'][:])
+        hrs = np.float32(np.linspace(1,23,58))
+        vals = np.float32(ncd.variables['salinity'][:])
+        errs = np.float32(np.matlib.repmat(2.258986,len(lons) , 1)) 
+        qcs = ncd.variables['salinity_qc'][:]-1 # to unify the QC
+        errs = np.squeeze(errs)
+        ncd.close()
+
+        base_date = datetime(2018,4,15) #example SG547_0001_dn_AOML.nc
+        for i in range(len(hrs)):
+            if qcs[i] != 0:
+                continue
+            valKey = vName, self.writer.OvalName()
+            errKey = vName, self.writer.OerrName()
+            qcKey = vName, self.writer.OqcName()
+            dt = base_date + timedelta(hours=float(hrs[i]))
+            locKey = lats[i], lons[i], dpth[i], dt.strftime(
+                "%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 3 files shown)


src/marine/glider_S2ioda.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
+
+#
+# (C) Copyright 2019 UCAR
```
guillaumevernieres, line 4: 2019-2020

guillaumevernieres (changes requested):
Thanks for doing this. 
A bit of work requiered:
- Remove all the references to hgodas and update to something that is meaningful to this work
- Consolidate the 2 scripts into one. Passing T or S as argument is up to you and would depend on how T&S are collected. Are they always colocated? If so you could just output 1 single file containing T&S, possibly masked with a missing value flag ...
- Your obs error assignment is questionable
- Add a ctest
- Add @srherbener as a reviewer
- I don't think this repo uses git-lfs anymore, question to @srherbener .

Next step would be to add the glider obs as part of the soca-science ctests.

src/marine/glider_S2ioda.py
lliu1220: @guillaumevernieres 
1. will work on this, thanks for checking on what was neglected in the scripts.
2. in-situ T and S has profile identifier whereas AOML glider data don't, only named "temperature", "salinity", thought was easier to write T S separately at my first shot, but I will work on merging together
3. I agree but glider data do not have obs error, I assigned constant error , I could calculate standard deviation and assign if you like
4. will do
5. already done
6. waiting on @srherbener comment

srherbener:
Sorry about the slow response. This repository is still using git-lfs. It's only the ioda repository that has switched to AWS S3. I believe the plan is to eventually move all test data to AWS, but for now it's okay to put test files in git-lfs for this repository.

srherbener (commented):
I agree with @guillaumevernieres. It would be good to consolidate into one script, plus it would be good to add a ctest.

The test failure is due to coding style errors. In your local clone's build area you can run:

```
ctest -VV -R iodaconv_marine_coding_norms
```

which will show you what to fix in your script.

If you do decide to consolidate into one script, I'd recommend fixing the coding style errors after you consolidate (less work to do). Note that being able to avoid fixing the same issue due to copies of the same code is two scripts is one of the reasons for encouraging the consolidation of these two scripts into one.

lliu1220:
Just to tread lightly, obsSpace in yaml file does have insitutemperature and insituSalinity read in separately, also hgodas_sst2ioda.py.in and smap_sss2ioda.py.in only does temperature or salinity individually. I just wonder if I could only update glider_T2ioda.py just like smap and hgodas did . Thank you.

 @guillaumevernieres. @srherbener @jkbk2004
    - ObsSpace:
        name: InsituTemperature
        ObsDataOut: {obsfile: ./Data/prof.T.out.nc}
        ObsDataIn:  {obsfile: ./Data/prof_T.nc}
        simulate:   {variables: [ sea_water_temperature ]}
      ObsOperator:
        name: InsituTemperature
      Covariance:
        covariance: diagonal
      ObsFilters:
      - *land_mask
      - Filter: Background Check
        threshold: 5

    - ObsSpace:
        name: InsituSalinity
        ObsDataOut: {obsfile: ./Data/prof.S.out.nc}
        ObsDataIn:  {obsfile: ./Data/prof_glider.nc}
        simulate:   {variables: [ sea_water_salinity ]}
      ObsOperator:
        name: MarineVertInterp
      Covariance:
        covariance: diagonal
      ObsFilters:
      - *land_mask

srherbener:
@lliu1220 I agree that these two scripts are producing ioda files that will go into two distinct ObsSpace objects. Plus I anticipate that holding the salinity and temperature in two distinct ObsSpace objects will not change. In other words there is going to be a need to produce two different ioda files.

Right now the algorithms for generating the temperature and salinity ioda files are nearly identical. In other words the two scripts are nearly identical. I think it makes sense to consolidate the two scripts into one using a command line option to select temperature versus salinity. The reason for this is that maintaining the script over time will be much easier. What you want to avoid is having a lot of duplicated code, which brings up situations such as finding and fixing a defect in one copy and forgetting to do the same in the other copy.

If over time, it becomes the case that the algorithms for generating the temperature and salinity significantly diverge, at that point I think it might make sense to have two scripts. This situation would not have a lot of duplicated code so the maintenance issues associated with duplicated code would be diminished.

guillaumevernieres:
@jkbk2004 and @lliu1220 , this work is simply not ready to review. I think @lliu1220 knows what needs to be done,  please close this PR and resubmit when ready.

lliu1220:
@srherbener @guillaumevernieres @jkbk2004 
We agree on run time command line option to have only one script for the AOML glider data.
we are going to close this PR and create a new draft PR here when the run time option is ready and an issue under soca-science with dependency and testing illustrated.

srherbener:
@lliu1220, sounds good! Thanks for taking care of this.

PR #343, opened 2020-10-15
Fix for using GSI-ncdiag converter with standard versions of GSI

CoryMartin-NOAA:
The modified version of GSI to produce GeoVaLs uses "Input_Observation_Error" rather than "Inverse_Observation_Error".

For those who just want to produce IODA obs files using GSI, a standard version of GSI will only have Inverse_Obs_Error.

This PR will default to Input, and if it doesn't exist, use Inverse (and compute the inverse of it).

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -1105,7 +1105,10 @@ class Radiances(BaseGSI):
                     varDict[vbc]['bcpKey'] = vbc, writer.ObiaspredName()
                     ibc += 1
         obsdata = self.var('Observation')
-        obserr = self.var('Input_Observation_Error')
+        try:
+            obserr = self.var('Input_Observation_Error')
+        except IndexError:
+            obserr = 1./self.var('Inverse_Observation_Error')
         obsqc = self.var('QC_Flag').astype(int)
         if (ObsBias):
             nametbc = [
@@ -1622,7 +1625,10 @@ class Ozone(BaseGSI):
         varDict[vname]['qcKey'] = vname, writer.OqcName()
 
         obsdata = self.var('Observation')
-        tmp = self.var('Input_Observation_Error')
+        try:
+            tmp = self.var('Input_Observation_Error')
+        except IndexError:
+            tmp = 1./self.var('Inverse_Observation_Error')
         tmp[tmp < self.EPSILON] = 0
         obserr = tmp
         obserr[np.isinf(obserr)] = self.FLOAT_FILL


rhoneyager (approved):

XuLi-NOAA (commented):
Cory,

I did a test with your update, it went through, but there was an issue as follows:
===========================================================
 test_gsidiag.py -i /scratch2/NCEPDEV/stmp3/Xu.Li/diag/data/v16rt2/diag_avhrr_n19_ges.2019111800.nc4 -o /scratch2/NCEPDEV/stmp3/Xu.Li/obsout -t rad
Input Parameters: ObsBias=None QCVars=None TestRefs=None
/scratch1/NCEPDEV/da/Xu.Li/git/ioda-converters/src/gsi-ncdiag/gsi_ncdiag.py:1111: RuntimeWarning: divide by zero encountered in true_divide
  obserr = 1./self.var('Inverse_Observation_Error')
Satellite radiance obs processed, wrote to: /scratch2/NCEPDEV/stmp3/Xu.Li/obsout/avhrr_n19_obs_2019111800.nc4
==================================================
:divide by zero encountered in true_divide" needs to be handled, I think.

I tested with a iasi diag file and it showed the same RuntimeWarning.

CoryMartin-NOAA:
@XuLi-NOAA what do you propose as a solution?
If Inverse_Observation_Error for certain observations is 0, 1/0 creates this runtime error which makes obserror=NaN. Should these obserror values be set to zero?

XuLi-NOAA:
We know, Inverse_Observation_Error=0 means these observations are rejected (not used). I don't have a solution to it (setting to be zero is not an option). If it can be distinguished from others (used ones), then obserror=NaN is ok.

XuLi-NOAA:
Cory,

One idea, 

In the modified GSI, Input_Observation_Error is used, instead of Inverse_Observation_Error.
How Input_Observation_Error is defined/assigned? I saw "double error_variance(nchans)" in the nc diag file.
Input_Observation_Error is related to error_variance in the modified GSI?

CoryMartin-NOAA:
@XuLi-NOAA sngl(error0(ich_diag(i))
https://github.com/CoryMartin-NOAA/GSI/blob/a7fdbbdc811c625ca5d984f81fb4f290fb5266a9/src/gsi/setuprad.f90#L2361

srherbener:
@CoryMartin-NOAA, @XuLi-NOAA if Inverse_Observation_Error is zero mean that the obs is rejected, would it make sense to either put a missing value mark in the ObsValue, or a setting in PreQC that is recognized as reject by a filter?

XuLi-NOAA:
In the standard GSI, Inverse_Observation_Error is included:

call nc_diag_metadata("Inverse_Observation_Error",             sngl(errinv)          )

Here, errinv = sqrt(varinv0(ich_diag(i))), and varinv0 is the weight given for each datum, it is assigned to be 1/error0**2 and then adjusted during QC

In the modified GSI, Input_Observation_Error is included, 

call nc_diag_metadata("Input_Observation_Error", sngl(error0(ich_diag(i)))                ) 

Here, error0 = varch(n), n is the channel number, read from satinfo fix file.

In both standard and modified GSI, error_variance is included

call nc_diag_chaninfo("error_variance",  varch(n)                        )

In summary, error_variance is identical to Input_Observation_Error , which is varch(n), the initial observation error before QC and any other adjustment. And Inverse_Observation_Error is the weight, which is adjusted during QC procedure. Therefore, the modified GSI  doesn't make sense, we should keep the standard GSI.

Note: 
        errinv = sqrt(varinv0(ich_diag(i)))
        varinv0 = varinv occurs after the QC tests, that means although errinv = 1/error0 initially, errinv /= 1/error0 for those data experienced weight adjustment in QC tests (for the extreme case, the weight is assigned to zero: varinv=0, but the corresponding obs error adjustment is not followed).

XuLi-NOAA:
It needs to be confirmed with a diag file sample generated by the modified GSI to see if

Input_Observation_Error = error_variance

guillaumevernieres:
Thanks for doing this @CoryMartin-NOAA , why not just assign a large number where inverse sigo is zero? This can be revisited by @XuLi-NOAA in the future if need be.

CoryMartin-NOAA:
@XuLi-NOAA thanks for pointing this out, @emilyhcliu and I plan to re-make a "modified" GSI to produce GeoVaLs from the GFSv16 version of GSI soon.

@XuLi-NOAA does this mean that this pull request is good? If you need to generate IODA obs using GSI this should work and Emily and I will correct / confirm this when we re-make the modified GSI branch.

XuLi-NOAA:
There must be more changes/updates in the "modified GSI", and in terms of "error0 " (Input_Observation_Error ) or "errinv" (Inverse_Observation_Error)  in diag file, clearly errinv is better since it represents the datum dependent weight, and error0 (Input_Observation_Error ) is identical to varch (error_variance). 
I understand a new "modified" GSI to produce GeoVaLs from the GFSv16 version of GSI will be produced soon. I assume the "errinv" will be used there?

As to this pull request, your fix is Ok, but it is better to set Input_Observation_Error to a huge value (instead of NaN) while Inverse_Observation_Error is zero, as Guillaume said.

XuLi-NOAA (approved):
Ready to go

srherbener (approved):
The current ioda reader is tolerant of "inf" and "nan" values in the file. It will convert both of these to the float JEDI missing value.

In the future we want to make "inf" and "nan" values in the file an error, so please keep that in mind for the solution to handle the inverse error equal to zero cases.

Thanks!

PR #344, opened 2020-10-16
Bugfix for pressure units

CoryMartin-NOAA:
There is a mismatch between standard GSI (hPa) and our "modified" GSI (Pa) in units of pressure, both for use as a coordinate (e.g. radiosondes) and for surface pressure obs.

This will check to see if the fields for GeoVaLs exist, and if so, do no unit conversion, otherwise it will convert from hPa to Pa.

Once @emilyhcliu and I have a new GSI GeoVaL generator, we will make sure we 1) remove the unit conversions in the source code and 2) modify the ioda-converters to handle all unit conversions.

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -694,6 +694,11 @@ class Conv(BaseGSI):
 
                 for o in range(len(outvars)):
                     obsdata = self.var(conv_gsivarnames[v][o])[idx]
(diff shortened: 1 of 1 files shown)


danholdaway (approved):

danholdaway:
Thanks @CoryMartin-NOAA. This works great for our GEOS ncdiag data. I have only tested with data that has hPa but looks good.

srherbener (approved):
Thanks for fixing this!

PR #348, opened 2020-10-24
Fix GIIRS IODA metadata initalization if not all variables are in first file

markjolah:
## Description

The `giirs_ssec2ioda.py` executable reads in a sequence of GIIRS NetCDF data files and accumulates the combined data into a single IODA file.  The first GIIRS file read in is used to initialize the metadata variables that will be sent to the `NcWriter` object.

Some metadata variables are optional and don't occurr in every GIIRS file, primarily the `Cloud_Fraction` variable.  If the first file read in does not contain `Cloud_Fraction` or any other optional variable, the output numpy array is never initalized and the conversion will error out if a subseqent file then has this variable.

This PR seperates the metadata variable initalization into a private method and calls it as needed if a new variable is discovered in a later file.

This is a blocker for the NRT GFS website workflow, so I am moving the active workflow to point to this branch.  Please don't delete this branch after merging until I can move the NRT rapids-bundle back to develop sometime after next week.

### DIFF
--- a/src/ssec/giirs_ssec2ioda.py
+++ b/src/ssec/giirs_ssec2ioda.py
@@ -188,7 +188,7 @@ class Giirs2Ioda:
                 # Read metadata
                 for new_key, old_key in self.LOC_MDATA_MAP.items():
                     if old_key in ncd.variables:
-                        if not new_key in self.LocMdata:
+                        if new_key not in self.LocMdata:
                             # If we haven't seen this key before initialize it now and set to fillvalue for all previous locs
                             self._initialize_metadata_key(new_key, old_key, ncd)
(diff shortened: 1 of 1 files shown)


weihanucar (approved):

srherbener (approved):
I fixed a coding style error which was needed to get the CI tests to pass. It's on line 191 of giirs_ssec2ioda.py. The old line was:

`if not old_key in sefl.LocMdata:`

and the new line is:

`if old_key not in self.LocMdata:`

I don't think the ctests cover this new code, but I think that the two lines are equivalent logic. I'm not sure why the code style checker even cares about this, but it might be flagging something that is better to fix.

I'm going to approve and merge in. I'll leave the branch around until you can adjust the NRT flow

PR #354, opened 2020-11-03
Added relative tolerance spec to nccmp calls.

### LINKED ISSUE #353: GIIRS test is failing on iMac platform
The `test_iodaconv_giirs_ssec2ioda` test fails with the following messages from nccmp:

```
DIFFER : VARIABLE : brightness_temperature_4@ObsValue : POSITION : [6] : VALUES : 219.741 <> 219.741 : PERCENT : 1.3888e-05
DIFFER : VARIABLE : brightness_temperature_9@ObsValue : POSITION : [7] : VALUES : 226.292 <> 226.292 : PERCENT : 1.34859e-05
DIFFER : VARIABLE : brightness_temperature_10@ObsValue : POSITION : [2] : VALUES : 229.168 <> 229.168 : PERCENT : 1.33167e-05
DIFFER : VARIABLE : brightness_temperature_13@ObsValue : POSITION : [1] : VALUES : 231.1 <> 231.1 : PERCENT : 1.32054e-05
DIFFER : VARIABLE : brightness_temperature_14@ObsValue : POSITION : [7] : VALUES : 232.624 <> 232.624 : PERCENT : 1.31189e-05
DIFFER : VARIABLE : brightness_temperature_17@ObsValue : POSITION : [7] : VALUES : 235.51 <> 235.51 : PERCENT : 1.29581e-05
Variable                           Group Count          Sum      AbsSum          Min          Max Range         Mean StdDev
brightness_temperature_4@ObsValue  /         1  3.05176e-05 3.05176e-05  3.05176e-05  3.05176e-05     0  3.05176e-05      0
brightness_temperature_9@ObsValue  /         1  3.05176e-05 3.05176e-05  3.05176e-05  3.05176e-05     0  3.05176e-05      0
brightness_temperature_10@ObsValue /         1 -3.05176e-05 3.05176e-05 -3.05176e-05 -3.05176e-05     0 -3.05176e-05      0
brightness_temperature_13@ObsValue /         1  3.05176e-05 3.05176e-05  3.05176e-05  3.05176e-05     0  3.05176e-05      0
brightness_temperature_14@ObsValue /         1  3.05176e-05 3.05176e-05  3.05176e-05  3.05176e-05     0  3.05176e-05      0
brightness_temperature_17@ObsValue /         1 -3.05176e-05 3.05176e-05 -3.05176e-05 -3.05176e-05     0 -3.05176e-05      0
```

The differences are quite small, plus the tolerance options to nccmp are not being used. In general, a relative tolerance (-T option to nccmp) should be used to account for tiny differences when testing for matching floating point values.

srherbener:
## Description

This PR adds the ability to pass in a relative tolerance value to iodaconv_comp.sh (nccmp inside) when checking the contents of netcdf4 files. The need for this became apparent when the GIIRS test started failing on my iMac (despite passing in CI testing).

### Issue(s) addressed

- fixes #353 

## Dependencies

None

## Impact

None

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -81,6 +81,12 @@ endforeach(FILENAME)
 #  testoutput/)
 #===============================================================================
 
+# Typically, a converter is simply doing a format change (ie, copying data directly
+# from the input file). For these cases, use a tolerance of zero for the
+# iodaconv_comp.sh test.
+#
+# For a converter that is generating new data, use a non-zero tolerance.
+set(IODA_CONV_COMP_TOL_ZERO "0.0")
 set(IODA_CONV_COMP_TOL "0.5e-4")
 
 #===============================================================================
@@ -94,7 +100,7 @@ add_test(NAME test_iodaconv_gds2_sst_l2p
             -o testrun/gds2_sst_l2p.nc
             -d 2018041512
             -t 0.5"
-            gds2_sst_l2p.nc ${IODA_CONV_COMP_TOL})
+            gds2_sst_l2p.nc ${IODA_CONV_COMP_TOL_ZERO})
 
 add_test(NAME test_iodaconv_gds2_sst_l3u
          COMMAND ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh netcdf
@@ -103,7 +109,7 @@ add_test(NAME test_iodaconv_gds2_sst_l3u
             -o testrun/gds2_sst_l3u.nc
             -d 2018041512
             -t 0.5"
-            gds2_sst_l3u.nc ${IODA_CONV_COMP_TOL})
+            gds2_sst_l3u.nc ${IODA_CONV_COMP_TOL_ZERO})
 
 add_test(NAME test_iodaconv_smap_sss
          COMMAND ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh netcdf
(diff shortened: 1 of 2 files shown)


CoryMartin-NOAA (commented):
Is the tolerance the same units as the observation? What exactly is this tolerance value representing? Seems that each observation type/test/file should have a different tolerance.

aerorahul (commented):
adding tolerance is a good idea where data is being manipulated.
Is data being manipulated in these converters? If not, shouldn't the tolerance be 0 for those converters that are essentially a format change?

CoryMartin-NOAA:
@aerorahul looks like the test in question that was failing is converting radiance to brightness temperature

srherbener:
@CoryMartin-NOAA, @aerorahul thanks for the comments. I set this up to pass in the same tolerance value to each test, but it could be easily changed to pass in tolerances per test. I'm okay making that change.

Even with a format change, would there be a possibility of a least-significant-bit being different with different python packages, compilers, etc. The tests have been pretty much immune to tiny differences creeping in, so maybe not.

The -T option of nccmp is being used and specifies a percent relative tolerance. This should take into account the different units and ranges of values, correct?

If you think it makes sense to have zero tolerance for the converters that are changing format only, then I can make that change.

aerorahul:
@srherbener 
I understand the argument with numpy arrays sometimes causing differences; I have been bit with those in the past.
So far, from our testing without tolerances, on the converters we have, I have not seen that happen.  So, I am inclined to make the tolerance 0 for those.  If a difference arises, I would rather spend the time to figure out why and understand it, than to just accept a difference and move on.  I am happy to approve as it stands.

CoryMartin-NOAA (approved):

srherbener:
@aerorahul, good point, and I agree it's better to use the zero tolerance on the "format change only" converters so I changed that back.

aerorahul (approved):
Thanks @srherbener

srherbener:
There appears to be something wrong with the CodeBuild clang configuration. I see all the tests pass on my iMac in addition to the CodeBuild gnu and intel tests passing. Because of this, I'm going to merge this PR in now.

PR #357, opened 2020-11-09
Move logic enabling the GNSSRO tests back to the root CMakeLists

markjolah:
## Description

Enable GNSSRO converter tests.  The logic for setting the variable `iodaconv_gnssro_ENABLED` was moved out of the root CMakeLists.txt, which causes the `tests/CMakeLists.txt` to salwar skip the GNSSRO tests.  The same problem will occur with the logic for the `iodaconv_bufr_ENABLED` variable whenever those converters begin to have tests, so I moved that also.

### Issue(s) addressed

Enable GNSSRO converter tests.


aerorahul:
@markjolah 
Yeah, sorry.

I moved it back in `feature/bufr-converter` PR #331

aerorahul (approved):
this was a mistake on my part and has been addressed in #331 
However, there is no reason, this cannot be merged in before #331 
Thanks @markjolah

srherbener (approved):
Thanks for this fix!

PR #359, opened 2020-11-13
Add simple single observation generation script

### LINKED ISSUE #358: Simple Script to Make Single Observation File
For single observation tests.

CoryMartin-NOAA:
## Description

This PR adds a new single observation generation python script that takes an input YAML file and produces an IODA file with one observation of a prescribed type, value, location, etc.

### Issue(s) addressed

- fixes #358 
## Dependencies

None

## Impact

None

Note: to automatically run saber, ioda and ufo tests with this PR, push a commit containing "trigger pipeline", e.g.:
git commit --allow-empty -m 'trigger pipeline'


src/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ add_subdirectory(conventional)
 add_subdirectory(ncep)
 add_subdirectory(ssec)
 add_subdirectory(wrfda-ncdiag)
+add_subdirectory(pseudo)
```
aerorahul, line 11: Is `pseudo` appropriate?  It is not a pseudo observation.
Perhaps `single_observation`.

aerorahul (approved):
looks good to me.

src/CMakeLists.txt
CoryMartin-NOAA: I'm fine with 'single_observation', or maybe even 'singleob' just for brevity's sake

CoryMartin-NOAA: I will wait and see what any other reviewers say, as I don't have a strong opinion one way or another.

gthompsnJCSDA: I like the dir name of single_observation because it is quite clear.

gthompsnJCSDA: I fully support and approve the PR. It's a great idea to have this addition.  To have it work with a satellite brightness temperature, does it need potentially another variable to handle channel?

gthompsnJCSDA (approved):

src/CMakeLists.txt
CoryMartin-NOAA: I'll change to single_observation and push. As of now, I think this will work for brightness_temperature as we use the convention brightness_temperature_N for N variables. When we switch to brightness_temperature being 2D (nlocs,nchan), then yes, this will probably need to be modified (along with many other converters).

PR #363, opened 2020-11-18
Bug fix: missing temperature/dewpoint/wind in the CSV file causes code failure

gthompsnJCSDA:
## Description

This is a bug fix. When the CSV file contains a null element for some variables, the code would fail on the try clause attempting to convert to float.  What is distressing to me is why the try clause would not immediately go to the except clause and it seems to require the added step of testing for null string explicitly.  If someone knows better, please advise.

### Issue(s) addressed

One was not created for this issue, it was just fixed and a PR issued.

## Dependencies

There should be no dependencies.


CoryMartin-NOAA:
@gthompsnJCSDA it's because the except explicitly states 'csv.Error', add the exception and try will work again.

try:
  stuff
except csv.Error, NEWERROR

CoryMartin-NOAA:
If the error message was something like:
`ValueError: could not convert string to float: `

Then make the code:

```
try:
  stuff
except csv.Error, ValueError:
  whatever
```

CoryMartin-NOAA:
Or, you can do:

```
try:
  stuff
except csv.Error:
  handle this exception with whatever
except ValueError:
  handle this exception a different way
```

gthompsnJCSDA:
Cory: the idea of comma separated except clauses did not work.  It says: Invalid Syntax for that one.  So I will try your second idea, but is it really any better than my testing for null with single quotes?

CoryMartin-NOAA:
@gthompsnJCSDA whoops sorry, it is a tuple:
https://docs.python.org/3/tutorial/errors.html

Should be:
`except (csv.Error,ValueError):
`

I think it is better to check the ValueError rather than the if '' because what if it is a character introduced? If it is always going to be a blank quote then your test is fine but if it is ' ', then it fails.

gthompsnJCSDA:
> @gthompsnJCSDA whoops sorry, it is a tuple:
> https://docs.python.org/3/tutorial/errors.html
> 
> Should be:
> `except (csv.Error,ValueError): `
> 
> I think it is better to check the ValueError rather than the if '' because what if it is a character introduced? If it is always going to be a blank quote then your test is fine but if it is ' ', then it fails.


Yes, I went looking as well to find the solution and found it while you replied.  I'm fixing. Thanks!

src/conventional/metar_csv2ioda.py
```diff
@@ -87,23 +87,23 @@ def _rd_metars(self):
...
                     temp = float(row['Temp']) + self.meteo_utils.C_2_K
-                except csv.Error:
+                except (csv.Error, ValueError):
                     temp = self.float_fill
                 try:
                     dewp = float(row['Dewp']) + self.meteo_utils.C_2_K
-                except csv.Error:
-                    dewp = self.float_fill
+                except (csv.Error, ValueError):
+                        dewp = self.float_fill
```
CoryMartin-NOAA, line 99: Fix indentation here.

gthompsnJCSDA:
Thanks Cory for the help. Perhaps I have things right this time. Worth the effort as always!

CoryMartin-NOAA:
No problem at all, and absolutely. Things look good now!

CoryMartin-NOAA (approved):

srherbener (approved):
Thanks for the fix!

Issue #365, opened 2020-11-24
Split Bufr Ingester lib.


rmclaren:
Split the ingester lib into ioda_encoder and bufr_parser in order to facilitate the creation of new parsers.

huishao-r:
Removed the AOP20 Epics from the Epic link.

PR #366, opened 2020-11-24
BUFR library split.

rmclaren:
## Description

This PR splits the the ingester library into ioda_encoder and bufr_parser libraries. The namespaces have been accordingly renamed (IodaEncoder and BufrParser). Some of the files were moved around in order to enforce the idea that IodaEncoder is a framework that can be used to build additional parsers. 

### Issue(s) addressed

#365: Split Bufr Ingester lib.


srherbener (commented):
Thanks for doing this update. I think the code looks okay, but I have some comments about the organization of the split.

It seems that you are moving toward an architecture that has a set of parsers (bufr, netcdf, binary, etc.) that read their respective files and convert data to a neutral intermediate format, followed by a ioda encoder that converts the intermediate format to an ObsGroup (memory or file). It also looks like you are perhaps moving toward creating a separate library for each parser.

My apologies if I've got the wrong picture, but if this is correct it seems awkward to me to have the parser factory in the encoder library where you have to connect it up to potentially a set of libraries for the parsers. I think it might make things easier if you created one parser library that contains all the parsers and the factory. For example, putting all the parsers in one library and one namespace would facilitate organizing the parsers with class inheritance if you choose to go in that direction.

As I mentioned I'm making assumptions about where the implementation is going so I'm happy to discuss this further.

srherbener:
I took a look at the output ioda files and have a couple of comments:

- Both nlocs and nchans are defined with unlimited size. Having multiple dimensions be resizable can sometimes be slow since when you increase their sizes the data potentially needs to get shuffled around. With that said, the data shuffling can be minimized by lining up chunk sizes with increments in the dimension sizes.

  More importantly, the resize capability with the ObsStore (in-memory) backend is currently limited to only resizing along the first dimension (nlocs). If it is truly necessary to be able to resize along more than the first dimension, then we need to know so we can fix the in-memory backend to support that.

- We are trying to limit the precision of obs data (in-memory, file) to single precision as much as possible. Would it be acceptable to use floats (instead of doubles) by default?

Thanks!

rmclaren:
> Thanks for doing this update. I think the code looks okay, but I have some comments about the organization of the split.
> 
> It seems that you are moving toward an architecture that has a set of parsers (bufr, netcdf, binary, etc.) that read their respective files and convert data to a neutral intermediate format, followed by a ioda encoder that converts the intermediate format to an ObsGroup (memory or file). It also looks like you are perhaps moving toward creating a separate library for each parser.
> 
> My apologies if I've got the wrong picture, but if this is correct it seems awkward to me to have the parser factory in the encoder library where you have to connect it up to potentially a set of libraries for the parsers. I think it might make things easier if you created one parser library that contains all the parsers and the factory. For example, putting all the parsers in one library and one namespace would facilitate organizing the parsers with class inheritance if you choose to go in that direction.
> 
> As I mentioned I'm making assumptions about where the implementation is going so I'm happy to discuss this further.

rmclaren:
@srherbener

First of all sorry for all the extra email traffic, I accidentally pressed the wrong button on the GitHub GUI....

To your points:

> It also looks like you are perhaps moving toward creating a separate library for each parser.

I've been troubled by these same issues that you raised above. I would even take it a step further and say that we don't need *any* libraries. Really you could treat it like a self contained application (extending it with new parsers as needed) [lets call it **makeIoda**]. I don't really see the benefit of slicing it into libraries... It might still might make sense to have two namespaces (one for encoder and one for parsers), but even here I'm torn (is there really much benefit).

> Both nlocs and nchans are defined with unlimited size.

This is unintentional and we should probably create an issue for it.

> We are trying to limit the precision of obs data (in-memory, file) to single precision as much as possible. 

Again this is unintentional. I though I was limiting it to floats, but now I notice that the output from Panoply says otherwise. I will create an issue for this as well.

PR #368, opened 2020-12-01
Feature/ncar-bufr2nc-fortran

jamiebresch:
## Description

Add fortran code to convert PrepBUFR and GNSSRO/AMSU-A/MHS/AIRS BUFR files.
The output ioda files are being tested/used by some mpas-jedi applications.

src/ncar-bufr2nc-fortran/README.md

```
Usage: bufr2nc_fortran.x -i input_dir -o output_dir bufr_filename(s)_to_convert
```
If [-i input_dir] [-o output_dir] are not specified in the command line, the default is the current working directory.  
If [bufr_filename(s)_to_convert] is not specified in the command line, the code looks for file name, **prepbufr.bufr** (also **gnssro.bufr**, **amsua.bufr**, **airs.bufr**, **mhs.bufr**), in the input working directory. If the file exists, do the conversion, otherwise skip it.

src/ncar-bufr2nc-fortran/gnssro_mod.f90 is adapted from src/gnssro/gnssro_bufr2ioda.f90 and the gnssro output are identical using the two converters (gnssro_bufr2ioda and bufr2nc_fortran.x).

test/testinput/prepbufr.bufr contains ADPUPA reports of NCEP prepbufr.gdas.20200930.t18z.nr.48h

### Issue(s) addressed

None

## Dependencies

None

## Impact

None


src/ncar-bufr2nc-fortran/define_types_mod.f90
```diff
@@ -0,0 +1,371 @@
+module define_types_mod
+
+use kinds, only: r_kind, i_kind
+use ufo_vars_mod, only: var_ps, var_prs, var_u, var_v, var_ts, var_tv, var_q, var_tb
+use netcdf, only: nf90_float, nf90_int, nf90_char
+
+implicit none
+
+real(r_kind),    parameter :: t_kelvin          = 273.15
```
CoryMartin-NOAA, line 9: t_kelvin should be a 'constant' , not a 'type', correct?

src/ncar-bufr2nc-fortran/define_types_mod.f90
```diff
@@ -0,0 +1,371 @@
...
+      case ( 616 ); sensor = 'viirs'
+   end select
+
+end subroutine set_name_sensor
+
+subroutine set_brit_obserr(name_inst, ichan, obserr)
+
+! set brightness temperature observation errors
+
+! For now it is a temporary subroutine to assign AMSU-A and MHS observation errors
```
CoryMartin-NOAA, line 310: Assuming the permanent solution is to read in an input file, not writing the errors in the source explicitly.

src/ncar-bufr2nc-fortran/gnssro_mod.f90
```diff
@@ -0,0 +1,581 @@
+!!!--------  README --------------------------------------------------------------------
+!  this is a temporary routine to generate netcdf file
+!  for jedi/ufo/gnssro/ operator test
+!  Copyright UCAR 2019
+!  Authors: Hailing Zhang
+!           Mark Olah
+
+!!!---------  to run   -----------------------------------------------------------------
+!  ./gnssro_bufr2ioda yyyymmddhh $bufrfile_input $netcdffile_output
```
CoryMartin-NOAA, line 9: Remove this if it's a module not a program now. Does this mean we can remove the old converter then?

src/ncar-bufr2nc-fortran/gnssro_mod.f90
```diff
@@ -0,0 +1,581 @@
...
+character(80) hdr1a
+
+data hdr1a / 'YEAR MNTH DAYS HOUR MINU PCCF ELRC SAID PTID GEODU SCLF' / 
+data nemo /'QFRO'/ 
+nrec =0
+ndata=0
+nvars=2
+maxobs=0
+
+!call getarg(1,anatime)
```
CoryMartin-NOAA, line 110: Remove these commented out lines.

src/ncar-bufr2nc-fortran/makefile
```diff
@@ -0,0 +1,56 @@
+#! /bin/sh -v
```
CoryMartin-NOAA, line 1: use CMake, not a makefile.

src/ncar-bufr2nc-fortran/netcdf_mod.f90
```diff
@@ -0,0 +1,316 @@
+!ifort  -L${NETCDF}/lib -lnetcdf -lnetcdff -I${NETCDF}/include test.f90 -o a.out
```
CoryMartin-NOAA, line 1: Remove this comment.

src/ncar-bufr2nc-fortran/netcdf_mod.f90
```diff
@@ -0,0 +1,316 @@
...
+
+contains
+
+subroutine open_netcdf(fname,ncfileid)
+   character(len=*), intent(in) :: fname
+   integer, intent(out) :: ncfileid
+
+   ncstatus = nf90_open(path=trim(adjustl(fname)),mode=nf90_nowrite,ncid=ncfileid)  ! open file
+   if ( ncstatus .eq. 0 ) then
+!     write(*,fmt='(a)') 'opened '//trim(adjustl(fname))//' for reading'
```
CoryMartin-NOAA, line 41: Make this a not equal if and remove the commented out lines?

src/ncar-bufr2nc-fortran/ufo_variables_mod.F90
```diff
@@ -0,0 +1,216 @@
...
+!
+!  This software is licensed under the terms of the Apache Licence Version 2.0
+!  which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+!
+
+module ufo_vars_mod
+
+implicit none
+private
+!public :: ufo_vars_read, ufo_vars_getindex
```
CoryMartin-NOAA, line 12: remove comment

src/ncar-bufr2nc-fortran/ufo_variables_mod.F90
```diff
@@ -0,0 +1,216 @@
...
+character(len=MAXVARLEN), public, parameter :: var_sulfate = "mass_fraction_of_sulfate_aerosols_in_air"
+character(len=MAXVARLEN), public, parameter :: var_no3an1 = "mass_fraction_of_nitrate001_in_air"
+character(len=MAXVARLEN), public, parameter :: var_no3an2 = "mass_fraction_of_nitrate002_in_air"
+character(len=MAXVARLEN), public, parameter :: var_no3an3 = "mass_fraction_of_nitrate003_in_air"
+
+
+! ------------------------------------------------------------------------------
+contains
+
+!subroutine ufo_vars_read(f_vars, vars)
```
CoryMartin-NOAA, line 173: remove commented out subroutine

test/CMakeLists.txt
```diff
@@ -92,6 +92,20 @@ if( iodaconv_bufr_ENABLED )
...
 
+if( iodaconv_pbfortran_ENABLED )
+  list( APPEND test_input
+    testinput/gnssro.bufr
+    testinput/prepbufr.bufr
+    testinput/amsua.bufr
+  )
+
+  list( APPEND test_output
+    testoutput/gnssro_obs_2020093018.nc4
```
CoryMartin-NOAA, line 103: We already have GNSSRO, MHS, etc. BUFR files in this repository, can we reuse them instead of adding new ones? Do we already have a PrepBUFR file too?

CoryMartin-NOAA (changes requested):
Went through a did a quick review.

My general comments are:

- This doesn't use IODA-engines, how will that be rectified later?
- How does this fit in with the BUFR Ingester that @rmclaren wrote? Ideally we wouldn't want to duplicate efforts to convert BUFR to IODA
- Way too much excess code added here from GSI
- Make sure you are using CMake to build this Fortran executable. You shouldn't need a 'makefile'

jamiebresch:
Thank you @CoryMartin-NOAA for your time and comments. I think I have made the changes you requested.
I understand the path forward is to use ioda-engine and BUFR ingester. But since the current fortran code is working reasonably well and I feel it does not hurt to make it available for others.
As for the test files, if I know how to link gdas.t18z.1bmhs.tm00.bufr_d to mhs.bufr and gnssro_kompsat5_20180415_00Z.bufr to gnssro.bufr for ctest, I'm happy to use the existing files.

CoryMartin-NOAA:
@jamiebresch can you make the executables take any input filename rather than these specific names easily enough? Otherwise there should be a way to link the BUFR files to some other name using CMakeLists.txt but I'm not sure how to do it off the top of my head, maybe @srherbener does, otherwise we'd have to do some trial and error.

jamiebresch:
> @jamiebresch can you make the executables take any input filename rather than these specific names easily enough? Otherwise there should be a way to link the BUFR files to some other name using CMakeLists.txt but I'm not sure how to do it off the top of my head, maybe @srherbener does, otherwise we'd have to do some trial and error.

I did think about the suggestion you mentioned before deciding to add new testinput files. The current implementation is the simplest way (in my opinion) to run this fortran converter. Sorry that I am reluctant to make future changes purely for ctest purposes.

srherbener:
@jamiebresch I'm doing my review now and will hopefully have that submitted soon.

I agree with @CoryMartin-NOAA's comment to take any input file name. I think this will make this code more flexible and thus more usable in other flows. Note that this goes beyond testing and helps make the script more accessible to other users.

For example, it might work well to change this code so that it operates on only one file. The -i option could take the path to the input bufr file as an argument and the -o option could also take the path to the output ioda file. This would allow different users to write their own scripts to set up and run this converter on a set of files. As a side benefit, this would also solve the ctest issue.

You could move the functionality that loops through the files with the expected file names in this code to a script that just calls this converter to covert each file. Then if someone else wants to use this converter, they can write their own script to walk through their files (which could be organized completely differently than your case) and do the conversions.

jamiebresch:
> @jamiebresch I'm doing my review now and will hopefully have that submitted soon.
> 
> I agree with @CoryMartin-NOAA's comment to take any input file name. I think this will make this code more flexible and thus more usable in other flows. Note that this goes beyond testing and helps make the script more accessible to other users.
> 
> For example, it might work well to change this code so that it operates on only one file. The -i option could take the path to the input bufr file as an argument and the -o option could also take the path to the output ioda file. This would allow different users to write their own scripts to set up and run this converter on a set of files. As a side benefit, this would also solve the ctest issue.
> 
> You could move the functionality that loops through the files with the expected file names in this code to a script that just calls this converter to covert each file. Then if someone else wants to use this converter, they can write their own script to walk through their files (which could be organized completely differently than your case) and do the conversions.

Thank you for the suggestions. I will think more about it.
Just want to clarify that the current code works with only one or any combination/order of files.

srherbener (changes requested):
Thank you for this update!

I took a look at the output files that are generated by this code and I have a comment. We are obsoleting the representation of date time stamps using the "date_time" global attribute with the offset time in the variable time@MetaData. We will move forward with the ISO 8601 strings in the variable datetime@MetaData.

For this PR I'm requesting:
- Remove the date_time attribute and time@MetaData from the output ioda files
- Add the command line option to disable the call to the filtering operations
- Make the appropriate changes with the tolerances on the tests (to get them passing)

For future work, I strongly recommend:
- Make this code do just one conversion per call. Ie, change the command line to where one can specify any input file and any output file. Then move the functionality that loops through a set of files into a separate script. This will make this program much more accessible to other flows.

Here are some more general comments for consideration.

There is a lot of duplicated functionality in this implementation such as looping through bufr messages and subsets, extracting bufr mnemonics, formatting the extracted data, setting constants, etc. that has come about by copy-and-pasting separate programs into one system. It would help a lot for maintainability if the common code could be consolidated into a module that is shared among the parts of code that need to be different.

There are a lot of hardwired instrument ids, bufr mnemonics, variable names. If these could be specified in an external configuration (such as a YAML file) it would go a long way to facilitating the consolidation of the duplicated code into a common module.

The duplicated code and hardwired data need to be addressed at some point. Otherwise, this code is going to be difficult to maintain as more observation types are added. I realize that this code is meeting NCAR needs for now and it's important to support that. I think it's okay to merge this in for now and consider the comments for future work.

Thanks!

test/CMakeLists.txt
```diff
@@ -456,6 +470,41 @@ if(iodaconv_bufr_ENABLED)
...
+                            gnssro_obs_2020093018.nc4 ${IODA_CONV_COMP_TOL_ZERO}
+                    DEPENDS bufr2nc_fortran.x)
+
+  ecbuild_add_test( TARGET  test_${PROJECT_NAME}_prepbufr_conv
+                    TYPE    SCRIPT
+                    COMMAND ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                    ARGS    netcdf
+                            "${CMAKE_BINARY_DIR}/bin/bufr2nc_fortran.x
+                            -i testinput -o testrun prepbufr.bufr"
+                            sondes_obs_2020093018.nc4 ${IODA_CONV_COMP_TOL_ZERO}
```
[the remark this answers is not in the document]
jamiebresch: specific_humidity@ObsError is a derived value, so I have changed the tolerance. Now ctest passes for intel.

src/ncar-bufr2nc-fortran/main.f90
```diff
@@ -0,0 +1,158 @@
...
+   filename = 'prepbufr.bufr'
+   if ( trim(flist(ifile)) == trim(filename) ) then
+      inquire(file=trim(inpdir)//trim(filename), exist=fexist)
+      if ( .not. fexist ) then
+         write(*,*) 'Warning: ', trim(inpdir)//trim(filename), ' not found for decoding...'
+      else
+         ! read prepbufr file and store data in sequential linked list for conv obs
+         call read_prepbufr(trim(inpdir)//trim(filename), filedate)
+
+         call filter_obs_conv
```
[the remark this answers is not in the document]
jamiebresch: I have changed the default to not apply additional filtering and updated testoutput and README to reflect the status.

jamiebresch:
> For this PR I'm requesting:
> 
>     * Remove the date_time attribute and time@MetaData from the output ioda files
> 
I have removed time@MetaData from the output. I didn't not have ``date_time = 2018041500`` in the global attribute. You probably looked at the output from ncdiag converter.

jamiebresch:
Thank you @srherbener for the review and recommendation! The code was put together rather quickly to meet our current practical needs. We will be developing more sustainable solutions moving forward as you pointed out.

srherbener (changes requested):
Thank you for addressing my earlier comments! I missed the broken gnssro test that I've commented on below. Sorry about that (but it is good to catch this now rather than later).

test/CMakeLists.txt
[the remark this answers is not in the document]
srherbener: Thanks!

src/ncar-bufr2nc-fortran/main.f90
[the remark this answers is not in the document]
srherbener: Thanks!

test/CMakeLists.txt
```diff
@@ -458,6 +471,41 @@ if(iodaconv_bufr_ENABLED)
...
+
+if( iodaconv_pbfortran_ENABLED )
+
+  ecbuild_add_test( TARGET  test_${PROJECT_NAME}_gnssro_conv
+                    TYPE    SCRIPT
+                    COMMAND ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                    ARGS    netcdf
+                            "${CMAKE_BINARY_DIR}/bin/bufr2nc_fortran.x
+                            -i testinput -o testrun gnssro_kompsat5_20180415_00Z.bufr"
+                            gnssro_kompsat5_2018041500.nc4 ${IODA_CONV_COMP_TOL_ZERO}
```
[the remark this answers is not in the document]
jamiebresch: Thank you @srherbener for pointing this out. gnssro_obs_2018041500.nc4 is actually identical to gnssro_kompsat5_2018041500.nc4. I have restored the check on gnssro_obs_2018041500.nc4.

srherbener (approved):
@jamiebresch thanks for fixing the test!

srherbener:
@CoryMartin-NOAA, have your comments all been addressed? I'd like to merge this in soon since it's been outstanding for a long time. Thanks!

CoryMartin-NOAA (approved):
I think all my comments have been addressed. Approved.

srherbener:
@CoryMartin-NOAA thanks for the follow up. I'll merge this in now.

PR #380, opened 2020-12-09
Add the capability of combining multiple AHI files into one

YaliWu0219:
## Description

Add the capability of combining multiple Himawari8 AHI radiance data files in the ioda format into one. Have tested it in a 4DEnVar experiment in the NCAR/MMM group.


CoryMartin-NOAA (changes requested):
I think it looks good, but now that you're adding AHI, can you change the name of the script from 'combine_conv' to something else? 'conv' implies conventional only. Maybe just 'combine_files'?

YaliWu0219:
> I think it looks good, but now that you're adding AHI, can you change the name of the script from 'combine_conv' to something else? 'conv' implies conventional only. Maybe just 'combine_files'?

@CoryMartin-NOAA Thank you for your review. I agree, 'combine_files' makes more sense. This script also combines the GeoVals files.

src/gsi-ncdiag/combine_files.py
```diff
@@ -12,10 +12,10 @@
 import datetime as dt
 from pathlib import Path
 
-IODA_CONV_PATH = Path(__file__).parent/"@SCRIPT_LIB_PATH@"
-if not IODA_CONV_PATH.is_dir():
-    IODA_CONV_PATH = Path(__file__).parent/'..'/'lib-python'
-sys.path.append(str(IODA_CONV_PATH.resolve()))
+IODA_FILES_PATH = Path(__file__).parent/"@SCRIPT_LIB_PATH@"
```
CoryMartin-NOAA, line 15: This CONV is for converters not, conventional, need to change these 4 lines back.

CoryMartin-NOAA (changes requested):

src/gsi-ncdiag/combine_files.py
YaliWu0219: Oh, I see now.

CoryMartin-NOAA (approved):

srherbener (approved):

PR #381, opened 2020-12-15
Modified gsi ncdiags converter to remove extreme negative values, like -9.99e+9.

rhoneyager:
This is an experiment to solve an issue with the ncdiags files detailed in JCSDA-internal/ufo#466. One of the variables uses an anomalous value to denote missing data (-9.99e+09), and this is causing problems in ufo's bias correction algorithms.

Currently not tested on S4 due to https://github.com/JCSDA-internal/jedi-stack/issues/49, but in theory it should work.

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -733,6 +733,7 @@ class Conv(BaseGSI):
                                 tmp[tmp > 4e4] = self.INT_FILL
                             else:
                                 tmp[tmp > 4e8] = self.FLOAT_FILL
+                                tmp[tmp < -9e9] = self.FLOAT_FILL
                             outdata[gvname] = tmp
                     # store values in output data dictionary
                     outdata[varDict[outvars[o]]['valKey']] = obsdata


CoryMartin-NOAA (approved):
Should work, approving assuming it fixes your problem.

CoryMartin-NOAA (changes requested):
Actually, looking more closely, this is only going to impact the obs, the issue said it is a problem with the geovals. Need to fix it around line 615 or so

rhoneyager:
Yeah, I'm testing now and waiting for the converter to finish running. Do you have any idea why GSI is writing out a value of -9989999616.0?

CoryMartin-NOAA:
No, the missing value that I'm familiar with is a large positive value, not large negative. Perhaps it's specific to this field being written out?

srherbener (commented):
I think this change is fairly safe. The danger of using the "<" test is when a valid value passes that test. But a huge negative number seems unlikely to be a valid value.

Line 735 check seems not quite as safe since you might see channel frequencies in that range. However, we haven't seen any issues with that test, and perhaps it's not possible for channel frequencies variable to make it to this place in the code. i would leave it for now.

Does GSI mark -9.99e+9 as the fill value in the ncdiag file? If so, then you can query the fill value and check for equality to that to mark missing values. If not, then I don't think you have much choice but to do the "<" check.

rhoneyager:
I just looked through the source code, and that is the fill value used everywhere. However, my local Python changes around line 615 had no effect. Reading an unfamiliar code combined with Python language fatigue, I guess. Anyways, for our satellite geovals (tsavg5 -> average_surface_temperature_within_field_of_view) **always** appears to be set to the fill value.

CoryMartin-NOAA:
@rhoneyager @srherbener line 735 is in the toIODAobs method, so it will not affect how the geovals file is written out. That's why I suggested around line 615. Should be something like vdata[vdata < -9e8] = FILL

srherbener:
The geovals netcdf reader is a separate code from the ioda reader. It's in `ufo/src/ufo/ufo_geovals_mod.F90`. This code doesn't recognize the netcdf fill values. Lines 1090,1091 in `ufo_geovals_mod.F90` are:

```
1090   ! set the missing value equal to IODA missing_value
1091   where (self%geovals(ivar)%vals > 1.0e08) self%geovals(ivar)%vals = self%missing_value
```
So it looks like the missing value mark is only placed when the value is > 1.0e8. This was the logic ioda used way back before we changed it to recognize the netcdf fill value. Seems like you will need to either fix the geovals reader, or have the converter change missing values in geovals files to > 1.0e8.

srherbener:
@rhoneyager, if I recall correctly you took the approach to modify the geovals reader to solve this issue. If that is the case, is this PR still needed?

rhoneyager:
I'll close it. Superseded.

PR #382, opened 2020-12-17
Add in fix for sfc p and geop hgt to ensure it is converted for geovals

CoryMartin-NOAA:
Add in new options to the python dictionary for GeoVaLS to make sure surface pressure and surface geopotential height are converted properly, especially for surface ps obs.


gthompsnJCSDA (approved):
I approve.  Should line#293 really convert a variable name of surface_height to surface_geopotential_height?  Might this be the one that gets called "surface_altitude?"  It would be nice to know/fix since there are minor differences with geopotential vs. geometric height.

srherbener (approved):
Thanks @CoryMartin-NOAA for all the hard work you've been putting in to support the OBS code sprint.

Issue #373, opened 2020-12-03
BufrConverter: Output variables are double precision instead of single precision.


rmclaren:
The variables in the NetCDF output should be encoded as single precision (floats) instead of double precision (doubles) numbers.

Issue #372, opened 2020-12-03
BufrConverter: Both nlocs and nchans are defined with unlimited size.


rmclaren:
For some reason the dimensions for the NetCDF files created are reported to have UNLIMITED size. According to @srherbener these sizes should be fixed.

srherbener:
The unlimited size is coming from this line:

https://github.com/JCSDA-internal/ioda-converters/blob/1ed4019613a611deea243f36cdb4392df45232a9/src/bufr/IodaEncoder/IodaEncoder.cpp#L62

This argument to the NewDimensionScale constructor is setting the maximum allowed dimension size. 

It's best to set dimensions to fixed sizes if you know those sizes beforehand. I'm not sure if it's possible to know nlocs before reading the bufr file, but hopefully you can determine the other dimension sizes.

- Best case is all dimensions be fixed sizes. Best performance.
- Next best case is nlocs is unlimited and all other dimensions are fixed. ObsStore backend can handle this already.
- Next is multiple dimensions are unlimited. ObsSpace backend can't handle this yet, so more development is necessary to support this.

rmclaren:
@srherbener 

I can determine the number of locations ahead of time very easily, but I had some additional questions:

1) Should chunk sizes for dimension variables be defaulted to a value like (ex: 1000) or just be the length of that dimension (ex: nlocs)?

2) Should the user be able to specify a chunksize for a dimension variable manually through the configuration or by adding an optional "chunkSize" parameter to the dimension description?

srherbener:
@rmclaren sorry about the slow response on this. It dropped through the cracks.

1. The HDF5 developers recommend keeping the chunk size <= 1 MB. The chunk size determines how much space is allocated so ideally you want the chunk size to be a factor of the total size. If the total size is 20, then chunk sizes of 4, 5, 10 are good because the memory gets utilized at a 100%. If you were to select a chunk size of 7, you waste one location at the end, ie three chunks, where the usage is 7, 7, and 6 (20 items in a space big enough for 21 items). If the chunk size is 9, then there is a lot of waste: 3 chunks holding 9, 9 and 2 items (8 unused spaces).

2. Having the user specify chunk sizes might be a good feature but only for the sophisticated user. It might be a bit much to ask users in general to be up to speed on hdf5 chunking. If you decide to add this, it would probably be good to make it optional and thus hidden by default.

Issue #367, opened 2020-11-24
Enhanced Exports features to BUFR Parser.


rmclaren:
- Add the ability to split output into separate files based on the value of one or more mnemonics.
- Add ability to apply some simple filtering to the data (remove data where mnemonic is outside a specified bounds for example).
- Update source descriptions to be more specific "variable/radiance" vs just radiance.
- Add some attributes to the data so you can do stuff like "variable/radiance.nrows" to get the number of datapoints in an observation.

PR #383, opened 2020-12-17
Enhanced BUFR Exports

rmclaren:
## Description

This branch enhances the options the user has when exporting BUFR data to IODA.

### Issue(s) addressed

#367 
#386 
#373
#372 
#387 
#389


src/bufr/BufrParser/BufrDescription.cpp
```diff
@@ -30,14 +24,13 @@ namespace
         const char* Mnemonics = "mnemonics";
         const char* Channels = "channels";
         const char* Exports = "exports";
-        const char* Datetime = "datetime";
-        const char* Mnemonic = "mnemonic";
```
ilianagenkova, line 34: Could these deleted field break the downstream code? I managed to compile your branch but asking for posterity and my own understanding.

src/bufr/BufrParser/BufrParser.h
```diff
@@ -51,10 +54,20 @@ namespace Ingester
         /// \brief Exports collected data into a DataContainer
         std::shared_ptr<DataContainer> exportData(const BufrDataMap& sourceData);
 
+        /// \brief Function responsible for splitting the data into the categories. This function
+        ///        gets called over and over for each split speciefied, sub-splitting the data
```
ilianagenkova, line 58: Fix typo in specified, and I suggest you add "specified by filter"

src/bufr/BufrParser/Exports/Export.cpp
```diff
@@ -0,0 +1,188 @@
+/*
```
ilianagenkova, line 1: Is this a newly added file?

src/bufr/BufrParser/Exports/Splits/CategorySplit.h
```diff
@@ -0,0 +1,54 @@
...
+#include "Split.h"
+
+#include <string>
+#include <vector>
+#include <map>
+
+
+namespace Ingester
+{
+    /// \brief Data splitter class that splits data according to a predefined categories.
```
ilianagenkova, line 19: May be add an example of one or two categories.

ilianagenkova (approved):

src/bufr/BufrParser/BufrDescription.cpp
rmclaren: These constants are not visible outside of BufrDescription as they are only declared in the .cpp file. So its not possible for them to affect downstream code (third party code that makes use of the library being generated by this code).

src/bufr/BufrParser/Exports/Splits/CategorySplit.h
rmclaren: Added extra documentation to explain things with more detail.

src/bufr/BufrParser/Exports/Export.cpp
rmclaren: Yes

src/bufr/BufrParser/BufrParser.h
rmclaren: Thanks for finding the typo. I'm trying to draw a distinction between splits (divide data into sub-parts according to a metric) and filters (throw away data that does not meet a given requirement) so I changed the string to "for each specified Split object".

ilianagenkova: This is a good way of explaining it!

rmclaren: Thanks!

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
```diff
@@ -0,0 +1,70 @@
...
+            std::ostringstream datetimeStr;
+            datetimeStr << std::setfill('0')
+                        << std::setw(4) << map.at(yearKey_)(idx) << "-" \
+                        << std::setw(2) << map.at(monthKey_)(idx) << "-" \
+                        << std::setw(2) << map.at(dayKey_)(idx) << "T" \
+                        << std::setw(2) << map.at(hourKey_)(idx) << ":" \
+                        << std::setw(2) << map.at(minuteKey_)(idx) << ":" \
+                        << std::setw(2) << map.at(secondKey_)(idx);
+
+            if (isUTC_)
```
CoryMartin-NOAA, line 60: Not needed for this PR, but suggestion:
if not isUTC_, provide a time offset in YAML and convert to UTC time. Ex. UTC_offset=-5, to convert obs in EST to UTC.

src/bufr/DataContainer.cpp
```diff
@@ -18,39 +18,155 @@
...
-        if (!hasKey(fieldName))
+        if (!hasKey(fieldName, categoryId))
         {
-            std::ostringstream errorStr;
-            errorStr << "ERROR: Field called " << fieldName << " doesn't exists.";
-            throw eckit::BadParameter(errorStr.str());
+            std::ostringstream errStr;
+            errStr << "ERROR: Either field called " << fieldName;
+            errStr << " or category " << makeSubCategoryStr(categoryId);
+            errStr << " doesn't exists.";
```
CoryMartin-NOAA, line 56: 'does not exist'

src/bufr/DataContainer.h
```diff
@@ -20,36 +20,76 @@
...
-        DataContainer() = default;
+        /// \brief Simple constructor
+        DataContainer();
+
+        /// \brief Construct to create container with subcategories.
+        /// \details constructor that creates a underlying data structure to store data in separate
+        ///          sub categories defined by combining all possible combinations of categories
+        ///          defined in the category map.
+        /// \param categoryMap map of major category types ex: "SatId" to the possible sub types
+        ///        for the category type ex: {"GEOS-15", "GEOS-16", "GEOS-17"}.
```
CoryMartin-NOAA, line 48: GOES not GEOS

srherbener (approved):
Looks good!

I've left some details below about the Parameters class in OOPS for your consideration.

I also was curious and took a look inside bufr2ioda.cpp. Just a comment for after this PR. It seems not quite accurate to call the routine that does the entire conversion "parse", since parsing the input BUFR file is what that first part of the flow is doing. It might make sense to rename the Ingester::parse routine to something that indicates parsing and encoding.

jamiebresch:
The legacy ioda format uses 1-d array for each channel `brightness_temperature_1@ObsValue(nlocs)` while the current examples output a 2-d arrays `@ObsValue(nlocs, nchans)`.  Could @rmclaren provide an example yaml showing how to export TMBR of multiple channels to individual_channel@ObsValue(nlocs)?

test/CMakeLists.txt
```diff
@@ -86,11 +86,17 @@ if( iodaconv_bufr_ENABLED )
     testinput/gdas.t00z.1bhrs4.tm00.bufr_d
```
jamiebresch, line 86: Line 84 (two lines above, I can not add a comment there) lists testinput/bufr_test.yaml, but that file does not seem exist.

src/bufr/BufrParser/Exports/Export.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "Splits/CategorySplit.h"
+#include "Variables/MnemonicVariable.h"
+#include "Variables/DatetimeVariable.h"
+#include "Variables/Transforms/Transform.h"
+#include "Variables/Transforms/TransformBuilder.h"
+
+
+namespace
+{
+    namespace ConfKeys
```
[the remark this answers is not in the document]
rmclaren: Cool, I'll look into it and add Zenhub ticket. It would be really good to have YAML validation as you described. Currently I'm effectively validating the YAML in many different places manually... not the greatest design.

tools/run_satwnds/satwnds_new_subset_template.yaml
```diff
@@ -0,0 +1,170 @@
...
+        - mnemonics: [SAID, CLATH, CLONH, YEAR, MNTH, DAYS, HOUR, MINU, SECO, SWCM, SAZA, OGCE, SCCF, SWQM]
+        - mnemonics: [EHAM, PRLC, WDIR, WSPD]
+        - mnemonics: [TCOV, CVWD]
+        - mnemonics: [PCCF]
+          channels: 1-4
+
+      exports:
+        filters:
+          - bounding:
+              mnemonic: CLON
```
jamiebresch, line 20: CLON should be CLONH for new

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -68,28 +71,72 @@ namespace Ingester
         return exportData(collectors.finalize());
     }
 
-    std::shared_ptr<DataContainer> BufrParser::exportData(const BufrDataMap& srcData)
+    std::shared_ptr<DataContainer> BufrParser::exportData(const BufrDataMap& dataData)
```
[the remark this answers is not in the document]
rmclaren: Thanks for noticing this. I changed it back to srcData. The local mutable copy was changed to dataCopy.

src/bufr/BufrParser/Exports/Filters/RangeFilter.cpp
```diff
@@ -0,0 +1,54 @@
+/*
```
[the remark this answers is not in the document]
rmclaren: Replaced all these filters with one filter called "Bounding filter". You have the choice of setting an upperBound, lowerBound or both.

src/bufr/BufrParser/Exports/MnemonicExport.cpp
```diff
@@ -20,6 +24,15 @@ namespace Ingester
 
     std::shared_ptr<DataObject> MnemonicExport::exportData(const BufrDataMap& map)
     {
+        if (map.find(mnemonic_) == map.end())
+        {
+            std::stringstream errStr;
+            errStr << "Mnemonic " << mnemonic_;
+            errStr << " couldn't be found during export.";
```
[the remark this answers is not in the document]
rmclaren: I went ahead and changed these.

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
```diff
@@ -53,18 +77,58 @@ namespace Ingester
...
+        {
+            if (map.find(key) == map.end())
+            {
+                isKeyMissing = true;
+                errStr << key;
+                break;
+            }
+        }
+
+        errStr << " couldn't be found during export of datetime object.";
```
CoryMartin-NOAA, line 127: one remaining contraction

CoryMartin-NOAA (approved):
Going to approve, I only see one issue and it is a lingering contraction, so I think we get this merged in before the PR becomes any larger.

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
rmclaren: I went ahead and fixed these because I don't want a lot of YAML files being generated with out of date syntax. I added a new parameter called "hoursFromUtc" as an optional and I've deprocated "isUtc". By default hoursFromUtc is 0. I also made the parameter "second" optional while I was at it. I discovered I'd inadvertently created duplicates for this class (DatetimeExport) so I removed that and MnemonicExport (also a duplicate).

src/bufr/BufrParser/Exports/Export.cpp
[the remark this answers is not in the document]
rmclaren: Resolved by writing Zenhub issue.

src/bufr/README.md
```diff
@@ -0,0 +1,199 @@
+# Bufr Converter
```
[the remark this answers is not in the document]
rmclaren: thanks!

tools/run_satwnds/satwnds_new_subset_template.yaml
rmclaren: Good catch!

CoryMartin-NOAA:
@srherbener I say this is probably ready to merge. 155 commits is enough :-P

tools/run_satwnds/README.md
```diff
@@ -0,0 +1,23 @@
...
+with the proper configuration for that subset.
+
+## Dependencies
+
+* **Python 3.6+**
+
+The following executables must be available in the shell path.
+
+* **bufr2ioda.x** - From ioda_converters _feature/bufr-enhanced-exports_
+* **split_by_subset.x** - From NCEPLibs-Bufr _feature/split_by_subset_
```
jamiebresch, line 13: feature/split_by_subset has been merged to NOAA-EMC/NCEPLIBS-bufr/develop
feature/bufr-enhanced-exports will be merged soon.

test/CMakeLists.txt
rmclaren: Good catch! I removed it.

rmclaren:
> The legacy ioda format uses 1-d array for each channel `brightness_temperature_1@ObsValue(nlocs)` while the current examples output a 2-d arrays `@ObsValue(nlocs, nchans)`. Could @rmclaren provide an example yaml showing how to export TMBR of multiple channels to individual_channel@ObsValue(nlocs)?

@jamiebresch It is possible to export a single channel from a mnemonic like TMBR by setting the `channel` to a single number (ex: 3 or something) when you define it for reading in the mnemonic sets section and then exporting that. I don't think that's really what you are asking for, however. I think what you are saying is that you want the ability split individual column indices from a mnemonic  into separate exported variables. I don't have any way to do that right now. I've been thinking I could eventually add something like the following example to support your use case (NOTE: I've been thinking I want to replace the term "channels" with "indices" to be more generic. I don't think the term "channels" makes sense for all multi-dimensional mnemonics, really what I meant is column indices):

```yaml
export:
    variables:
        radiance_chan_1:
            mnemonic: TMBR
            indices: 1
        radiance_others:
            mnemonic: TMBR
            indices: 2-5
``` 

Would this meet your needs?

aerorahul:
> > The legacy ioda format uses 1-d array for each channel `brightness_temperature_1@ObsValue(nlocs)` while the current examples output a 2-d arrays `@ObsValue(nlocs, nchans)`. Could @rmclaren provide an example yaml showing how to export TMBR of multiple channels to individual_channel@ObsValue(nlocs)?
> 
> @jamiebresch It is possible to read a single channel from a mnemonic like TMBR by setting the `channel` to a single number (ex: 3 or something) when you define it for reading in the mnemonic sets section and then exporting that. I don't think that's really what you are asking for, however. I think what you are saying is that you want the ability split individual column indices from a mnemonic into separate exported variables. I don't have any way to do that right now. I've been thinking I could eventually add something like the following example to support your use case (NOTE: I've been thinking I want to replace the term "channels" with "indices" to be more generic. I don't think the term "channels" makes sense for all multi-dimensional mnemonics, really what I meant is column indices):
> 
> ```yaml
> export:
>     variables:
>         radiance_chan_1:
>             mnemonic: TMBR
>             indices: 1
>         radiance_others:
>             mnemonic: TMBR
>             indices: 2-5
> ```
> 
> Would this meet your needs?

Note that this e.g. `brightness_temperature_1[nlocs]` is the legacy format.  This converter uses the ioda-engines format that allows storing multi-dimensional data e.g. `brightness_temperature[nlocs, nchans]`
Is this capability required in this bufr converter code base? @srherbener

rmclaren:
@aerorahul @jamiebresch What Rahul said is true. We should avoid writing extra code to support the "old" format. It would be far better to update the dependent code bases to work with the "new" format.

jamiebresch:
> @aerorahul @jamiebresch What Rahul said is true. We should avoid writing extra code to support the "old" format. It would be far better to update the dependent code bases to work with the "new" format.

Yes, I agree. Thanks.

aerorahul (changes requested):
merged `develop`

aerorahul (approved):

PR #392, opened 2021-01-12
Feature/smap ssm

YoulongXia-NOAA:
## Description

SMAP Level 2 NRT (Near Real-Time) surface volumetric soil moisture converter is initially implemented and added to ioda-converters bundle. The purpose is to convert native h5 format to ioda-converters' NETCDF format.

### Issue(s) addressed
link to initially implement SMAP surface soil moisture

## Dependencies

No dependency

## Impact

The changes in this PR will not affect other repositories


srherbener:
@YoulongXia-NOAA, all three tests are failing because the container does not have the Python h5py module installed.
```
test 38
--
514 | Start 38: test_iodaconv_smap_ssm
515 |  
516 | 38: Test command: /build_container/bin/iodaconv_comp.sh "netcdf" "/build_container/bin/smap_ssm2ioda.py
517 | -i testinput/SMAP_L2_SM_P_NRT_24342_A_20190822T224858_N16023_002.h5
518 | -o testrun/smap_ssm.nc
519 | -m maskout" "smap_ssm.nc" "0.0"
520 | 38: Environment variables:
521 | 38:  OMP_NUM_THREADS=1
522 | 38: Test timeout computed to be: 1500
523 | 38: Traceback (most recent call last):
524 | 38:   File "/build_container/bin/smap_ssm2ioda.py", line 9, in <module>
525 | 38:     import h5py
526 | 38: ModuleNotFoundError: No module named 'h5py'
527 | 1/1 Test #38: test_iodaconv_smap_ssm ...........***Failed    0.02 sec
528 |  
529 | 0% tests passed, 1 tests failed out of 1
530 |  
531 | Label Time Summary:
532 | iodaconv    =   0.02 sec*proc (1 test)
533 | script      =   0.02 sec*proc (1 test)
534 |  
535 | Total Test time (real) =   0.02 sec
536 |  
537 | The following tests FAILED:
538 | 38 - test_iodaconv_smap_ssm (Failed)
539 | Errors while running CTest
```

Up until now, the python converters have been using the netcdf module (netCDF4). It looks like it might not be too much work to switch the new script `smap_ssm2ioda.py` from the h5py calls to the netCDF4 calls. Would it be possible to do this (use the netCDF4 module instead of the h5py module)?

If there is a strong reason to be using the h5py module instead of netCDF4, then we can do that and it will take a bit of coordination to get the containers updated.

Thanks!

YoulongXia-NOAA:
I googled " To read HDF5 file in Python? How to read h5 file in python? Use netCDF4 module to read h5 file in python" and I did not find any example to read h5 file using netCDF4 module. All examples use h5py to read h5 file. Furthermore, the website also recommended to use h5py to read their SMAP h5 file - http://hdfeos.org/zoo/index_openNSIDC_Examples.php#SMAP and https://github.com/nsidc/smap_python_notebooks/tree/main/notebooks. I think that h5py is needed to read h5 file, in particular if we want to use NASA data in the future.

srherbener:
@YoulongXia-NOAA, Thank you for switching to the netCDF4 module. I think that is a good short term solution. 

I did some investigation and getting the h5py and netCDF4 modules to live together takes a bit of doing. This is so because they both use a copy of the hd5f library but tend to want different versions of that library. This can be remedied by telling pip to install and use a separate copy of the hdf5 library when installing h5py, but that practice needs to either be adopted by the HPC IT groups, or the HPC IT groups need to allow the loading of python modules to a user's local space ($HOME/.local).

There is some coordination to work out if we want both h5py and netCDF4, and I'm okay doing that eventually if it makes sense. But for now using netCDF4 is a good short-term solution.

srherbener (approved):

CoryMartin-NOAA (approved):
Looks good, glad to hear you were able to use the netCDF API, @YoulongXia-NOAA . My only comment is the .h5 file should be in git-lfs, @srherbener can you move it? Do we need to add .h5 as a suffix to the automatic list?

srherbener:
@CoryMartin-NOAA good catch. I can fix this before merging.

YoulongXia-NOAA:
Than you for your quick reviw, Stephen and Cory. Well done.

PR #401, opened 2021-01-19
Fix typo that is causing bufr headers to install to the root directory.

markjolah:
## Description

Fix a typo in variable name for bufr CMake header install directory.

### Issue(s) addressed

Bad variable name causes headers to attempt to install to the root directory.

## Dependencies

None

## Impact

None


rmclaren (approved):

srherbener (approved):
Thanks for finding and fixing this!

Issue #396, opened 2021-01-13
IODA converter example to be added


nicholasesposito:
I have opened the [feature/converter-example](https://github.com/JCSDA-internal/ioda-converters/tree/feature/converter_example) branch on JCSDA-internal/ioda-converters.

The purpose is to add an sample of how the ioda converter would be. 

In the branch, there is:
- a README file dictating how to run the converter, information about the input, output, yaml input, and environment file.
- an input data directory and bufr file
- an output data directory and sample output 
- sample yaml file used with the converter
- a sample file that sets the environment required to run the converter. It is what can be used on Hera. 

The branch was pushed at 4:52pm EST on 1/13/21

nicholasesposito:
Additional information:
When the feature/converter-example branch is checked out, the project is in the CONVERTER_EXAMPLE directory.

The README and HERA environment file are both in the main directory.
There is also the yaml used with the converter in the main directory, and it is named "bufr_adpsfc.yaml"
The input data is in the DATA/ directory and is named "adpsfc"
The output directory is named testoutput/ and there is sample output named "ioda_encoder_result_adpsfc.nc"

aerorahul:
1. Please don't include the HERA file.  This is not a place for it.
2. Name the testinput file for adpsfc bufr as the original; e.g. gdas.t00z.adpsfc.bufr_tm00 
3. name the testoutput file for adpsfc netcdf as e.g. gdas.t00z.adpsfc.nc
4. The Input and output data needs to be in testinput and testoutput directories.
5. The README.md should be updated in the bufr directory.

CoryMartin-NOAA:
Just for my understanding/clarification, is this supposed to parallel the existing test/example that @rmclaren has (https://github.com/JCSDA-internal/ioda-converters/blob/develop/test/testinput/bufr_mhs.yaml and https://github.com/JCSDA-internal/ioda-converters/blob/a6140dc9107817f8dd23b4c57a50ef54ceba0dbc/test/CMakeLists.txt#L450) but for adpsfc?

nicholasesposito:
Cory, yes, this will be parallel to it. I'll probably name it "test_iodaconv_bufr_adpsfc2ioda".
Rahul, will do. Working on it now.

nicholasesposito:
[...]

PR #404, opened 2021-01-22
IODA converter example for adpsfc

nicholasesposito:
## Description

The purpose of the PR is to add an IODA converter example to IODA-Converters. This example will use the bufr2ioda executables to convert an adpsfc bufr file into an IODA-formatted netcdf file using a specified yaml file. This is an enhancement that adds test data.

The yaml and input bufr file have been placed in test/testinput
The output IODA-formatted netcdf file has been placed in test/testoutput
A test named "test_iodaconv_bufr_adpsfc2ioda" has been added to test/CMakeLists.txt

### Issue(s) addressed

The issue is an enhancement documented with issue #396 

## Dependencies

There are no other PRs that need to be merged before this one.

## Impact

Requires changes in the following repositories:
- [ ] ioda-converters

Requires updating test data for the following repositories:
- [ ] ioda-converters.


CoryMartin-NOAA:
@nicholasesposito looks like you need to merge in 'develop', also for pull requests instead of 'assigning' people you make people 'reviewers'. I've taken the liberty of doing the latter already, but you'll need to merge in develop as this PR has too many changes associated with it right now.

CoryMartin-NOAA:
git checkout develop
git pull
git pull # (for good measure again)
git checkout feature/converter_example
git merge develop

Note you may have to resolve conflicts, if so look for "HEAD and =======" in files.

aerorahul:
Screen Shot 2021-01-22 at 3 10 07 PM
I was able to fix the conflicts, update the yaml file and run the converter.
Attached is a figure.

I found there are some missing values:

```
ncdump -v temperatureAir@ObsValue  test/testrun/gdas.t06z.adpsfc.tm00.nc
   1e+11, 1e+11, 1e+11, 276.55, 276.05, 275.45, 275.35, 274.95, 274.75,
    277.15, 277.15, 277.15, 277.15, 277.15, 277.15, 277.15, 277.15, 278.15,
```

How are `missing_values` handled here?  There should be a attribute for it so that netCDF understands that.  @rmclaren

rmclaren:
@aerorahul The appropriate way to do this is to set the "range" attribute on the ioda variable (in the YAML). At least panoply will ignore values outside the valid range, not sure what ncdump will do.

aerorahul:
netCDF can prescribe a variable attribute such as `_FillValue` or `missing_value`. See [here](https://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html)
We should be able to put that attribute into the netCDF file.


NCEPlibs-bufr assigns missing values as `1e11`.  This is defined in `bfrini.f`

`valid_range` is fine, but we should define a variable attribute `_FillValue` or `missing_value` using the value from bufr data.

rmclaren:
Good point, I should add that.

PR #409, opened 2021-01-27
GSI satbias file -> ioda file converter

### LINKED ISSUE #402: Write a converter for satbias files
To simplify ufo code, bias correction coefficients need to be stored in self-described file.

Write a ioda converter that read bias coeffs from GSI satbias file and stores them in a ioda-readable file containing:
- coeffiicients
- predictor names
- channel indices

shlyaeva:
## Description

This adds a simple converter from GSI bias coefficient `satbias_in` file to the bias coefficients files readable with ioda-engines.

For each sensor (instrument+satellite), a separate bias coefficients netcdf file is created, containing:
- coefficients (2D variables)
- list of channels
- names of predictors

Based on discussion in https://github.com/JCSDA-internal/ufo/issues/539, different predictor names are used for:
- `goessndr .or. goes_img .or. ahi .or. seviri .or. ssmi .or. ssmis .or. gmi .or. abi` (`scan_position_*` for scan-angle-dependent predictors)
- other sensors (`scan_angle_*` for scan-angle-dependent predictors).

I am not aware of other cases when GSI uses different bias predictors for `bias_coefficient(jj)` (or `pred(npred)` in GSI) depending on the sensor; please let me know if I missed something.

### Issue(s) addressed

- fixes #402


src/satbias/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# (C) Copyright 2021 UCAR
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+ecbuild_add_executable( TARGET  satbias2ioda.x
+                        SOURCES SatBiasConverter.cpp GsiSatbiasReader.cpp GsiSatbiasReader.h
+                        LIBS    Eigen3::Eigen ${oops_LIBRARIES} ioda_engines )
+
+ecbuild_add_test( TARGET  test_satbias_coding_norms
```
aerorahul, line 10: I think the test name should be: `test_${PROJECT_NAME}_satbias_coding_norms`

src/satbias/GsiSatBiasReader.cpp
```diff
@@ -0,0 +1,159 @@
...
+                                                     "lapse_rate_order_2",
+                                                     "lapse_rate",
+                                                     "cosine_of_latitude_times_orbit_node",
+                                                     "sine_of_latitude",
+                                                     "emissivity",
+                                                     "scan_position_order_4",
+                                                     "scan_position_order_3",
+                                                     "scan_position_order_2",
+                                                     "scan_position"
+                                                    };
```
aerorahul, line 47: Just wondering if this is the best possible way and for future flexibility.
Would it be better to provide this information via the yaml file?

src/satbias/GsiSatBiasReader.h
```diff
@@ -0,0 +1,32 @@
...
+
+/// Read bias coefficients from the GSI bias coefficients file (satbias_in)
+/// \param filename file with bias coefficients (GSI style)
+/// \param sensor instrument+satellite that is to be read from the file
+/// \param(out) channels channels for \p sensor
+/// \param(out) coeffs bias coefficients
+void readObsBiasCoefficients(const std::string & filename, const std::string & sensor,
+                             std::vector<int> & channels, Eigen::ArrayXXf & coeffs);
+
+#
```
aerorahul, line 32: stray mark?

aerorahul (changes requested):
A couple of requested changes; stray mark and test name.
Something to discuss regarding predictors.
It is understood that this is GSI related converter so the predictors won't change (atleast for the global application), but could this be provided via anchors in the yaml file? So, we know exactly what predictors are being used/read for which sensor and are not hard-wired in the code.
Thoughts?

src/satbias/GsiSatBiasReader.cpp
aerorahul: @shlyaeva Yes. That is what I was thinking. One could put the list of predictors in the yaml.
For e.g. `satbias_converter.yaml` could be something like:
```
input file: testinput/satbias_crtm_in
default_predictors: &default_preds
  constant, zenith_angle, cloud_liquid_water, lapse_rate_order_2, lapse_rate, cosine_of_latitude_times_orbit_node, sine_of_latitude, emissivity, scan_angle_order_4,scan_angle_order_3, scan_angle_order_2, scan_angle
scanpos_predictors: &scanpos_preds
  constant, zenith_angle, cloud_liquid_water, lapse_rate_order_2, lapse_rate, cosine_of_latitude_times_orbit_node, sine_of_latitude, emissivity, scan_position_order_4, scan_position_order_3, scan_position_order_2, scan_position
output:
- sensor: amsua_n15
  output file: satbias_amsua_n15.nc4
  predictors: *default_preds
```

Thoughts?

shlyaeva: I think it's a very good idea; I'm working on it.

src/satbias/GsiSatBiasReader.h
shlyaeva: Oops; I've removed it now.

src/satbias/CMakeLists.txt
shlyaeva: thank you; fixed.

src/satbias/GsiSatBiasReader.cpp
shlyaeva: I've changed to read predictor names from the yaml.

src/satbias/GsiSatBiasReader.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "GsiSatBiasReader.h"
+
+#include <string>
+#include <vector>
+#include <fstream>
+
+#include <Eigen/Dense>
+
+constexpr size_t npredictors = 12;
```
aerorahul, line 16: should this be in the header?

shlyaeva: I've decided to keep it here, since it's only used in this file.
But it probably should be in the header; there should be a check whether number of predictors specified in yaml is the same as this one; I'll add that.

aerorahul (commented):
I checked out, built the branch and ran the tests.
I verified the contents of `satbias_amsua_n18.nc4` against `amsua_n18` from `satbias_crtm_in`.  The `bias_coefficients` variable is correctly populated in the netCDF file.

The changes are good to go.

Side question:
Are the variables `tlap`, `tsum` and `ntlapupdate` not necessary to be encoded in the netCDF file?

shlyaeva:
@aerorahul thank you for testing and reviewing!

> Side question:
> Are the variables `tlap`, `tsum` and `ntlapupdate` not necessary to be encoded in the netCDF file?

To be honest, I have no idea :) I just replicated the code Xin wrote in ufo for the I/O of satbias.
It would be easy to add though; do you know what those variables are?

aerorahul:
> @aerorahul thank you for testing and reviewing!
> 
> > Side question:
> > Are the variables `tlap`, `tsum` and `ntlapupdate` not necessary to be encoded in the netCDF file?
> 
> To be honest, I have no idea :) I just replicated the code Xin wrote in ufo for the I/O of satbias.
> It would be easy to add though; do you know what those variables are?

From `radinfo.F90` in the GSI [here](https://github.com/NOAA-EMC/GSI/blob/master/src/gsi/radinfo.f90)
`tlap` is the mean lapse rate
`tsum` is the sum of errors in the lapse rate
`ntlapupdate` is 999. How often/when to update the lapse rate (I think).

@emilyhcliu would know exactly what these are.

If it is not too much effort, since these are just 3 variables, I would lean towards adding them in the netCDF file.

aerorahul (approved):

src/satbias/GsiSatBiasReader.cpp
shlyaeva: I've moved this to the header, and added a check on number of predictors.

shlyaeva:
> If it is not too much effort, since these are just 3 variables, I would lean towards adding them in the netCDF file.

I've looked a bit at these, and I'm undecided on including them in the file now.
These are not bias correction coefficients, so this would require adding new variables.
One of them (`tlap`) seems to be used in the calculation of one of the predictors, so I think this information would have to be some place to replicate GSI. But I'm not sure whether the bias coefficients file is the best place for it (right now I plan to have separate yaml sections for describing bias predictors, and for reading the bias coefficients). It looks like Xin is also using a separate file for `tlap` I/O, maybe for the same reason: https://github.com/JCSDA-internal/ufo/blob/develop/test/testinput/amsua_crtm_bc.yaml#L56. I've compared the contents of that `amsua_n19_tlapmean.txt` file with satbias_in that is used in the tests, and the contents are the same as `tlap` from `satbias_in`. Perhaps eventually `tlap` could be in the ioda obs files? Or computed on the fly? I'm not sure how it's computed and whether it's changing with time.
The other two variables don't seem to be used anywhere in the bias correction computation, at a quick glance; I've likely missed something.
What do you think?

aerorahul:
> > If it is not too much effort, since these are just 3 variables, I would lean towards adding them in the netCDF file.
> 
> I've looked a bit at these, and I'm undecided on including them in the file now.
> These are not bias correction coefficients, so this would require adding new variables.
> One of them (`tlap`) seems to be used in the calculation of one of the predictors, so I think this information would have to be some place to replicate GSI. But I'm not sure whether the bias coefficients file is the best place for it (right now I plan to have separate yaml sections for describing bias predictors, and for reading the bias coefficients). It looks like Xin is also using a separate file for `tlap` I/O, maybe for the same reason: https://github.com/JCSDA-internal/ufo/blob/develop/test/testinput/amsua_crtm_bc.yaml#L56. I've compared the contents of that `amsua_n19_tlapmean.txt` file with satbias_in that is used in the tests, and the contents are the same as `tlap` from `satbias_in`. Perhaps eventually `tlap` could be in the ioda obs files? Or computed on the fly? I'm not sure how it's computed and whether it's changing with time.
> The other two variables don't seem to be used anywhere in the bias correction computation, at a quick glance; I've likely missed something.
> What do you think?

I am not very familiar with this, so I cannot really make a convincing argument for/against.  It is possible that these 3 variables don't need to be in the GSI abias file, but were placed there for convenience.  In either case, I am fine with not including them in the IODA satbias files at this time and revisit later if the need arises.  The only "for" argument would have been to have a complete transformation of GSI abias txt file into a IODA file.  But, it is not really that convincing.

This PR is complete and can be merged.

srherbener (approved):
Thank you for this contribution.

I tried this on my iMac, it builds successfully, and all the tests pass.

srherbener:
I think it's better to merge this in now, since it is complete, and address any issues that come up in subsequent PRs. I'm do the merge now.

PR #411, opened 2021-01-29
Fix case sensitivity issue in GsiSatBias

CoryMartin-NOAA:
GsiSatBiasReader.cpp and GsiSatBiasReader.h are the names of the files (note the camel case)
I suspect this passed for @srherbener because MacOS is case-insensitive?

Fixed so it will compile on Hera (and presumably elsewhere)

### DIFF
--- a/src/satbias/CMakeLists.txt
+++ b/src/satbias/CMakeLists.txt
@@ -4,7 +4,7 @@
 # which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
 
 ecbuild_add_executable( TARGET  satbias2ioda.x
-                        SOURCES SatBiasConverter.cpp GsiSatbiasReader.cpp GsiSatbiasReader.h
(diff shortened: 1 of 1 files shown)


shlyaeva (approved):
@CoryMartin-NOAA thank you for catching this! Probably passed for me on mac for the same reason.

aerorahul (approved):
Thanks for catching this.
It slipped my review on macOS.

srherbener (approved):
Thanks for fixing this.

Slipped by me on my Mac too. However, not sure how the CI testing all passed. I thought the AWS compute instances were more like linux systems, but I may be wrong.

shlyaeva:
Looking at CLang one, I think it's not built in CI. At least Clang CI doesn't include oops & ioda.

aerorahul:
You are right @shlyaeva 
The CI does not yet test components in this repository that depends on OOPS and IODA.
As a result, all of the Bufr-converter tests are not built and tested.

I believe @mer-a-o is working on updating the CI to use ioda-bundle to test this repository.

PR #439, opened 2021-02-23
Add support of global attributes 

### LINKED ISSUE #420: Add support to user defined variables and attributes when using the new c++ converter
## Description

Need the capability to add certain variables or attributes created by the user rather than transferring from their bufr equivalents. 

For example: 
1. Add metadata such as`platformCategory` or `instrumentName` for designated platform/instrument as a global attribute to the output netcdf file. 
2. Add variable such as: `Center Wavelength` to specify the wavelength of each radiance channel.

## Requirements
None

Capable of defining and adding variables when output the ioda netcdf file. 

## Dependencies

None

cyberbass:
## Description

Adding support of string type global attributes to the output netcdf file.
The changes to the yaml file is shown in the test yaml for MHS.

```yaml

    ioda:
      backend: netcdf
      obsdataout: "./testrun/gdas.t18z.1bmhs.tm00.nc"

      dimensions:
        - name: "nlocs"
          size: variables/radiance.nrows
        - name: "nchans"
          size: variables/radiance.ncols

      globals:
        - name: "centerFrequency@MetaData"
          type: floatVector
          value: [89.0, 157.0, 183.311, 183.311, 190.311]

        - name: "platformCommonName@MetaData"
          type: string
          value: "MHS"

```

Can also be used to add float and int types of global attributes. Example yaml section:
```yaml

        - name: "testfloat"
          type: float
          value: 1.0

        - name: "testfloatVector"
          type: floatVector
          value: [1.0, 2.0, 3.0]

        - name: "testint"
          type: int
          value: 5

        - name: "testintVector"
          type: intVector
          value: [5, 10, 15]

```

### Issue(s) addressed

- fixes #420

If this does not close an existing issue, then [be sure to add a Pipeline, Label, Estimate, Assignees, and Epic](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/practices/issues.html)

Global attributes as shown by HDF view when using the test yaml: 
Screen Shot 2021-02-23 at 9 14 35 AM

## Dependencies

None

## Impact

None

## Test Data

None


rmclaren:
In the YAML. Instead of "key" (which implies a hash table key) could we please call it "name"? This would then be consistent with the terminology we use  in the other sections of the code.

src/bufr/IodaEncoder/IodaDescription.h
```diff
@@ -41,8 +42,16 @@ namespace Ingester
         int compressionLevel;  // Optional
     };
 
+    struct GlobalDescription
```
rmclaren, line 45: If you feel comfortable with c++ templates consider making this a template so that the value can be of the type specified in the config. That way you could get rid of the "type" property altogether and it would save you work later converting strings to number types etc... Be aware if you do this you will need to make a GlobalDescriptionBase (non-template) to inherit from so you can put them in lists as shared_ptrs (not the direct values or you might end up slicing the object later). ex:

```cpp
 // IodaDescription.h

 struct GlobalDescriptionBase
 {
    std::string name;
 };

 template<typename T>
 struct GlobalDescription : GlobalDescriptionBase
 {
    T value;
 };

 //....
 std::vector<std::shared_ptr<GlobalDescriptionBase>> globals_;


 // IodaDescription.cpp

 if (conf.getString(ConfKeys::Global::Name) == ConfKeys::Global::StringType)
 {
     auto global = std::make_shared<GlobalDescription<std::string>>();
     global->name = // get the name

     // get the value string from the YAML.
     // validate it and convert it to the correct type (string is trivial)
     global->value = my_value 

     globals_.push_back(global)
 }
```

If all this sounds really overwhelming you can keep it the way it is.

rmclaren (changes requested):
Overall looks good. the biggest thing is to enforce the type information we provide in the YAML file, and to be consistent with naming, and .

src/bufr/IodaEncoder/IodaDescription.h
cyberbass: Thank you Ron! I have been working on implementing this but didn't quite know how to use the C++ templating. What you have here really helps and hopefully I can get this done correctly.

rmclaren: No problem, let me know if you run into any issues.

cyberbass: @rmclaren I ran into a problem when defining vector 
```cpp
std::vector<std::shared_ptr<GlobalDescriptionBase>> globals_;
```
The compiler asks me to include arguments for GlobalDescriptionBase and I wrote 
```cpp
std::vector<std::shared_ptr<GlobalDescriptionBase<std::string>>> globals_;
```
to make it compile. But after this I found it converts all values in struct ```GlobalDescription``` into ```cpp std::string```. 

The latest commit I implemented it in a different way which is closer to the original code, but the issue remains. When I try to print the type of the value after getFloat it gives me this: 
```
NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```
I suppose it is string. 

Currently the string option completely works but the float option doesn't work.

rmclaren: Could you please push your code. It sounds like you made your GlobalDescriptionBase a templated struct... you do not want this. Only its bubclass GlobalDescription should be templated.

rmclaren: What you have in the repo right now won't work. Please look at the coding example I provided.

rmclaren: Don't forget to #include <memory> at the top of the file.

cyberbass: I was trying and error and pushed a wrong version I think. I'll go back to the previous version which is following your example.

rmclaren: As a side note: You actually don't need to include type_traits in IodaDescription.h, that was kinda a typo on my part.

cyberbass: I just rewinded it to the previous commit and added the code try to handle floats. The two problems I have are as of below: 
I can't compile the code when writing the if statement in ``` IodaEncoder.cpp``` as: 
```cpp
if (std::is_same<std::string, global->value>::value)
{
}
else if (std::is_same<float, global->value>::value)
{
}
```
The compiler complains on ```global``` not being a typename. So I passed the ```global->type``` to get around it. 

Then if I implemented the above work around, I can't compile the code when I write the float attributes: 
```cpp
attr.write<float>({global->value});
```
The compiler will complain about the argument type. I printed its value and type above and the result is as below: 
```
e
NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```
I think this means it is still a string which goes back to the problem I talked about above.

rmclaren: So one problem is line 77 in IodaDescription.h: std::vector<std::shared_ptr<GlobalDescription<std::string>>> globals_;

You actually want to specify GlobalDescriptionBase not GlobalDescription<std::string>.

I'll go ahead and make some modifications to your code and check them into this repository to help speed things up a bit (will have to be later, you'll have my changes by Monday). If you want we can have a call and I can explain what is going on.

Basically a template struct (or class), is like a prototype for that class.  So when you specify a template class like above GlobalDescription<std::string> the compiler creates that class for you at compile time (there is literally now a new kind of class GlobalDescription__String (we could call it)). Templates are literally functions and classes which tell the compiler how to automatically generates new functions and classes on the fly :) So its almost like a meta programming language, literally code that the compiler uses to create more code. As such your mindset has to be a bit different when thinking about these things. For instance you can't just stick a template in a like this vector<GlobalDescription>, because that thing GlobalDescription is literally not defined (its a template for a class, not a class). GlobalDescritpion<String> of course is actually a class, but if we stick that in a vector (like above) then that vector can only hold the version of GlobalDescription the compiler made that can hold strings :) So what to do... well we can use a concept called polymorphism to make a vector that can hold any kind of GlobalDescription. When one object inherits from another object it **is a** type of that object. So the way we've arranged things above GlobalDescription<String>, GlobalDescription<Float>, and so on are all types of GlobalDescriptionBase, even though the compiler generated them on the fly. Cool, so we can have the compiler make a vector<shared_ptr<GlobalDescriptionBase>> for us, so we can hold any kind of global description. Anyways, I could go one. We can talk more on Monday. Take care!

cyberbass: Thank you so much Ron!! That's actually where I suspect the problem is but don't know how to correct it. I have spent quite some time on it and It looks like the compiler doesn't like me to pass explicit argument when defining that vector, nor it allows me to pass no argument as well. If you can quickly show me how to work around I think will help a lot.  Thanks again!

rmclaren:
@cyberbass So I went ahead and fixed the implementation of the support for Global variables. Let me know if you are curios as to how it works and we can talk about it.

cyberbass:
> @cyberbass So I went ahead and fixed the implementation of the support for Global variables. Let me know if you are curios as to how it works and we can talk about it. Don't forget to add support for addtional types you want to support (ex: int).

Thank you! I'll finish the rest and will be asking you how this was done when you get time.

src/bufr/IodaEncoder/IodaDescription.h
cyberbass: @rmclaren  I have finished the implementation of int type as well as the floatVector and intVector. However, when I test it to write a vector type of attribute it will report the following error: 
```
terminate called after throwing an instance of 'std::logic_error'
  what():  Bad assertion / unsupported fundamental type at the frontend side of the ioda type system.
Aborted
```
I believe this has something to do with the Ioda frontend. I printed the vector type and vector values and they look good to me but just can't use ```ioda::Attribute attr = group.atts.create<T>(name, {vectorLength});``` to create the attribute. Maybe @srherbener can help here? 
To reproduce the error, just add a global attribute in the global section as: 
```yaml
      globals:
        - name: "sensorCentralWavelength@MetaData"
          type: floatVector
          value: [101.5, 102.5, 103.5]
```

cyberbass:
@rmclaren  I have finished the implementation of int type as well as the floatVector and intVector. However, when I test it to write a vector type of attribute it will report the following error: 
```
terminate called after throwing an instance of 'std::logic_error'
  what():  Bad assertion / unsupported fundamental type at the frontend side of the ioda type system.
Aborted
```
I believe this has something to do with the Ioda frontend. I printed the vector type and vector values and they look good to me but I just can't use ```ioda::Attribute attr = group.atts.create<T>(name, {vectorLength});``` to create the attribute. Maybe @srherbener can help here? 
To reproduce the error, just add a global attribute in the global section as: 
```yaml
      globals:
        - name: "sensorCentralWavelength@MetaData"
          type: floatVector
          value: [101.5, 102.5, 103.5]
```

rmclaren:
@cyberbass Sorry Nan, I was a little busy trying to get something else merged into the  NCEPLIB-bufr library. I'll try to take a look at it today.

srherbener:
@cyberbass my apologies for the slow response. I'm in a code sprint this week so I likely won't be able to get to this until next Monday.

cyberbass:
> @cyberbass Sorry Nan, I was a little busy trying to get something else merged into the NCEPLIB-bufr library. I'll try to take a look at it today.

No problem, thank you!

cyberbass:
> @cyberbass my apologies for the slow response. I'm in a code sprint this week so I likely won't be able to get to this until next Monday.

Thanks Steve!

rmclaren:
@cyberbass Ok, I checked in something that I think will make the GlobalDescription template able to deal with both vector and non-vector types (and use the ioda functions the right way in both cases). I'm using whats called  SFINAE (Substitution failure is not an error) to get the compiler to generate the correct version of an private _addTo method in both cases. Let me know how it goes...

cyberbass:
> @cyberbass Ok, I checked in something that I think will make the GlobalDescription template able to deal with both vector and non-vector types (and use the ioda functions the right way in both cases). I'm using whats called SFINAE (Substitution failure is not an error) to get the compiler to generate the correct version of an private _addTo method in both cases. Let me know how it goes...

Thank you so much I am checking it out now.

cyberbass:
> > @cyberbass Ok, I checked in something that I think will make the GlobalDescription template able to deal with both vector and non-vector types (and use the ioda functions the right way in both cases). I'm using whats called SFINAE (Substitution failure is not an error) to get the compiler to generate the correct version of an private _addTo method in both cases. Let me know how it goes...
> 
> Thank you so much I am checking it out now.

@rmclaren I added int and intVector type, fixed an issue with vector length. I think it is ready for review now.

src/bufr/IodaEncoder/IodaDescription.h
```diff
@@ -41,8 +44,52 @@ namespace Ingester
...
+            ioda::Attribute attr = group.atts.create<T>(name, {1});
+            attr.write<T>({value});
+        }
+
+        // T is a vector
+        template<typename U = void>
+        void _addTo(ioda::Group& group,
+                    std::enable_if_t<is_vector<T>::value, U>* = nullptr)
+        {
+            ioda::Attribute attr = group.atts.create<typename T::value_type>(name, {vectorSize});
```
rmclaren, line 85: Instead of `{vectorSize}` you can just call `value.size()` (just assume the value is a vector, the compiler will let you know if you made a mistake).

src/bufr/IodaEncoder/IodaDescription.h
```diff
@@ -57,6 +104,9 @@ namespace Ingester
         /// \brief Add Variable defenition
         void addVariable(const VariableDescription& variable);
 
+        /// \brief Add Globals defenition
+//       void addGlobal(const GlobalDescription& global);
```
rmclaren, line 108: You are now ready to implement this method. You can pass it a `const std::shared_ptr<GlobalDescriptionBase>&` and go from there.

src/bufr/IodaEncoder/IodaEncoder.cpp
```diff
@@ -8,6 +8,7 @@
 #include "IodaEncoder.h"
 
 #include <memory>
+#include <type_traits>
```
rmclaren, line 11: Not sure you need to include this here... I would try compiling without it.

rmclaren (changes requested):

src/bufr/IodaEncoder/IodaDescription.h
cyberbass: I tried this but it doesn't work. The compiler will complain something.

src/bufr/IodaEncoder/IodaEncoder.cpp
cyberbass: I don't know, maybe something from the previous commit not purged.

rmclaren: If you need the include its fine to leave it in. I was just wondering if it could be removed, as I don't like having unused includes in files.

src/bufr/IodaEncoder/IodaDescription.h
rmclaren: Let me know how it goes.

cyberbass: This is what I got: 
```
/data/users/nchen/CodeSprintDec20/src/ioda-bundle/iodaconv/src/bufr/IodaEncoder/IodaDescription.h(85): error #3291: invalid narrowing conversion from "unsigned long" to "long"
              ioda::Attribute attr = group.atts.create<typename T::value_type>(name, {value.size()});
```

cyberbass: I found if I use int(value.size()) then it will compile. Don't quite understand why I need to do so but I'll eliminate that extra variable then.

src/bufr/IodaEncoder/IodaDescription.cpp
```diff
@@ -150,6 +163,60 @@ namespace Ingester
...
+                  auto global = std::make_shared<GlobalDescription<std::vector<int>>>();
+                  global->name = globalConf.getString(ConfKeys::Global::Name);
+                  global->value = globalConf.getIntVector(ConfKeys::Global::Value);
+                  global->vectorSize = global->value.size();
+                  globals_.push_back(global);
+              }
+              else
+              {
+                 throw eckit::BadParameter("Unsupported global attribute type");
+              }
```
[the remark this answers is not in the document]
cyberbass: I implemented it similarly as the variable section.

src/bufr/IodaEncoder/IodaDescription.h
rmclaren: So the type returned by value.size() is a size_t which is actually an "unsigned long". The problem is that the function group.atts.create takes a "long" which is a little surprising actually :) So instead of int(value.size()) the more correct thing would be to do static_cast<long>(value.size()). Thanks for figuring this out!

cyberbass: Yeah I used ``` static_cast<int>(value.size())``` and passed all the test. I think a regular int should be good enough for specifying the length of this vector in our case.

test/testinput/bufr_mhs.yaml
```diff
@@ -40,7 +57,41 @@ observations:
         - name: "nchans"
           size: variables/radiance.ncols
 
+      globals: 
+        - name: "sensorCentralFrequency@MetaData"
+          type: string
+          value: "89.0, 157.0, 183.311 +- 1.00, 183.311 +- 3.00, 190.311"
```
emilyhcliu, line 63: @cyberbass 
I see that you want to put **sensor/channel-related information** under the global attribute.  Is this the direction we are heading? 

If so, we will need to add polarization (for MW) and original channel numbers (for all sensor types).

gthompsnJCSDA: Emily, in IODA variable naming convention, MetaData sheet, I created a variable called `sensorPolarizationDirection` that could do the job here.  But, now that I see what Nan typed in for this example, it reveals a small issue.  I had envisioned a list of frequencies in which each channel is a different freq, as is more typical on many platforms.  But this example has different tolerance ranges on the same freq (183GHz +- ??).  So perhaps another new addition is needed for channel tolerance or similar, because I wasn't expecting we would want to use strings to represent what are mostly floats.

cyberbass: I think what I have there is just an example, we can fix it after discussion and I will update this PR.

rmclaren (approved):

test/testinput/bufr_mhs.yaml
emilyhcliu: @cyberbass @gthompsnJCSDA @rmclaren 
In our re-engineering meeting today, I mentioned to @rmclaren that we will need the unit information for variables defined under global attributes. 
For example, 
the  globals: 
        - name: "sensorCentralFrequency@MetaData"
          type: string
          value: "89.0, 157.0, 183.311 +- 1.00, 183.311 +- 3.00, 190.311"

We need a unit (GHz) for this variable.  For IR, the unit could be wavelength (cm or micron).  

Another question I have is the data type for sensorCentralFrequency@MetaData.
The type is string.  This means that the users have to convert it to float type to use in numerical operations in UFO.  Can we define this variable directly to float type?

cyberbass: @emilyhcliu I think we can define that directly into float type or float vectors but we can't add unit information to it. 
From the IODA global attributes examples I don't see there is a way to add units to a float type of global attributes: 

https://github.com/JCSDA-internal/ioda/blob/develop/src/engines/Examples/Basic/02-Attributes.cpp

Unless @rmclaren can think of a way to get around it.

gthompsnJCSDA: Nan:  can we just change this from a string of frequencies into a vector of floating point values, even if the 183GHz is repeated?  I want to see this PR approved asap.  I suggest we sideline the discussion of ways to deal with polarization and tolerance or other "edge cases" until later.  This is needing to get done and we can come back to those edge cases with different solutions later.  Other variables and/or attributes will be needed.

cyberbass: @gthompsnJCSDA Yeah I would like to do that since we should get this PR in ASAP and to fully finish the discussion here may require some more time.

emilyhcliu: @gthompsnJCSDA @cyberbass 
We can complete this PR as is.
We can open a new issue to continue the discussion.

emilyhcliu (approved):

test/testinput/bufr_mhs.yaml
cyberbass: I just pushed the change to the yaml and the testoutput file in. I think we should be good to go.

Issue #418, opened 2021-02-03
Bufr Converter: Add support for standard WMO bufr files.


rmclaren:
Add support for reading standard WMO bufr files. Standard WMO Bufr files do not include the BUFR table data as part of the file (must be read seperatly).

PR #455, opened 2021-03-03
Add support for standard WMO BUFR Files.

rmclaren:
## Description

Adds support for standard WMO Bufr files. Also includes updates for latest version of NCEPLIB-bufr develop where they changed the name of a header file (uuuugggg).

### Issue(s) addressed
#418


src/bufr/BufrParser/BufrCollectors/BufrIntCollector.cpp
```diff
@@ -5,7 +5,7 @@
  * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-#include "bufr.interface.h"
+#include "bufr_interface.h"
```
CoryMartin-NOAA, line 8: does this mean that this will only work with the latest BUFR develop branch now? Can we insure backwards and forwards compatibility somehow?

src/bufr/BufrParser/BufrDescription.cpp
```diff
@@ -21,6 +21,8 @@ namespace
     namespace ConfKeys
     {
         const char* Filename = "obsdatain";
+        const char* IsStdFormat = "isStdFormat";
```
CoryMartin-NOAA, line 24: suggest isWMOFormat instead of isStdFormat, just to be explicit.

test/CMakeLists.txt
```diff
@@ -87,17 +87,20 @@ endif()
 
 if( iodaconv_bufr_ENABLED )
   list( APPEND test_input
+    testinput/bufr_tables
     testinput/gdas.t18z.1bmhs.tm00.bufr_d
     testinput/gdas.t00z.1bhrs4.tm00.bufr_d
     testinput/gdas.t06z.adpsfc.tm00.bufr_d
     testinput/bufr_read_2_dim_blocks.bufr
+    testinput/20210203.radiosonde
```
CoryMartin-NOAA, line 95: This is just YYYYMMDD, is this for a particular hour/cycle or the full day?

test/testinput/bufr_read_std_wmo_radiosonde.yaml
```diff
@@ -0,0 +1,82 @@
...
+  - obs space:
+      name: bufr
+      obsdatain: "./testinput/20210203.radiosonde"
+      isStdFormat: true
+      tablepath: "./testinput/bufr_tables"
+
+      mnemonicSets:
+        - mnemonics: [CLATH, CLONH]
+        - mnemonics: [TMDB, TMDP, WDIR, WSPD]
+          channels: 1-5880
```
CoryMartin-NOAA, line 16: This illustrates why 'channels' is a bad word to use here... but tackle that in a future PR, obviously.

CoryMartin-NOAA (changes requested):
Two main comments:

- suggest changing isStdFormat (and similar vars) to isWMOFormat, to be explicit that this is the WMO format, as one could argue the old way is the standard NCEP format
- if going from bufr.interface.h to bufr_interface.h is only in NCEPLIBS-BUFR develop, then this will break backwards compatibility, no? Is there a way we can include either depending on version? Perhaps some cmake/ecbuild magic?

src/bufr/BufrParser/BufrCollectors/BufrIntCollector.cpp
rmclaren: Quick question (related to your request). What is the minimum version of compiler wee support for intel, gcc, micrsoft.. etc...

rmclaren: The easiest way to do what you asked is by using the preprocessor __has_include method... but some older compilers might not support it.

CoryMartin-NOAA: that would be a question for someone like @srherbener or @rhoneyager or @mmiesch maybe

aerorahul: intel 17 and gcc 7.3 are the oldest we have built our stuff on.
I think the current stacks are probably atleast intel 18 and gnu 9.x

src/bufr/BufrParser/BufrDescription.cpp
aerorahul: curious, what is considered "Std"

src/bufr/BufrParser/BufrCollectors/BufrIntCollector.cpp
rmclaren: Ryan is using it as part of the HH library, so maybe its ok.

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -139,18 +143,37 @@ namespace Ingester
...
+        {
+            openbf_f(fortranFileId_, "SEC3", fortranFileId_);
+
+            if (!tablepath.empty())  // else use the default tables
+            {
+                table1FileId_ = fortranFileId_ + 1;
+                table2FileId_ = fortranFileId_ + 2;
+                mtinfo_f(tablepath.c_str(), table1FileId_, table2FileId_);
+            }
+        }
```
aerorahul, line 167: I am not loving this `if-else` business all over the place.

test/CMakeLists.txt
aerorahul: where did this file come from?
WMO?  If so, we should atleast add something in the filename string a source of the data.

.gitattributes
```diff
@@ -9,3 +9,5 @@ src/ncep/xx120 filter=lfs diff=lfs merge=lfs -text
 *.bufr_d filter=lfs diff=lfs merge=lfs -text
 *.grib filter=lfs diff=lfs merge=lfs -text
 *.grib2 filter=lfs diff=lfs merge=lfs -text
+test/testinput/bufr_tables/** filter=lfs diff=lfs merge=lfs -text
```
aerorahul, line 12: what are these bufr tables and where are they coming from?
Are these copies from nceplibs-bufr? Somewhere else?
I don't think bufr tables belong in the ioda-converters repo.

aerorahul (changes requested):
I have several questions on this PR. See below.

src/bufr/BufrParser/BufrCollectors/BufrIntCollector.cpp
rmclaren: Support for __has_include was added in intel 18 (been part of gcc since version 5)

.gitattributes
CoryMartin-NOAA: How else will you test this new feature?

src/bufr/BufrParser/BufrDescription.cpp
rmclaren: So the NCEP bufr files we use here have the BUFR table included within the first couple of messges inside these BUFR files (basically the information you get when you do gettab.x or call dxdump). The "standard" WMO BUFR files do not include this information as part of the BUFR file, they get this info from external tables which are loaded seperatly. "STD" is the version of BUFR that follows the official WMO BUFR standard... I guess.

test/CMakeLists.txt
rmclaren: ok, greg sent me this file a little while ago.

.gitattributes
rmclaren: These tables are necessary in order to open the Std or WMO BUFR files. The data from these tables is not included in those files and must be side loaded.

src/bufr/BufrParser/BufrCollectors/BufrIntCollector.cpp
mmiesch: Yes - you can count on intel 18+ and gcc 5+

.gitattributes
aerorahul: can we add these tables in the `NCEPLIBS-bufr` repo.  That repo contains tables [here](https://github.com/NOAA-EMC/NCEPLIBS-bufr/tree/develop/tables). What are those? And are they different than these? I can see Jeff Ator not wanting to add to his maintenance burden.  We can propose. On the other hand, I don't think the burden of maintaining WMO bufr tables should fall on this repo either.

CoryMartin-NOAA: Or maybe just the table(s) necessary to read the one test file we have for the unit test? I think it should be self contained here for the unit testing capability, after all we have numerous .nc, .bufr, etc. files in this repo that are fairly specific.

aerorahul: I looked in the nceplibs-bufr repo and most of these tables are identical there.
I suggest we consolidate.  
nceplibs-bufr can read any "std" bufr file by providing an external table.

src/bufr/BufrParser/BufrParser.cpp
rmclaren: @aerorahul This is quiet literally the only place in the program where we check isStdFormat and do anything with that information (I don't know what you mean by "all over the place"). WMO BUFR have a slightly different procedure to open the files which we handle succinctly in this one if statement which gets called exactly once, otherwise there is no difference at all with how you deal with these files. The altrernative would be orders of magnitude more complex. It would mean to create a BufrParser, NcepBufrParser and WMOBufrParser classes (the two new classes would do nothing other than open the file in a slightly different way), but then you'd still need an if statement in order to instantiate the correct flavor based on the contents of the YAML file. So how do you work around that? Well we add a factory to construct the correct version of the BufrParser based on the information we get from the YAML file. So now you've created at least 2 classes and a full fledged factory just to avoid 1 if statement.

It would make sense to have different BufrParser subclasses if these files needed to be treated differently and the code if (isStdFormat) appeared in multiple places, but we just don't.

.gitattributes
rmclaren: So it turns out Jeff added those tables to their repo very recently, and they are being installed in in a tables directory when you do the `make install`. I guess the question is how we could figure out where this directory was (wish NCEPLIB-bufr would just read the tables automatically). I'll ask Jeff.

The BUFR message has in its message header information about what version of bufr table to use (both WMO and NCEP maintain slightly different versions of these tables). NCEPLIB-bufr looks for tables that have the correct name based on this info (can't just consolidate tables without code changes to NCEPLIB-bufr). Our code simply points it to the directory where the tables are stored.

src/bufr/BufrParser/BufrCollectors/BufrIntCollector.cpp
rmclaren: Thanks @mmiesch! I went ahead and used __has_include to check the header files. I think in the future we will be able to remove this so its only temporary.

test/CMakeLists.txt
gthompsnJCSDA: For my feature branch doing more tests of WMO-style BUFR files, I am using names like this:
  synop_wmo_single.bufr
in which a single BUFR obs of a SYNOP is contained in the sample file.  So you could consider making yours "sonde_wmo_single.bufr" which I probably already have such a file in my branch.

In my own future work, then I plan to make 'multiple' replace single and change WMO reference to prepBUFR or something similar to handle next phases of this work.

.gitattributes
rmclaren: In CMake is there any way to get the install path for a package (bufr in my case) that has been previously found with find_package? Tried a bunch of things... but no dice. My goal is to get the **<bufr install path>/tables** directory and symlink it into the testinput directory at build time...

gthompsnJCSDA: I will tag @mmiesch as he may know better.  Also, a quick question. I have been using your own repo for bufrlib, but will your own mods go back into the master branch of NCEP's bufr?

rmclaren: The WMO related updates are actually there now.

gthompsnJCSDA (approved):
Things are looking very good to me.  I am not a fan of `bufr_read_radiosonde_wmo` for yaml or named bufr/nc file because I had devised a shorter naming schema for 7 other conventional data types, but that is relatively minor point.  I can begin exercising this code with other conventional data that does not need to handle sequences/replications pretty soon after this is merged.

.gitattributes
rhoneyager: @rmclaren @aerorahul nceplibs-bufr should make a small change to their [PackageConfig.cmake.in file](https://github.com/NOAA-EMC/NCEPLIBS-bufr/blob/develop/cmake/PackageConfig.cmake.in) to also provide the path to the tables installation. [See here](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#id14) for an example. By defining ```set(bufr_TABLES_DIR ${PREFIX}/tables)``` in the package config file, a downstream consumer can use this variable when making the symlink.

aerorahul: Thanks for the tip. Will open a PR shortly.

rhoneyager: Having symbolic links all over the place is messy and prevents us from having independent, relocatable packages. Of course, environment variables are also equally problematic.

Really, though, jedi-wide we need a good policy on how to manage resource data. Similar issues occur in ufo, rttov, crtm, etc. whenever coefficient files are used. These coefficients are *not* test data and are required at runtime. @ytremolet & @mmiesch this is definitely a discussion worth having.

srherbener:
The CI tests are failing due to the version of the NCEPLIBS-bufr repo that is being used in the test container. I've built and tested on my local clone using the latest develop of the NOAA-EMC/NCEPLIBS-bufr and all the test pass except for the coding norms:
```
3780 481/529 Testing: test_iodaconv_bufr_coding_norms
3781 481/529 Test: test_iodaconv_bufr_coding_norms
3782 Command: "/Users/stephenh/projects/BUFR_CONV/ioda-bundle/build/bin/iodaconv_cpplint.py" "--quiet" "--recursive" "/U     sers/stephenh/projects/BUFR_CONV/ioda-bundle/iodaconv/src/bufr"
3783 Directory: /Users/stephenh/projects/BUFR_CONV/ioda-bundle/build/bin
3784 "test_iodaconv_bufr_coding_norms" start time: Mar 19 10:11 MDT
3785 Output:
3786 ----------------------------------------------------------
3787 /Users/stephenh/projects/BUFR_CONV/ioda-bundle/iodaconv/src/bufr/BufrParser/BufrCollectors/BufrRepCollector.cpp:8:       Missing username in TODO; it should look like "// TODO(my_username): Stuff."  [readability/todo] [2]
3788 /Users/stephenh/projects/BUFR_CONV/ioda-bundle/iodaconv/src/bufr/BufrParser/BufrCollectors/BufrIntCollector.cpp:8:       Missing username in TODO; it should look like "// TODO(my_username): Stuff."  [readability/todo] [2]
3789 /Users/stephenh/projects/BUFR_CONV/ioda-bundle/iodaconv/src/bufr/BufrParser/BufrParser.cpp:16:  Missing username in      TODO; it should look like "// TODO(my_username): Stuff."  [readability/todo] [2]
3790 <end of output>
3791 Test time =   1.24 sec
3792 ----------------------------------------------------------
3793 Test Failed.
3794 "test_iodaconv_bufr_coding_norms" end time: Mar 19 10:11 MDT
3795 "test_iodaconv_bufr_coding_norms" time elapsed: 00:00:01
3796 ----------------------------------------------------------
```

We are still using the jcsda-internal fork of NCEPLIBS-bufr for the sole reason that we need a tag on the NOAA-EMC/NCEPLIBS-bufr repo that is newer than the "11.4.0" tag since we need some fixes that have been merged into develop since 11.4.0.

For now I've put a new tag (11.4.0.jcsda2) on our fork that we can use for jedi-stack and the containers. The 11.4.0.jcsda2 tag is on today's sync of the develop branch from the NOAA-EMC/NCEPLIBS-bufr repo.

I'm going to submit a PR to jedi-stack and put in the request to rebuild the test containers using the new 11.4.0.jcsda2 tag. This should get the CI testing working properly.

srherbener:
I think the code looks good, and I'm okay approving after the coding norms errors are fixed.

@CoryMartin-NOAA and @aerorahul if you are good with the changes @rmclaren made please approve so that the change requests are cleared out.

Thanks!

CoryMartin-NOAA (approved):
I think all my comments/concerns have been addressed.

srherbener (approved):
Coding norms passes on my iMac (plus all the other tests). Looks good!

aerorahul (approved):

srherbener:
One issue remains, unrelated to these code changes, which is that if this gets merged in now the CI testing will break for all the other PRs. This is because the NCEPLIBS-bufr version needs to get updated in the test containers. With the current version of NCEPLIBS-bufr in the test containers, this code will not compile successfully.

There is a ZenHub issue in jedi-stack (jcsda-internal/jedi-stack#67) for this issue, and it would be best to wait until this issue is resolved before merging. Sorry about this delay.

Thank you for your patience with this!

aerorahul:
Can we disable the bufr converter till that issue is resolved and then enable it? 
This will allow others to build what they need. 

Alternately, we can add NCEPlibs-bufr to the bundle and build it each time. It is quick. I imagine we will be updating the bufr library again. 
Thoughts?

srherbener:
> Can we disable the bufr converter till that issue is resolved and then enable it?
> This will allow others to build what they need.
> 
> Alternately, we can add NCEPlibs-bufr to the bundle and build it each time. It is quick. I imagine we will be updating the bufr library again.
> Thoughts?

My preference is to get the test containers updated so that we don't have to conditionally build the bufr converter code. See jcsda-internal/jedi-stack#67

If we need an immediate fix for this, I like the idea of placing a switch in the CMakeLists.txt file to conditionally enable the bufr converter build to hold us over until the containers are updated. We can have the build disabled by default and that would allow the other converters to be tested using the current container.

@mer-a-o, @mmiesch is it possible to get the containers updated soon. If not, then I think we should proceed with putting in the switch as described above.

Thanks!

srherbener:
Just want to let you know that @mer-a-o is working on getting the test containers updated with the latest NCEPLIBS-bufr version.

rmclaren:
Thanks so much @srherbener, @mer-a-o and @aerorahul!

mer-a-o:
@srherbener all CI containers are updated with the new NCEPBUFR-lib. All the tests passed when I tested the containers. Is this test `test_iodaconv_bufr_read_wmo_radiosonde` failure expected?

srherbener:
@mer-a-o thank you for all your hard work to get the test containers updated. The test failure is unexpected. Here is a sample from the log file:
```
test 48
975 | Start 48: test_iodaconv_bufr_read_wmo_radiosonde
976 |  
977 | 48: Test command: /build_container/bin/iodaconv_comp.sh "netcdf" "/build_container/bin/bufr2ioda.x testinput/bufr_read_wmo_radiosonde.yaml" "bufr_read_wmo_radiosonde.nc" "0.0"
978 | 48: Environment variables:
979 | 48:  OMP_NUM_THREADS=1
980 | 48: Test timeout computed to be: 10000000
981 | 48: terminate called after throwing an instance of 'std::bad_alloc'
982 | 48:   what():  std::bad_alloc
983 | 48: /build_container/bin/iodaconv_comp.sh: line 20: 23671 Aborted                 $cmd
984 | 1/1 Test #48: test_iodaconv_bufr_read_wmo_radiosonde ...***Failed    0.02 sec
985 |  
986 | 0% tests passed, 1 tests failed out of 1
987 |  
988 | Label Time Summary:
989 | iodaconv    =   0.02 sec*proc (1 test)
990 | script      =   0.02 sec*proc (1 test)
991 |  
992 | Total Test time (real) =   0.02 sec
993 |  
994 | The following tests FAILED:
995 | 48 - test_iodaconv_bufr_read_wmo_radiosonde (Failed)
996 | Errors while running CTest
997
```

Not sure why the std::bad_alloc is being thrown. I'm going to try running on my Mac and see if the same happens.

srherbener:
This test passes on my Mac. @mer-a-o would I be able to debug this by starting up docker with the container matching what the new CI containers have? If so, how would I do this? Thanks!

mer-a-o:
@srherbener I pushed the recent clang docker container to dockerhub: `jcsda/docker-clang-mpich-dev:latest`. Would it be helpful for debugging?

srherbener:
> @srherbener I pushed the recent clang docker container to dockerhub: `jcsda/docker-clang-mpich-dev:latest`. Would it be helpful for debugging?

Thanks @mer-a-o! I'll take a look.

srherbener:
@mer-a-o I wonder if the CI test failures are due to running out of memory. I think the containers are good given that the CI testing in #485 is working.

srherbener:
The tests are still getting the failures with the std::bad_alloc exception:
```
test 48
--
975 | Start 48: test_iodaconv_bufr_read_wmo_radiosonde
976 |  
977 | 48: Test command: /build_container/bin/iodaconv_comp.sh "netcdf" "/build_container/bin/bufr2ioda.x testinput/bufr_read_wmo_radiosonde.yaml" "bufr_read_wmo_radiosonde.nc" "0.0"
978 | 48: Environment variables:
979 | 48:  OMP_NUM_THREADS=1
980 | 48: Test timeout computed to be: 10000000
981 | 48: terminate called after throwing an instance of 'std::bad_alloc'
982 | 48:   what():  std::bad_alloc
983 | 48: /build_container/bin/iodaconv_comp.sh: line 20: 23673 Aborted                 $cmd
984 | 1/1 Test #48: test_iodaconv_bufr_read_wmo_radiosonde ...***Failed    0.02 sec
985 |  
986 | 0% tests passed, 1 tests failed out of 1
987 |  
988 | Label Time Summary:
989 | iodaconv    =   0.02 sec*proc (1 test)
990 | script      =   0.02 sec*proc (1 test)
991 |  
992 | Total Test time (real) =   0.02 sec
993 |  
994 | The following tests FAILED:
995 | 48 - test_iodaconv_bufr_read_wmo_radiosonde (Failed)
996 | Errors while running CTest
```

srherbener:
Thanks @mer-a-o! The memory was the issue then. I noticed that the commit was empty so is the increase in memory a configuration you adjusted somewhere else? In other words is is safe to merge this PR into develop now? Thanks!

mer-a-o:
@srherbener Yes, increasing the memory in CI solved the issue. This is ready for merge.

srherbener:
> @srherbener Yes, increasing the memory in CI solved the issue. This is ready for merge.

@mer-a-o thank you for all your hard work to get the CI test containers updated! It is much appreciated.

PR #456, opened 2021-03-03
Assign dimensions to all variables in satbias files

rhoneyager:
## Description

This is a small fix to the bias files. See title.

## Test data

Once this PR is in, satbias files should be regenerated for all instruments. ufo develop can work with either the existing files or the new ones, but the ioda sprint needs the new files.


wsmigaj (approved):

srherbener (approved):
thanks for this fix!

Issue #461, opened 2021-03-09
Add mnemonic GNAPS to  satwnds_new_subset_template.yaml


ilianagenkova:
### **Objective** - Adding GNAPS to satwnds_new_subset_template.yaml 
---

### **Background Info** 

Mnemonics GNAP (in old bufr satwnd files) and GNAPS (in new bufr satwnd files) determine now to interpret the multiple PCCF occurrences in a satwnd bufr file. 
For example, 
if  GNAPS=4  the PCCF will contain the Common quality index (QI) without forecast
if  GNAPS=5 the PCCF will contain QI without forecast
if GNAP=4  the PCCF contains the Expected Error
if GNAP=5  the PCCF conatins the QI (Quality Indicator) excluding the forecast. 

Therefore, we can't rely on hardwired order (as it is in GSI now) to interpret what PCCF holds. For example in read_satwnd.f90 : 
call ufbseq(lunin,amvqic,2,4,iret, 'AMVQIC') ! AMVQIC:: GNAPS PCCF
       qifn = amvqic(2,2) ! QI w/ fcst does not exist in this BUFR
       ee = amvqic(2,4) ! NOTE: GOES-R's ee is in [m/s]

satwnds_old_subset_template.yaml already includes the GNAP mnemonic, i.e. "generating_application"

Here we are adding GNAPS to satwnds_new_subset_template.yaml 

### **Working Branch:**
[feature/iodaconv_GNAPS](https://github.com/JCSDA-internal/ioda-converters/tree/feature/iodaconv_GNAPS
)

emilyhcliu:
@ilianagenkova Thanks for working on this.  Please test on your branch and check if the output IODA contains GNAPS as expected.  If so, you can do a pull request based on your branch.

**There is one more thing related to this update and requires your attention.**
In February, you created a table ([Iliana's presentation on February 8, 2021](https://drive.google.com/file/d/1jL1aOcb7Qz5shx0_SmoniWiUkSXGh1rD/view)) that documents the switching dates from the old to the new BUFR format for various data types.  The table is a bit outdated for some of the data types.  Could you please update the information in the table.  I think the switching dates will need to include cycle information (e.g. 00Z)  as well.  Thanks!

PR #462, opened 2021-03-09
Added GNAPS mnemonic to satwnds_new_subset_template.yaml

ilianagenkova:
Branch: [/feature/iodaconv_GNAPS](https://github.com/JCSDA-internal/ioda-converters/tree/feature/iodaconv_GNAPS)

Issue: #461 
https://github.com/JCSDA-internal/ioda-converters/issues/461

This PR means to add mnemonic GNAPS to satwnds_new_subset_template.yaml for improved ioda-converters robustness. 

This PR will close issue #461

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -93,6 +93,8 @@ if( iodaconv_bufr_ENABLED )
     testinput/gdas.t00z.1bhrs4.tm00.bufr_d
     testinput/gdas.t06z.adpsfc.tm00.bufr_d
     testinput/bufr_read_2_dim_blocks.bufr
+    testinput/bufr_satwnd_new_format.bufr
+    testinput/bufr_satwnd_old_format.bufr
     testinput/bufr_mhs.yaml
     testinput/bufr_hrs.yaml
     testinput/bufr_filtering.yaml
@@ -110,6 +112,8 @@ if( iodaconv_bufr_ENABLED )
     testoutput/gdas.t18z.1bmhs.tm00.15.7.filter_split.nc
     testoutput/gdas.t06z.adpsfc.tm00.nc
     testoutput/bufr_read_2_dim_blocks.nc
+    testoutput/NC005031.nc
+    testoutput/NC005066.nc
   )
 endif()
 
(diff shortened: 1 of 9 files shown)


aerorahul (commented):
@ilianagenkova 
Can you also add a test where this is exercised?
If it is already covered in the tests, then you will need to update the output test file since it now contains a new variable `generating_application@MetaData`.

gthompsnJCSDA (commented):
I am fine with this.  Is it generally true that ncols_conf will be a constant like 4 - since I thought there are typically 4 confidence settings.  And why would there be a second dimension of generating application for ncols_conf?  Are you sure that isn't just a single dimension variable [nloc]?

ilianagenkova:
@gthompsnJCSDA  At the moment ncols_conf is generally 4, true, but that could change if the producers decide so. Therefore it's appropriate to have ncols_conf be a variable. And no, it isn't  a single dimension variable [nloc] . nloc is the number of obs/winds, and for each wind there are X repetitions of GNAPS-PCCF pairs. At the moment X=4. The 4 quality parameters are QI with FC, QI, w.out FC, Common QC w.out FC, Expected Error.

rmclaren:
@aerorahul I added unit tests.

aerorahul (approved):
looks good.

rmclaren:
Can anyone see why the build is failing?

rmclaren (approved):

emilyhcliu:
@rmclaren The built is OK.  Two tests failed.

Screen Shot 2021-03-09 at 3 52 42 PM

emilyhcliu (changes requested):
@ilianagenkova
Your changes look good.  But, need to fix the following two tests related to your changes:

47- test_iodaconv_bufr_satwnd_old_format (Failed)
48 - test_iodaconv_bufr_satwnd_new_format (Failed)

It is complaining that split_by_subset.x not found

rmclaren:
@emilyhcliu That means that the NCEPLIB-bufr utility split_by_subset.x is not available in your shell. I guess this would need to be updated on the automated build machines as well.... 

@aerorahul How should we handle this?

emilyhcliu:
@rmclaren 
The running log I showed in the previous comment is from AWS Build above.  I have access to AWS so I can see the build and run-time message there.

aerorahul:
@rmclaren 
Perhaps adding a test that does not need the `split_by_subset` utility.  The files are already split.

rmclaren:
@aerorahul I went ahead and added YAML files for the two different test cases and added those (in a sense, testing the same way we always do)

Unfortunately this is not a very good test, as it doesn't really exercise the code in question (run_satwnds.py). I was a bit hesitant to add these tests this way as they do not really test any different territory than the tests we already have. Generally you have "unit tests" that test individual code units (classes and functions), functional tests that test scenarios or use cases (basically make sure the software as a whole does what you intended it too), and regression tests that specifically test for bugs that you have fixed in the past in order to make sure they are not re-introduced. From the softwares perspective these YAML files do not meaningfully add to any of these categories... its a kind of beating a dead horse test. In the long run tests are resources that have to be maintained in their own right so I would avoid adding tests that add no new value to the ones that already exist.

rmclaren (approved):

srherbener (approved):
I see that the new tests are running and passing, and these changes look good. Thanks for contributing this!

PR #473, opened 2021-03-18
Added yamls for AIRS, AMSUA, ATMS, MHS, SSMIS sensors following the new IODA naming convention

cyberbass:
## Description

Added yamls for bufr2ioda converter following the new IODA naming convention for AIRS, AMSUA, ATMS, HIRS, MHS, SSMIS sensors. These yaml files are being prepared for the time when the new variable naming conventions are adopted across IODA/UFO.  Other satellites/platforms will be added as further additions are made to NCEP's BUFR-libs at a later time.

### Issue(s) addressed

- closes #899

## Dependencies

None

## Impact

None

## Test Data
Emily provided full global files for us to test on s4:/data/users/eliu/bufr_samples
but we need to obtain some smaller versions of starting BUFR files for ATMS, AMSUA, AIRS, and SSMIS.  The existing files for MHS and HIRS are small enough and already part of the existing data files.


cyberbass:
For AIRS, only the first 10 channels are converted due to the memory issue as specified in 
[#474]. The "sensorCentralFrequency@MetaData" global attribute is not specified for this sensor as well.

test/testinput/bufr_airs_new_format.yaml
```diff
@@ -0,0 +1,149 @@
...
+            datetime:
+              year: YEAR
+              month: MNTH
+              day: DAYS
+              hour: HOUR
+              minute: MINU
+              second: SECO
+          height: 
+            mnemonic: HMSL
+          hols:
```
rmclaren, line 29: Please avoid using abbreviations as keys, I would use names that are easily understandable. I would use the same name you use when you define the ioda object (heightOfSurface in this case). Part of the intent here is to leave the BUFR conventions and names behind (they are isolated in the BUFR section).

rmclaren (changes requested):
Please avoid using abbreviations as keys, I would use names that are easily understandable. I would use the same name you use when you define the ioda object. Part of the intent here is to leave the BUFR conventions and names behind (they are isolated in the BUFR section).

test/testinput/bufr_airs_new_format.yaml
cyberbass: Agreed. I just changed all keys to avoid using of abbreviations.

rmclaren (approved):

gthompsnJCSDA (changes requested):
Nan,
I know there are only a handful more hours for you, but here is what I see in the PR that I can suggest improvements.  I just number a list below, so try to get as many as you can please.

1. Suggest changing heightOfLandSurface to "heightOfSurface" (because it does not have to be a land data point as it could be ocean, lake, etc.
2. Suggest using "earthSurfaceType" in place of landSeaQualifier.
3. Suggest you get rid of "stationElevation" in any yaml for radiance, because there are no "stations" associated with satellites.
4. Suggest you change any instance of translating TMBR mnemonic into a variable called "radiance" since it is just a "brightnessTemperature" later on.  In the IODA part of yaml, you just assign radiance to brightnessTemperature, but in the exports mnemonics section, you map TMBR to radiance.  That is the location you could change to become brightness temperature.  Unless of course if the incoming sat data is truly a radiance as in Watts per meter squared per steradian.
5. I don't understand why there is a letter "u" in the yaml for SSMIS satellite - could you rename that file to skip that letter?
6. I think you may need to adjust the ranges of anything called azimuth or zenith angle.  I would assume azimuth would always be done [0,360] but zenith angle is probably [-180,180]; however I do not know for certain.

cyberbass:
> Nan,
> I know there are only a handful more hours for you, but here is what I see in the PR that I can suggest improvements. I just number a list below, so try to get as many as you can please.
> 
> 1. Suggest changing heightOfLandSurface to "heightOfSurface" (because it does not have to be a land data point as it could be ocean, lake, etc.
> 2. Suggest using "earthSurfaceType" in place of landSeaQualifier.
> 3. Suggest you get rid of "stationElevation" in any yaml for radiance, because there are no "stations" associated with satellites.
> 4. Suggest you change any instance of translating TMBR mnemonic into a variable called "radiance" since it is just a "brightnessTemperature" later on.  In the IODA part of yaml, you just assign radiance to brightnessTemperature, but in the exports mnemonics section, you map TMBR to radiance.  That is the location you could change to become brightness temperature.  Unless of course if the incoming sat data is truly a radiance as in Watts per meter squared per steradian.
> 5. I don't understand why there is a letter "u" in the yaml for SSMIS satellite - could you rename that file to skip that letter?
> 6. I think you may need to adjust the ranges of anything called azimuth or zenith angle.  I would assume azimuth would always be done [0,360] but zenith angle is probably [-180,180]; however I do not know for certain.

Thanks Greg! I'll implement these changes during this afternoon.

cyberbass:
> > Nan,
> > I know there are only a handful more hours for you, but here is what I see in the PR that I can suggest improvements. I just number a list below, so try to get as many as you can please.
> > 
> > 1. Suggest changing heightOfLandSurface to "heightOfSurface" (because it does not have to be a land data point as it could be ocean, lake, etc.
> > 2. Suggest using "earthSurfaceType" in place of landSeaQualifier.
> > 3. Suggest you get rid of "stationElevation" in any yaml for radiance, because there are no "stations" associated with satellites.
> > 4. Suggest you change any instance of translating TMBR mnemonic into a variable called "radiance" since it is just a "brightnessTemperature" later on.  In the IODA part of yaml, you just assign radiance to brightnessTemperature, but in the exports mnemonics section, you map TMBR to radiance.  That is the location you could change to become brightness temperature.  Unless of course if the incoming sat data is truly a radiance as in Watts per meter squared per steradian.
> > 5. I don't understand why there is a letter "u" in the yaml for SSMIS satellite - could you rename that file to skip that letter?
> > 6. I think you may need to adjust the ranges of anything called azimuth or zenith angle.  I would assume azimuth would always be done [0,360] but zenith angle is probably [-180,180]; however I do not know for certain.
> 
> Thanks Greg! I'll implement these changes during this afternoon.

Actually I might not implement point 1 and 2 of your comment since their meanings are read from the bufr directly. In my opinion it might be good to keep the original bufr definition in the translation stage so people know, for example, the "heightOfSurface" is written from the "heightOfLandSurface" in the bufr data.

cyberbass:
> Nan,
> I know there are only a handful more hours for you, but here is what I see in the PR that I can suggest improvements. I just number a list below, so try to get as many as you can please.
> 
> 1. Suggest changing heightOfLandSurface to "heightOfSurface" (because it does not have to be a land data point as it could be ocean, lake, etc.
> 2. Suggest using "earthSurfaceType" in place of landSeaQualifier.
> 3. Suggest you get rid of "stationElevation" in any yaml for radiance, because there are no "stations" associated with satellites.
> 4. Suggest you change any instance of translating TMBR mnemonic into a variable called "radiance" since it is just a "brightnessTemperature" later on.  In the IODA part of yaml, you just assign radiance to brightnessTemperature, but in the exports mnemonics section, you map TMBR to radiance.  That is the location you could change to become brightness temperature.  Unless of course if the incoming sat data is truly a radiance as in Watts per meter squared per steradian.
> 5. I don't understand why there is a letter "u" in the yaml for SSMIS satellite - could you rename that file to skip that letter?
> 6. I think you may need to adjust the ranges of anything called azimuth or zenith angle.  I would assume azimuth would always be done [0,360] but zenith angle is probably [-180,180]; however I do not know for certain.

For "stationElevation" it is actually reading out the sensor elevation above the Earth surface and I suggest keeping it but maybe rename it as sensorElevation.

cyberbass:
> Nan,
> I know there are only a handful more hours for you, but here is what I see in the PR that I can suggest improvements. I just number a list below, so try to get as many as you can please.
> 
> 1. Suggest changing heightOfLandSurface to "heightOfSurface" (because it does not have to be a land data point as it could be ocean, lake, etc.
> 2. Suggest using "earthSurfaceType" in place of landSeaQualifier.
> 3. Suggest you get rid of "stationElevation" in any yaml for radiance, because there are no "stations" associated with satellites.
> 4. Suggest you change any instance of translating TMBR mnemonic into a variable called "radiance" since it is just a "brightnessTemperature" later on.  In the IODA part of yaml, you just assign radiance to brightnessTemperature, but in the exports mnemonics section, you map TMBR to radiance.  That is the location you could change to become brightness temperature.  Unless of course if the incoming sat data is truly a radiance as in Watts per meter squared per steradian.
> 5. I don't understand why there is a letter "u" in the yaml for SSMIS satellite - could you rename that file to skip that letter?
> 6. I think you may need to adjust the ranges of anything called azimuth or zenith angle.  I would assume azimuth would always be done [0,360] but zenith angle is probably [-180,180]; however I do not know for certain.

For point 6, thank you for checking the range since I totally overlooked it. Actually for solar zenith angle the range is [0, 180] with values larger than 90 degrees representing the Sun is below horizon. And for sensor zenith angle it should assume values between 0 to 90 degrees. For both the solar and sensor azimuth angle the range is [0, 360]. I will update them in the yaml soon.

cyberbass:
> Nan,
> I know there are only a handful more hours for you, but here is what I see in the PR that I can suggest improvements. I just number a list below, so try to get as many as you can please.
> 
> 1. Suggest changing heightOfLandSurface to "heightOfSurface" (because it does not have to be a land data point as it could be ocean, lake, etc.
> 2. Suggest using "earthSurfaceType" in place of landSeaQualifier.
> 3. Suggest you get rid of "stationElevation" in any yaml for radiance, because there are no "stations" associated with satellites.
> 4. Suggest you change any instance of translating TMBR mnemonic into a variable called "radiance" since it is just a "brightnessTemperature" later on.  In the IODA part of yaml, you just assign radiance to brightnessTemperature, but in the exports mnemonics section, you map TMBR to radiance.  That is the location you could change to become brightness temperature.  Unless of course if the incoming sat data is truly a radiance as in Watts per meter squared per steradian.
> 5. I don't understand why there is a letter "u" in the yaml for SSMIS satellite - could you rename that file to skip that letter?
> 6. I think you may need to adjust the ranges of anything called azimuth or zenith angle.  I would assume azimuth would always be done [0,360] but zenith angle is probably [-180,180]; however I do not know for certain.

Just pushed the last changes for consistency.

srherbener (commented):
Thanks for adding these new files. They appear to be stored in the repo for the purpose of converting data on the HPC systems. I have the impression that these are not intended for iodaconv tests.

If this is the case, then perhaps these should be stored in a directory other than test/testinput (which is intended for files needed by ctest). Perhaps we could create a top level directory for holding all the YAML files?

gthompsnJCSDA:
Since Nan Chen has now departed JCSDA, I will finish up this PR.
@srherbener  These yaml files are destined to be used in iodaconv once new variable naming conventions are adopted in various codes.  So I think they can live in testinput directory.  If you feel it is better that we place a bunch of yaml files for the new names in some subdir, and "stage" them until we are ready to use them, then I am happy to work that way also.

srherbener:
> Since Nan Chen has now departed JCSDA, I will finish up this PR.
> @srherbener These yaml files are destined to be used in iodaconv once new variable naming conventions are adopted in various codes. So I think they can live in testinput directory. If you feel it is better that we place a bunch of yaml files for the new names in some subdir, and "stage" them until we are ready to use them, then I am happy to work that way also.

If these are destined to be used for new tests, then I agree that they belong in the test directory.

On the other hand if these are the beginning of a library of YAML files for converting data in operations and experiments, then I think we should create a new directory structure to contain this library. If these files get mixed in with the test data, then it will be confusing for a user to know which ones to use. Also we could set up install instructions in CMake to install this library as a unit separate from the code which I think would be useful.

rmclaren:
I agree @srherbener. In the long run we will want to have a directory for YAML files that are not tests.

rmclaren:
I just also just noticed that the YAML files being added input file is not set to any test data available as part of the git repo... If you want to add them as tests you will need to add associated test data to go along with them (use NCEPLIB-bufr iodabufr-testdata.x to miniaturize the data), add associated unit tests, etc...

If we don't want to treat them as tests we should follow steves advice.

gthompsnJCSDA:
> I agree @srherbener. In the long run we will want to have a directory for YAML files that are not tests.


Steve and Ron, the idea of a subdir for prototype and operational usage files is something we have talked about related to UFO yaml files.  Generally, it seems logical to set up a naming convention for the files to reference per application so GDAS would be one suggestion in the names of yaml files for that system, but another for regional apps, and others for other centers.

I dont currently know how to reduce the filesizes.  The files from Emily are massive for the AIRS data for example.  These clearly should be reduced if put into a repo.

rmclaren:
NCEPLIB-bufr iodabufr-testdata.x to reduce file sizes.

gthompsnJCSDA:
> NCEPLIB-bufr iodabufr-testdata.x to reduce file sizes.

Is this new?  I don't see it in any build/bin directory of mine, either iodaconv or the bufr-lib area.  I haven't done a recent update.

rmclaren:
Oops, your right. Its part of Rahuls bufr-tools **https://github.com/aerorahul/bufr-tools.git** in the branch called **feature/iodabufr-testdata**.

gthompsnJCSDA:
@srherbener  Requesting your guidance on how we proceed, possibly with a new subdirectory, for these and other new yaml files.  Two of the six test data sets already exist (MHS and HIRS), while the remaining four are relatively large and could be reduced with a tool from Rahul.  All six yamls are intended to work with the upcoming variable naming conventions and can be added to ctest list, so it seems wasteful to push them into subdir, then potentially bring them back to testinput subdir.

@rmclaren  I hit re-review button for you, because I more extensively tested Nan's work and added documentation in form of abbreviated BUFR mnemonics tables for all six platforms found in this PR (more to come).  The only remaining item that I think will help is your resolution of issue #491

jamiebresch:
Is it possible to call the format/convention by some version numbers or labels rather than 'new'?
It is new as of now. What if the format/convention changes again in the future?

gthompsnJCSDA:
> Is it possible to call the format/convention by some version numbers or labels rather than 'new'?
> It is new as of now. What if the format/convention changes again in the future?

This is definitely temporary and up to debate on how we work it.  The "new" indicates the usage of the new IODA variable naming convention document (note all the var names in the yaml differ from our current ioda names).  The IODA team will be adopting the new naming schema soon and we will be doing major changes over IODA/UFO to start using the new names.  This is just what Nan Chen started on this effort, and I continued.

srherbener:
My apologies for the slow response.

Since the new yaml files are for new tests, let's leave them in test/testinput.

Let's move forward with the test files you have now. Ie, add those tests in and do the missing tests in a subsequent PR. Please add ZenHub issues for the missing tests so we can track that work.

The new ioda file format is called "ioda version2", so let's change "new" to "ioda_v2".

Thanks for picking up this work @gthompsnJCSDA!

srherbener:
What's the status of this PR? Last I recall, I was asked to hold off until given the word to proceed. Are we still waiting for changes to come in before reviewing? Sorry if I misunderstood. Thanks!

gthompsnJCSDA:
> What's the status of this PR? Last I recall, I was asked to hold off until given the word to proceed. Are we still waiting for changes to come in before reviewing? Sorry if I misunderstood. Thanks!


Steve,
With the changes put into V2 IODA, do you know if these yamls will be fully compatible?  If all I need to do is rename the files to change "new" to "v2" (or similar), I can certainly wrap that up.  Technically, none of these files will work AS-IS since all the new naming conventions are used for internal variable names.  But, in meantime, how can we confirm the converted output is already v2-compliant?

rmclaren:
@gthompsnJCSDA There wont be any YAML changes needed to support IODA v2.

srherbener:
@gthompsnJCSDA at this point the only "check" for converted output is to attempt to read in the file, using the ioda-v2 reader, and perform an H(x). I realize that this is not an efficient way to validate file, and I believe that the OBS teams has it in the AOP to develop a ioda-v2 file checker which can be used to validate files produced by ioda converters before attempting the H(x) test.

I agree with @rmclaren that you should be able to use the YAML files as is (perhaps changing the file names).

There is a change however in the way a group, variable pairing is specified between ioda-v1 and ioda-v2 which is to switch `variable@group` to `group/variable`. There is a mechanism in ioda-v2 that does this conversion under the hood so you don't technically have to worry about that for now. I'm mentioning this so that you are aware of this change and we should probably publish the naming conventions using the `group/variable` form since that is the actual ioda-v2 convention. Plus, it's probably better to switch over now before a lot of YAML files get written.

gthompsnJCSDA:
@srherbener @rmclaren   If I go into the yaml files under the export, variables, section, is it safe to swap something like:
 `latitude@MetaData` to `MetaData/latitude`
or equivalently, something like:
 `brightness_temperature@ObsValue` to `ObsValue/brightness_temperature`
or is the "parsing" of the yaml going to balk with that change?

rmclaren:
@gthompsnJCSDA I believe it should work either way as this string is just passed to the IODA engines framework. I will have to make a small modification to the code however to make it ioda v2 compatible (but this won't have any effect at all on the YAML files themselves).

srherbener (approved):
@gthompsnJCSDA thanks for putting in all the work to get this PR ready. I think it looks good now.

srherbener:
I agree that it makes sense to merge this in sooner rather than later. The YAML can be updated in a subsequent PR when @rmclaren finishes the YAML upgrade work.

@rmclaren, @emilyhcliu are you okay with merging now? Thanks!

srherbener:
I'm going to merge this in now despite not having the other reviewer approval. The reason I'm doing this is because there are no functional code changes in this PR, and it's been sitting around for way too long.

PR #476, opened 2021-03-18
update documentation and tests following ufo#921

shlyaeva:
## Description

https://github.com/JCSDA-internal/ufo/pull/921 added an option in `scan_angle` predictor to use scan positions instead of scan angles. Thus, there is no need in `scan_position` predictor names inside bias coefficients files, for all sensors `scan_angle` predictor names should be used.

Here satbias converter documentation and tests are updated to be consistent with this.


weihanucar (approved):

gthompsnJCSDA (approved):
I agree with this change.  I altered my own copy to do the same thing based on discussions about this topic with Anna and testing some new BC/QC code with Wei for SSMIS.

PR #516, opened 2021-04-20
Bug fix for certain sensors on Orion

CoryMartin-NOAA:
## Description

Certain sensors/platforms would fail on Orion when trying to generate satbias IODA files (atms for example).

Two bugs were found:

- file close was located outside of the if that checks if the file is open
- nobs(jchan) was outside of an if statement, and thus, for all lines in the input file after the sensor of interest, the value of jchan would be greater than the size of nobs, on some machines I guess this isn't a problem, but it is (I think) overwriting unallocated memory, which led to downstream problems in the code on Orion. 
 
These changes should fix these bugs. Thanks to @fabiolrdiniz for tag teaming this debugging effort and testing the fix.

### Issue(s) addressed

I do no think this was associated with an issue but was found as part of JEDI-GDAS data preparation testing.

### DIFF
--- a/src/satbias/GsiSatBiasReader.cpp
+++ b/src/satbias/GsiSatBiasReader.cpp
@@ -61,8 +61,8 @@ void findSensorsChannels(const std::string & filename, std::vector<std::string> 
         nchannels.push_back(1);
       }
     }
+    infile.close();
   }
-  infile.close();
 }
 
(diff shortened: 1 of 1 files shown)


shlyaeva (approved):
Thank you for finding and fixing this!

aerorahul (approved):
looks good.

PR #518, opened 2021-04-22
Add support for GsiHofX from standard GSI output

CoryMartin-NOAA:
## Description

Our old feature/files_for_jedi wrote out 'Forecast_adjusted', etc. explicitly, but in GSI by default, only Obs_Minus_Forecast_adjusted and Observation are written to the netCDF diag files.

This PR will compute GsiHofX from O-(O-F) if F is not already present in the GSI diag file.

It was noticed that the IODA formatted files we prepared for the obs team for JEDI-GDAS did not have GsiHofX in them, this should address that.


emilyhcliu (approved):

fabiolrdiniz (approved):
Thanks, @CoryMartin-NOAA! I didn't know how this was implemented before, but these changes look good.

fcvdb (approved):
Thanks, Cory!

srherbener (approved):

PR #529, opened 2021-04-30
New BUFR tables for each subset of the ADPUPA data 

PraveenKumar-NOAA:
## Description

New six BUFR tables for each subset of the ADPUPA data for bufr2ioda conversion.

People can have access of these tables and find important Mnemonics informations. 

## Dependencies

There is no dependencies. 

## Impact

No impact

## Test Data

No test data

### DIFF
--- a/doc/bufr_table_samples/adpupa.nc002001.bufr.table.txt
+++ b/doc/bufr_table_samples/adpupa.nc002001.bufr.table.txt
@@ -1,0 +1,191 @@
+.------------------------------------------------------------------------------.
+| ------------   USER DEFINITIONS FOR TABLE-A TABLE-B TABLE D   -------------- |
+|------------------------------------------------------------------------------|
+| MNEMONIC | NUMBER | DESCRIPTION                                              |
+|----------|--------|----------------------------------------------------------|
+|          |        |                                                          |
+| NC002001 | A63218 | MSG TYPE 002-001  RAWINSONDE - FIXED LAND                |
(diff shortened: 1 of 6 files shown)


srherbener (approved):
Thanks for adding these tables.

jamiebresch:
Just curious - how are these tables generated?
There are existing NC002 tables as in https://www.nco.ncep.noaa.gov/pmb/codes/nwprod/decoders/decod_shared/fix_v5.7.0/bufrtab.002, why is it necessary to have separate subset tables?

gthompsnJCSDA:
Praveen and Jamie, would you like to have a joint telecon together to discuss all this?  In my bufr2ioda work, I have now compiled quite a large total list of individual tables to document a lot of data forms, starting from WMO BUFR and I think we should just come up with a unified convention for where to find all this stuff.  I am also planning to work all week on this topic for the data types that do not have "delayed sequences" that are not yet supported by Ron McLaren's code since ncep-bufr-libs is problematic for these types of BUFR messages.

PraveenKumar-NOAA:
Jamie: these tables contain very specific mnemonics used for only one obs-type, like rawinsonde, rather than grouping all ADPUPA into a single table. When we run bufr2ioda.x, at some point, we are likely to have one stream of incoming data type (rawinsonde for example, separated from adpupa obs) and using a yaml meant only for rawinsonde data.

emilyhcliu (approved):

PR #530, opened 2021-05-03
Move new dimension declarations to new ioda-v2 syntax

### LINKED ISSUE #505: [Bug] iodaconv doesn't build with IODA branch feature/ioda-v2

ioda-bundle with option to build ioda-converters fails to build when using IODA feature/ioda-v2 branch but builds with develop.

## To Reproduce

What computer are you running on?
Hera

What compilers/modules are you using?
Intel stack

Steps to reproduce the behavior

1. clone
2. edit CMakeLists.txt
3. ecbuild -DBUILD_IODA_CONVERTERS=ON -DBUILD_PYTHON_BINDINGS=ON -DMPIEXEC_EXECUTABLE=`which srun` -DMPIEXEC_NUMPROC_FLAG="-n"  ../

## Expected behavior
It should compile.

## Additional information (optional)
The error (for several files) is below:

```
/scratch1/NCEPDEV/da/Cory.R.Martin/JEDI/dev/ioda-bundle/iodaconv/src/satbias/SatBiasConverter.cpp(47): error: no instance of function template "std::make_shared" matches the argument list
            argument types are: (const char [12], long, long, long)
    newDims.push_back(std::make_shared<ioda::NewDimensionScale<int>>("npredictors",
                      ^
/usr/include/c++/4.8.5/bits/shared_ptr.h(610): note: this candidate was rejected because there is a type mismatch after argument substitution
      make_shared(_Args&&... __args)
      ^

/scratch1/NCEPDEV/da/Cory.R.Martin/JEDI/dev/ioda-bundle/iodaconv/src/bufr/IodaEncoder/IodaEncoder.cpp(104): error: no instance of function template "std::make_shared" matches the argument list
            argument types are: (const std::string, std::size_t, std::size_t, std::size_t)
                  auto newDim = std::make_shared<ioda::NewDimensionScale<int>>(scale.name,
                                ^
/usr/include/c++/4.8.5/bits/shared_ptr.h(610): note: this candidate was rejected because there is a type mismatch after argument substitution
      make_shared(_Args&&... __args)
      ^

compilation aborted for /scratch1/NCEPDEV/da/Cory.R.Martin/JEDI/dev/ioda-bundle/iodaconv/src/bufr/IodaEncoder/IodaEncoder.cpp (code 2)
```

srherbener:
## Description

This PR replaces the syntax for declaring new dimensions in develop (shared_ptr method) with the new syntax in feature/ioda-v2 (no shared_ptr, simpler interface).

### Issue(s) addressed

- fixes #505

The feature/ioda-v2 branch in ioda-converters compiles successfully. Note that you need to use the feature/ioda-v2 branch in ioda for this to work properly.

## Dependencies

None

## Impact

None

## Test Data

None

### DIFF
--- a/src/bufr/IodaEncoder/IodaEncoder.cpp
+++ b/src/bufr/IodaEncoder/IodaEncoder.cpp
@@ -101,10 +101,7 @@ namespace Ingester
                     }
                 }
 
-                auto newDim = std::make_shared<ioda::NewDimensionScale<int>>(scale.name,
-                                                                             size,
-                                                                             size,
-                                                                             size);
+                auto newDim = ioda::NewDimensionScale<int>(scale.name, size);
                 newDims.push_back(newDim);
 
                 if (size <= 0) { foundInvalidDim = true; }
--- a/src/satbias/SatBiasConverter.cpp
+++ b/src/satbias/SatBiasConverter.cpp
@@ -43,11 +43,10 @@ ioda::ObsGroup makeObsBiasObject(ioda::Group &empty_base_object,
   ASSERT(channels == channels_in_errorfile);
 
   // Creating dimensions: npredictors & nchannels
-  ioda::NewDimensionScales_t newDims;
-  newDims.push_back(std::make_shared<ioda::NewDimensionScale<int>>("npredictors",
-                    numPreds, numPreds, numPreds));
-  newDims.push_back(std::make_shared<ioda::NewDimensionScale<int>>("nchannels",
-                    numChans, numChans, numChans));
+  ioda::NewDimensionScales_t newDims {
+      ioda::NewDimensionScale<int>("npredictors", numPreds),
+      ioda::NewDimensionScale<int>("nchannels", numChans)
+  };
 
   // Construct an ObsGroup object, with 2 dimensions npred, nchans
   ioda::ObsGroup ogrp = ioda::ObsGroup::generate(empty_base_object, newDims);


CoryMartin-NOAA (approved):
Thanks Steve for the fix!

srherbener:
@CoryMartin-NOAA have you tested on Hera? Just want to double check before I merge.

I goofed up the version of ioda-bundle which is why the CI testing failed. I'm retrying the tests. Do you expect the `test_iodaconv_tropomi_no2` test to fail at this point. It looks like you were in the middle of switching that script to your new library.

CoryMartin-NOAA:
@srherbener Hera is down today for monthly maintenance so I can't check. Will be able to try it tomorrow. Yes I expect the test to fail because, as you said, I'm using that as the guinea pig and it's in the middle of being transitioned to the new IODA python bindings. I suspect that the python library is not being found as I have to manually add IODA to the PYTHONPATH and LD_LIBRARY_PATH to get it to work. I think we talked about this issue before but not sure if a solution was found.

srherbener:
Okay, it's a bit more involved to get the right branches to be selected than having a branch on ioda-bundle. Since this is working in my local clone and it's not being merged into develop, I'm going to go ahead and merge this in so it's ready for you to try when Hera comes back up.

rmclaren (approved):

CoryMartin-NOAA:
@srherbener builds on Hera now. Thanks!

PR #531, opened 2021-05-04
Addition of avhrr3 naming for GSI ncdiag files

fcvdb:
## Description

This PR adds the conversion of "avhrr3" named GSI ncdiag files into observation and geoval file.  

Provide a detailed description of what this PR does.

The "avhrr3" instrument has been added to the list of radiance sensors for GSI ncdiags input files, so that GSI-generated diag files "avhrr3" can be processed and "avhrr3" observation and geoval files can be created. 

This solves the naming convention problem noticed in the JEDI GDAS experiments.

### Issue(s) addressed

https://github.com/JCSDA-internal/experiments/issues/15

ioda-converters were tested with "avhrr" and "avhrr3" input files and generated the proper output file name and content. 

What does it mean for this PR to be finished?

## Dependencies

[GSI](https://github.com/JCSDA-internal/GSI) ncdiag scripts will have to use "avhrr3" naming from now.

## Impact

Please list the other repositories (if any) that this PR will require changes in (example below)

- [GSI](https://github.com/JCSDA-internal/GSI)
- [experiments](https://github.com/JCSDA-internal/experiments)

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -268,7 +268,6 @@ rad_sensors = [
     'ssmis',
     'abi',
     'ahi',
-    'avhrr',
     'avhrr3',
     'saphir',
     'gmi',
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -269,6 +269,7 @@ rad_sensors = [
     'abi',
     'ahi',
     'avhrr',
+    'avhrr3',
     'saphir',
     'gmi',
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):
Approved. However, this doesn't really fix the root issue which is that GSI outputs 'avhrr' when it really is 'avhrr3'. Where should this happen? Do we want to try to modify it in GSI? Have an intermediate mv/rename script? Or force the output in this script to be 'avhrr3' if named 'avhrr'?

fabiolrdiniz (approved):
Sorry, @fcvdb. I didn't realize I was listed here as a reviewer. Too late. Thank you!

PR #556, opened 2021-05-21
Replaced function in ioda-v2

rhoneyager:
This PR fixes the bufr converter code so that it can build again ioda-v2. A function was rewritten quite a while ago and somehow escaped CI detection.


rhoneyager:
Nevermind. Old branch.

PR #590, opened 2021-07-15
GOES-16 IODAv2 Converter

ericlingerfelt:
## Description

The GOES-16 converter classes generate two IODAv2 data files from a group of raw data files for all 16 channels of GOES-16 LB1 scan. The final result is two IODAv2 formatted data files - one for Reflectance Factor (RF, ABI channels 1-6) and one for Brightness Temperature (BT, ABI channels 7-16). Since GOES-16 is in a geostationary orbit, an auxiliary IODAv2 file containing relevant variables and attributes for latitude, longitude, scan angle, and elevation angle is accessed (or created if it does not exist) through the latlon_file_path input argument. This converter checks to see if the nadir for the satellite has changed and creates a new latlon IODAv2 file if a nadir change has occurred.

Note that GOES-16 ABI channels 1, 3, and 5 are subsampled from 1km to 2km resolution and ABI channel 2 is subsampled from 0.5km to 2km resolution.

Please refer to the header and function documentation for details concerning each class including the groups, variables, dimensions, and attributes generated as well as references to documentation used for all calculations. 

## Usage

goes16_converter = Goes16Converter(input_file_paths, latlon_file_path, output_file_path_rf, output_file_path_bt)
goes16_converter.convert()


legibe (approved):
I can't comment on the veracity of the code since I don't know the file format. I am seeing some nice Python idioms in there which is nice.

Does it mean you would have one (or more) auxiliary files(s) to archive for a particular date time?

rickgrubin:
Data from GOES may come from GOES16 or GOES17 instruments.

Per the document [NOAA GOES on AWS (CICS)](https://docs.opendata.aws/noaa-goes16/cics-readme.html), the formats and naming conventions are the same, save for satellite ID (_"satellite id for GOES-16 (future G17"_).

Thoughts on modifying the code to be more generic and consider both satellite IDs?

ericlingerfelt:
@legibe, We will have an auxiliary for both GOES-16 and GOES-17 for the current latitude and longitude of each instrument. In the near future, these could be a static files that are retrieved from r2d2.

ericlingerfelt:
@rickgrubin, I will create a commit to generalize this to just the "GOES" label and remove the reference to 16.

BenjaminTJohnson:
@ericlingerfelt @rickgrubin 

Just want to add: GOES-16 and GOES-17 Are two different satellites, and the sensors onboard the satellites while technically similar, are not the same, and do not occupy the same position in space.  These should be treated as separate satellites and the ABI for each instrument should be treated as separate sensors.  @gthompsnJCSDA may want to weigh in on this too.
I'm okay with genericity here where it makes sense to do so, but we need to be sure that these two datasets are able to be independently utilized. 

Also, given that UFO/OOPS/IODA does not possess the functionality to pipe through `reflectance` for the visible channels, I recommend that we focus on H(x) for IR channels (7+) only for proof of concept.  We can revisit the visible channels once the appropriate code changes are made (months from now).

ericlingerfelt:
@BenjaminTJohnson, We will certainly treat the input, lat/lon, and output data as separate satellites and data products. Since GOES-17 and GOES-16 have the same channel bands, only the metadata harvesting will need to be updated. I am creating a new commit to generalize the current programming. The metadata for each GOES input file is parsed in goes.py (soon to be renamed from goes16.py).

ericlingerfelt:
I have updated the programs to use both GOES-16 and GOES-17 data. The platform_identifier (either G16 or G17) is now parsed from the file name and is accessible through the goes.py get_platform_identifier function.

ytremolet (approved):
Looks good overall. I'm not a specialist for I'll leave it to the OBS experts for the details.

src/goes/test_goes_converter.py
```diff
@@ -0,0 +1,33 @@
...
+#
+import os
+import time
+
+from solo.basic_files import tree
+from goes_converter import GoesConverter
+
+
+def test_goes_converter():
+    input_files_path = "/Users/eric2/PROJECTS/GOES16_TEST_FILES/input/12_15_2020/00"
```
srherbener, line 15: I suspect that the files you are using for this test are large and inappropriate for unit testing (runtime is too long, ie longer than a few seconds). If that's the case, then I suggest making these variables, with the hardcoded paths, arguments to this script so that the hardcoded paths can be eliminated.

If these paths are pointing to small data that runs quickly, then please turn this into a "ctest" style test. You can look in one of the other source files to see how that is done. I can help with this too.

Thanks!

srherbener (commented):
I think the code looks good, and I think this is a great first version of the converter. Please address the comment attached to `src/goes/test_goes_converter.py` for this PR.

For the longer term (after this PR gets merged) please consider:
* We have a ioda Python API in IODA v2 and would like all of the python converters to use that API.
* The OBS team is getting close to publishing the first IODA naming convention documentation. Eventually we would like all of the converters to use YAML for configuring the names (following the IODA convention) that appear in the output file.

Thanks!

src/goes/test_goes_converter.py
ericlingerfelt: Yes. The process to convert 16 channels into the two IODAv2 files takes around 3 minutes. This Python script is really just a driver at this point. I will modify this script to be a proper "ctest" style test following guidance from you and the other source files. Thank you for this input.

ericlingerfelt:
I will definitely utilize the IODA Python API and the upcoming IODAv2 YAML standard in the new version. Thanks, @srherbener!

src/goes/test_goes_converter.py
srherbener: 3 minutes is a bit too long for a unit test. If it's simple and easy to trim the input files down so that the test runs in a couple seconds, then it may be worthwhile to do that for this PR. If it's going to take some work to create a small, fast test then I think it might be better to get rid of the hard coded paths (ie, add arguments to this script) and call it good for this PR. It would be okay to add the unit test in a subsequent PR.

ericlingerfelt: I will work on configuring a quick test in the next PR. I pushed a commit to remove the hard-coded paths for this test driver script.

ericlingerfelt (commented):
I have modified the test driver script as requested.

src/goes/test_goes_converter.py
```diff
@@ -0,0 +1,28 @@
...
+    output_file_path_rf - The path to write the IODAv2 reflectance factor data file
+    output_file_path_bt - The path to write the IODAv2 brightness temperature data file
+    """
+    goes_converter = GoesConverter(input_file_paths, latlon_file_path, output_file_path_rf, output_file_path_bt)
+    goes_converter.convert()
+
+
+if __name__ == '__main__':
+    start_time = time.time()
+    test_goes_converter()
```
srherbener, line 26: Do you need to add the arguments to the test_goes_converter() call here? These could be taken from sys.argv (import sys) so that you can specify them on your shell command line.

ericlingerfelt: Of course. That was an oversight. Thanks for pointing that out!

ericlingerfelt (commented):
I have made a commit to resolve this oversight.

ericlingerfelt (commented):
Thanks. @srherbener!

srherbener:
My apologies @ericlingerfelt, I left out a step in my instructions to add the `src/goes/CMakeLists.txt` file. I've added that which has enabled the coding style check on the goes converter scripts. In my local clone the coding style check failed with a few errors that will hopefully be quick to fix. We should see the same result in the CI testing when it finishes. I can help with getting these coding style errors fixed.

srherbener:
Here's the report from the Clang testing:
```
test 3
--
951 | Start 3: iodaconv_goes_coding_norms
952 |  
953 | 3: Test command: /build_container/bin/iodaconv_lint.sh "/jcsda/ioda-bundle/iodaconv/src/goes" "/jcsda/ioda-bundle/iodaconv"
954 | 3: Environment variables:
955 | 3:  OMP_NUM_THREADS=1
956 | 3: Test timeout computed to be: 10000000
957 | 3: + SRC_DIR=/jcsda/ioda-bundle/iodaconv/src/goes
958 | 3: + PYCODESTYLE_CFG_PATH=/jcsda/ioda-bundle/iodaconv
959 | 3: + cd /jcsda/ioda-bundle/iodaconv/src/goes
960 | 3: + pycodestyle -v --config=/jcsda/ioda-bundle/iodaconv/.pycodestyle --filename=*.py,*.py.in .
961 | 3: cli configuration: /jcsda/ioda-bundle/iodaconv/.pycodestyle
962 | 3: directory .
963 | 3: checking ./goes.py
964 | 3: ./goes.py:134:12: E741 ambiguous variable name 'l'
965 | 3: ./goes.py:139:17: E741 ambiguous variable name 'l'
966 | 3: ./goes.py:140:20: E741 ambiguous variable name 'l'
967 | 3: ./goes.py:141:21: E741 ambiguous variable name 'l'
968 | 3: ./goes.py:158:12: E741 ambiguous variable name 'l'
969 | 3: ./goes.py:174:17: E741 ambiguous variable name 'l'
970 | 3: ./goes.py:175:20: E741 ambiguous variable name 'l'
971 | 3: ./goes.py:176:21: E741 ambiguous variable name 'l'
972 | 3: ./goes.py:193:12: E741 ambiguous variable name 'l'
973 | 3: ./goes.py:211:17: E741 ambiguous variable name 'l'
974 | 3: ./goes.py:212:20: E741 ambiguous variable name 'l'
975 | 3: ./goes.py:213:21: E741 ambiguous variable name 'l'
976 | 3: ./goes.py:289:21: E127 continuation line over-indented for visual indent
977 | 3: checking ./goes_converter.py
978 | 3: checking ./goes_latlon.py
979 | 3: checking ./test_goes_converter.py
980 | 3: 1       E127 continuation line over-indented for visual indent
981 | 3: 12      E741 ambiguous variable name 'l'
982 | 1/1 Test #3: iodaconv_goes_coding_norms .......***Failed    0.16 sec
983 |  
984 | 0% tests passed, 1 tests failed out of 1
985 |  
986 | Label Time Summary:
987 | iodaconv    =   0.16 sec*proc (1 test)
988 | script      =   0.16 sec*proc (1 test)
989 |  
990 | Total Test time (real) =   0.16 sec
991 |  
992 | The following tests FAILED:
993 | 3 - iodaconv_goes_coding_norms (Failed)
994 | Errors while running CTest
```

I think the ambiguous variable name (for the variable 'l') is that the checker is expecting i, j, k to be used as index variables but not so for l. Then it says that 'l' is ambiguous because its name is not mnemonic. I think either changing the 'l' to a longer name, or perhaps 'n' would work to fix this.

If you run into trouble trying to fix these coding style errors, I can take a shot at it.

Thanks for your patience with this!

ericlingerfelt:
Hi @srherbener, no worries! Thanks for catching these formatting errors. I have committed the fix. Changing index 'l' to 'n' worked using PEP8 validator in PyCharm.

ericlingerfelt (commented):
All comments have been resolved. Please merge.

srherbener:
@ericlingerfelt thanks for fixing the coding style errors. I will merge now.

Issue #587, opened 2021-07-13
Add suffix to AMVs filename in bufr2ioda converter => add BUFR version instead


ilianagenkova:
The bufr2ioda AMVs converter uses CLATH mnemonic to recognize if it's reading from **old** or **new** BUFR format/sequence. We need to add a suffix to the output file name to preserve the information about the original BUFR file (old or new table).

**Old** (a.k.a **heritage** BUFR format) refers to using  BUFR sequence 3-10-014 in Table D. 
**New** (introduced with the start of GOES-R AMV products) refers to using  BUFR sequence 3-10-077 in Table D.   

Note: sequence 3-10-067 was used as intermediate while sequence 3-10-077 was finalized. 

Going forward, we will be using suffixes "014" (for old) and "077" (for new) in the filenames produced by the ioda converters.

**Winds using sequence 3-10-014**
GOES-13 and GOES-15 AMVs (and prior GOES)
AVHRR AMVs from NOAA-15..19 instruments
MODIS AMVs
VIIRS AMVs from S-NPP before 5 Dec 2018 (DATE to be verified)
EUMETSAT AMVs (from any instrument) before 2 Feb 2021  (DATE to be verified)
MTSAT and Himawari AMVs 

**Winds using sequence 3-10-077**
GOES-16 and GOES-17 AMVs (and future GOES)
VIIRS from S-NPP after 5 Dec 2018 (DATE to be verified)
VIIRS from NOAA-20
EUMETSAT AMVs  (from any instrument) after 2 Feb 2021  (DATE to be verified)

run_satwind.py code is updated to accommodate the needed suffixes.

ilianagenkova:
A very brief intro to 3-10-014 and 3-10-077 :
https://www.nco.ncep.noaa.gov/sib/jeff/TableD_0_STDv31_LOC7.html (i.e. search for  3-10-014)

For the meaning of a sequence number, i.e. x-xx-xxx, see:
https://rda.ucar.edu/docs/formats/bufr/WMO306_vI2_BUFR_TableD_en.pdf

Brief doc on BUFR:
https://www.eumetnet.eu/wp-content/uploads/2017/04/OPERA_BUFR_template_primer_V1.4.pdf
or 
http://dss.ucar.edu/docs/formats/bufr/bufr.pdf
Detailed BUFR doc from WMO:
http://meteo.ieec.uned.es:8086/PFC_JMEstepa/documentos/BUFR/layer3.pdf

[...]

PR #596, opened 2021-07-22
Add bufr_version as global attribute to bufr2ioda output (for AMVs only)

ilianagenkova:
## Description
See Issue [#587](https://github.com/JCSDA-internal/ioda-converters/issues/587)

We are adding global attribute _bufr_verson_ (BUFR sequence version) to the ioda output from bufr2ioda, allowing us to distinguish between two BUFR versions used for AMVs (satellite winds).
Later, UFO filters could read  bufr_verson and make quality control decisions.

This PR adds code to:
ioda-bundle/iodaconv/tools/run_satwnds/satwnds_old_subset_template.yaml
ioda-bundle/iodaconv/tools/run_satwnds/satwnds_new_subset_template.yaml
to add the global attribute.

- fixes #<587>

The ioda output will show bufr_version:

// global attributes:
		string :_ioda_layout = “ObsGroup” ;
		:_ioda_layout_version = 0 ;
		string :bufr_version@MetaData = “310077" ;  
 
(or "310014")

## Dependencies
None

## Impact
- [iodaconv] iodaconv in ioda-bundle

### DIFF
--- a/tools/run_satwnds/satwnds_new_subset_template.yaml
+++ b/tools/run_satwnds/satwnds_new_subset_template.yaml
@@ -73,9 +73,9 @@ observations:
           size: variables/percent_confidence.ncols
 
       globals:
-        - name: "bufr_version@MetaData"
+        - name: "bufr_seq_version"
           type: string
-          value: "310077"    
+          value: "3-10-077"    
 
       variables:
         - name: "satellite@MetaData"
--- a/tools/run_satwnds/satwnds_new_subset_template.yaml
+++ b/tools/run_satwnds/satwnds_new_subset_template.yaml
@@ -72,6 +72,11 @@ observations:
         - name: "ncols_conf"
           size: variables/percent_confidence.ncols
 
+      globals:
+        - name: "bufr_version@MetaData"
+          type: string
+          value: "310077"    
+
       variables:
         - name: "satellite@MetaData"
           source: variables/satellite
--- a/tools/run_satwnds/satwnds_old_subset_template.yaml
+++ b/tools/run_satwnds/satwnds_old_subset_template.yaml
@@ -68,9 +68,9 @@ observations:
           size: variables/percent_confidence.ncols
 
       globals:
-        - name: "bufr_version@MetaData"
+        - name: "bufr_seq_version"
           type: string
-          value: "310014"     
+          value: "3-10-014"     
 
       variables:
         - name: "satellite@MetaData"
--- a/tools/run_satwnds/satwnds_old_subset_template.yaml
+++ b/tools/run_satwnds/satwnds_old_subset_template.yaml
@@ -67,6 +67,11 @@ observations:
         - name: "ncols_conf"
           size: variables/percent_confidence.ncols
 
+      globals:
+        - name: "bufr_version@MetaData"
+          type: string
+          value: "310014"     
+
       variables:
         - name: "satellite@MetaData"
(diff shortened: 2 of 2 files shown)


tools/run_satwnds/satwnds_new_subset_template.yaml
```diff
@@ -72,6 +72,11 @@ observations:
         - name: "ncols_conf"
           size: variables/percent_confidence.ncols
 
+      globals:
+        - name: "bufr_version@MetaData"
+          type: string
+          value: "310077"
```
rmclaren, line 78: It might make sense to make the name something like seq_version and and the value equivalent to the mnemonic for that sequence (GEOSTWND or whatever), up to you :)

tools/run_satwnds/satwnds_old_subset_template.yaml
```diff
@@ -67,6 +67,11 @@ observations:
         - name: "ncols_conf"
           size: variables/percent_confidence.ncols
 
+      globals:
+        - name: "bufr_version@MetaData"
+          type: string
+          value: "310014"
```
rmclaren, line 73: Same here.

tools/run_satwnds/satwnds_new_subset_template.yaml
ilianagenkova: bufr_version, seq_version, or even bufr_seq_version - I'm fine with either choice.

I used "310077"  b/c it comes straight from WMO documents. In fact, it should be "3-10-077". GEOSTWND is more familiar to NCEP folks, I agree, but I don't think MetOffice/NRL /etc are familiar with our naming in Table D.

Let's get @emilyhcliu to chime in, she is leading the Obs Re-engineering effort and may have an opinion on this too.

tools/run_satwnds/satwnds_new_subset_template.yaml
```diff
@@ -72,6 +72,11 @@ observations:
         - name: "ncols_conf"
           size: variables/percent_confidence.ncols
 
+      globals:
+        - name: "bufr_version@MetaData"
```
srherbener, line 76: There is no need to use the `@MetaData` suffix on an attribute name. The `@MetaData` syntax is for variable names, and the purpose of the suffix is to denote which group the variable belongs to. You might want to consider dropping the `@MetaData` suffix so it's a little easier to read the attribute name.

srherbener (approved):
I think the changes look good.

I put in a comment to consider, but I don't see that as a blocking issue for merging this PR into develop.

tools/run_satwnds/satwnds_new_subset_template.yaml
ilianagenkova: @srherbener,  I updated the code to remove @MetaData and renamed bufr_version with bufr_seq_version to accommodate @rmclaren's suggestion for clarity.
Also, I updated the string values, so they look exactly like in most WMO documents, thus easier to Google search for them. 

No more changes are anticipated for a while so merging would be fine. thanks!

srherbener: Thanks for making this change!

srherbener: Thanks for letting me know that you anticipate no more changes. I still need one more approval before merging.

rmclaren (approved):

PR #602, opened 2021-08-06
Feature/aeronet2ioda aod

bhuang95:
## Description
        aeronet_aod2ioda.py was added in ioda-converters/src/chem/ to obtain and format online multi-wavelength AERONET AODs in IODA v1. A tension spline function was included to interpolate AOD at wavelengths (e.g., 550 nm) that are not available in AERONET. Please see details in the header of this code. 
       Its hofx calculation is introduced in #1425.
       This branch passed all ctests on Hera except for "30 - test_iodaconv_ncep_bufr" due to missing some python library on Hera.   

## Impact


andytangborn:
The ctests all worked fine for me. I'm curious why you chose to use the url for the aeronet input data file instead of having a small subset local file for the ctest. I would think that some machines might be able to reach the aeronet url.  Otherwise it looks good.

bhuang95:
@andytangborn Thanks for your tests, Andy. We could have done as you mentioned to read a subset local file. But we thought it would be beneficial for NRT or RT run by adding this online reading capability. The potential problem of aeronet url access may be resolved by the admin, if permitted. We need to further communicate with Steve and Ryan to decide which to choose for the implementation.

andytangborn (approved):
Bo,  I'm fine with the code as it is, but I think you should find out whether there might be a problem down the road if it gets used by an operational system, and the aeronet files are temporally unavailable.

Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
CoryMartin-NOAA, line 1: Please remove this file

Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Aug 04 21:26 UTC
```
CoryMartin-NOAA, line 1: Remove this file

src/chem/aeronet_aod2ioda.py
```diff
@@ -0,0 +1,381 @@
...
+        if inv_type is not None:
+            self.inv_type = 'ALM15'
+        else:
+            self.inv_type = inv_type
+        if 'AOD' in self.prod:
+            self.new_aod_values = interp_to_aod_values
+        self.build_url()
+        try:
+            self.read_aeronet()
+        except Exception:
```
CoryMartin-NOAA, line 217: if you can't read, just print the URL? Seems like it needs something else here.

src/chem/aeronet_aod2ioda.py
```diff
@@ -0,0 +1,381 @@
...
+                freq).mean().reset_index()
+        if self.new_aod_values is not None:
+            self.calc_new_aod_values()
+        return self.df
+
+    def calc_new_aod_values(self):
+
+        def _tspack_aod_interp(row, new_wv=[440., 470., 550., 670., 870., 1240.]):
+            try:
+                import pytspack
```
CoryMartin-NOAA, line 230: This looks like something Barry created? Is there a more standard way to do this that doesn't require yet another thing installed?

src/lib-python/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ list( APPEND libs
   collect_sources.py
   meteo_utils.py
   orddicts.py
+  pytspack.py 
+  tspack.cpython-37m-x86_64-linux-gnu.so
```
CoryMartin-NOAA, line 8: This is bad, no? this means it will ONLY work for python 3.7 on X86 64bit linux

CoryMartin-NOAA (changes requested):
First pass. Some comments/questions. I agree with @andytangborn I would prefer separate download and convert scripts. These are 'converters' not 'collectors'. However, if it works on Hera, then it is probably fine.

I am much more concerned about the compiled Python .so file being added/used here.

Testing/Temporary/CTestCostData.txt
bhuang95: Removed. Thanks.

Testing/Temporary/LastTest.log
bhuang95: This file has been removed. Thanks.

src/chem/aeronet_aod2ioda.py
bhuang95: I have remove codes of online reading.

src/chem/aeronet_aod2ioda.py
bhuang95: I am not sure how to apply tspack without installed library tspack.cpython-37m-x86_64-linux-gnu.so. So I have remove codes for AOD interpolation. Thanks.

src/lib-python/CMakeLists.txt
bhuang95: I agree. It will causes incompatibility issue on different machine while running this python code. So I removed codes for AOD interpolations. Thanks.

bhuang95:
@CoryMartin-NOAA @andytangborn Since there are concerns about potential online AERONET AOD data access and TSPACK library incompatibility issue. I have modified codes to remove the online reading and AOD interpolation features. The current code readS AERONET AOD ASCII file downloaded from NASA website and write it to IODA format. Please let me know if you have further comments. Thanks.

CoryMartin-NOAA (approved):

srherbener (approved):
Looks great. Thanks for contributing this converter!

PR #632, opened 2021-09-29
Feature/hbo_aeronet_aaod

bhuang95:
This PR creates an IODA converter for AEORNET absorption AOD (AAOD) at Level 1.5 or 2.0. Produced IODA files would serve assimilation and hofx calculation of AAOD (UFO #1540, #1541). But its completion is independent of these two PRs. 

    (1) Below files are included. 
        (1.a) src/chem/aeronet_aaod.py
        (1.b) test/testinput/aeronet_cad.dat, aeronet_ssa.dat and aeronet_tab.dat
        (1.c) test/testoutput/aeronet_aaod.nc

    (2) Three AERONET inversion products were read from their corresponding ASCII files and written into IODA format, that is, conicident AOT data with almucantar (CAD), single scattering albedo (SSA) and AOD aborption (TAB) (see details at https://aeronet.gsfc.nasa.gov/print_web_data_help_v3_inv_new.html). CAD is also written out,  since it serves as a reference for AAOD and will be used in 3.a and 3.b below.

    (3) Three MetaData variables are added as below. Because inversion data at Level 2.0 is very sparse, these MetaData variables would add more flexibility of using/filtering inversion data at Level 1.5 through "Domain check filter". 
        (3.a) l20_without_aod440_0.4_threshold_qc: whether AOD data reach Level 2.0 (0: yes, 1: no) without the threshold of aod_440nm > 0.4 (aod_440nm is obtained from CAD variable)
        (3.b) l20_qc: whether AOD data data reach Level 2.0 (0: yes, 1: no) with the threshold of aod_440nm > 0.4
        (3.3) inversion_l20_qc: inversion data quality level (0 for lev20 and 1 for lev15)


src/chem/aeronet_aaod2ioda.py
```diff
@@ -0,0 +1,221 @@
...
+    parser = argparse.ArgumentParser(
+        description=(
+                     'Reads AERONET inversion files downloaded from NASA website '
+                     ' and converts into IODA format')
+    )
+
+    required = parser.add_argument_group(title='required arguments')
+    required.add_argument(
+                    '-i', '--input',
+                    help="path of AERONET inversion ASCII files that include "
```
CoryMartin-NOAA, line 93: Make this explicitly say 'path to directory containing...' or something like that as this reads to me like you should provide it 3 paths `/path/to/aeronet_cad.dat /path/to/aeronet_ssa.dat` etc.

src/chem/aeronet_aaod2ioda.py
```diff
@@ -0,0 +1,221 @@
...
+                    help="path and filename of AERONET inversion IODA file",
+                    type=str, required=True)
+
+    args = parser.parse_args()
+    indir = args.input
+    outfile = args.output
+
+    # Read and extract online AERONET inversion data
+    print('Read and extract AERONET inversion data: CAD, SSA, TAB')
+    f3_cad_all = add_data('%s/aeronet_cad.dat' % (indir))
```
CoryMartin-NOAA, line 107: consider using "f-strings"

CoryMartin-NOAA: and `os.path.join`

src/chem/aeronet_aaod2ioda.py
```diff
@@ -0,0 +1,221 @@
...
+        exit(0)
+
+    locationKeyList = [("latitude", "float"), ("longitude", "float"), ("datetime", "string"),]
+    writer = iconv.NcWriter(outfile, locationKeyList)
+    varDict = defaultdict(lambda: defaultdict(dict))
+    outdata = defaultdict(lambda: DefaultOrderedDict(OrderedDict))
+    loc_mdata = defaultdict(lambda: DefaultOrderedDict(OrderedDict))
+    var_mdata = defaultdict(lambda: DefaultOrderedDict(OrderedDict))
+    units = {}
+    units['latitude'] = 'degree'
```
CoryMartin-NOAA, line 143: is the IODA convention `degree` or `degrees`? (can't recall)

src/chem/aeronet_aaod2ioda.py
```diff
@@ -0,0 +1,221 @@
...
+    # Define varDict variables
+    print('Define varDict variables')
+    for key, value in obsvars.items():
+        print(key, value)
+        varDict[key]['valKey'] = key, writer.OvalName()
+        varDict[key]['errKey'] = key, writer.OerrName()
+        varDict[key]['qcKey'] = key, writer.OqcName()
+
+    # Define loc_mdata
+    print('Define loc_mdata')
```
CoryMartin-NOAA, line 171: do we need all of these debug print statements?

CoryMartin-NOAA (commented):
A few comments, but overall looks good

src/chem/aeronet_aaod2ioda.py
bhuang95: @CoryMartin-NOAA Thanks for reviewing the codes. I modified and cleaned up the codes as you suggested except that units['latitude'] = 'degree' is retained.  'degree' is the unit for lon/lat in the IODA name convention document. Besides, because single scattering albedo (SSA) is not directly used in the AAOD assimilation and calculation,  I removed its related variables in the code, files in test/testinput and test/CMakeLists.txt.

src/chem/aeronet_aaod2ioda.py
```diff
@@ -0,0 +1,207 @@
...
+        type=str, required=True)
+
+    args = parser.parse_args()
+    indir = args.input
+    outfile = args.output
+
+    # Read and extract online AERONET inversion data
+    print('Read and extract AERONET inversion data: CAD, TAB')
+    f3_cad_all = add_data(os.path.join(f"{indir}/", "aeronet_cad.dat"))
+    f3_tab_all = add_data(os.path.join(f"{indir}/", "aeronet_tab.dat"))
```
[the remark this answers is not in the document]
bhuang95: @srherbener Thanks for reviewing. These two files need to be downloaded separately from NASA AERONET website. The downloaded files have the same names. So I define two arguments in the updated code as you suggested, that is,  "-i" for the CAD file and "-j" for the TAB file. Please let me know if further changes are expected.

CoryMartin-NOAA (approved):
I think it looks good to me now, @bhuang95

src/chem/aeronet_aaod2ioda.py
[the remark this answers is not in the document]
srherbener: @bhuang95 thank you for addressing this! I like the long option names you picked (`--incad`, `--intab`) which are mnemonic (remind you of what they represent). The short option names (`-i`, `-j`) however don't seem as mnemonic. I realize that one letter makes it very limited when trying to pick mnemonic names, but perhaps `-c` for the CAD file and `-t` for the TAB file would be useful. What do you think?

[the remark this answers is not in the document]
bhuang95: @srherbener Thanks, Steve. I think the short option names you picked looks better. I just updated them in the code.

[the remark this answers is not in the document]
srherbener: Sounds great! I'll wait for the CI testing to complete, and if all looks good will merge this into develop.

srherbener (approved):

src/chem/aeronet_aaod2ioda.py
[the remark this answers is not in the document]
bhuang95: Great, thank you very much!

PR #634, opened 2021-09-29
Add GsiEffectiveQC variable

### LINKED ISSUE #633: Create GSI Effective QC variable
## Description
The Obs team has suggested that it will be easier to evaluate UFO vs GSI QC if we have a binary GSI Effective QC variable that is 0 if pass and 1 if fail, based off of the value of Inverse_Observation_Error from the GSI diagnostic files.

## Requirements

update to GSI ncdiag converter
update to test reference files

PR to add this change

## Dependencies
None

CoryMartin-NOAA:
## Description

Adds a new variable, GsiEffectiveQC, to the GSI ncdiag output.

### Issue(s) addressed

- fixes #633 

Once Obs team and EMC can verify through a spot check that the 'good' obs have a value of 0, and all else, 1.

## Dependencies
None

## Impact

None

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -782,6 +782,12 @@ class Conv(BaseGSI):
                             else:
                                 tmp[tmp > 4e8] = self.FLOAT_FILL
                             outdata[gvname] = tmp
+                    # create a GSI effective QC variable
+                    gsiqcname = outvars[o], 'GsiEffectiveQC'
+                    errname = outvars[o], 'GsiFinalObsError'
+                    gsiqc = np.zeros_like(obsdata)
+                    gsiqc[outdata[errname] == self.FLOAT_FILL] = 1
+                    outdata[gsiqcname] = gsiqc
                     # store values in output data dictionary
                     outdata[varDict[outvars[o]]['valKey']] = obsdata
                     outdata[varDict[outvars[o]]['errKey']] = obserr
@@ -1306,6 +1312,12 @@ class Radiances(BaseGSI):
             outdata[varDict[value]['valKey']] = obsdatasub
             outdata[varDict[value]['errKey']] = obserrsub
             outdata[varDict[value]['qcKey']] = obsqcsub.astype(int)
+            # create a GSI effective QC variable
+            gsiqcname = value, 'GsiEffectiveQC'
+            errname = value, 'GsiFinalObsError'
+            gsiqc = np.zeros_like(obsdatasub)
+            gsiqc[outdata[errname] == self.FLOAT_FILL] = 1
+            outdata[gsiqcname] = gsiqc
             if (ObsBias):
                 valuebc = [
                     "constant_{:d}".format(chanlist[c]),
--- a/test/testoutput/amsua_aqua_obs_2018041500.nc4
+++ b/test/testoutput/amsua_aqua_obs_2018041500.nc4
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:269801067216b3da5c8a064dce88e8bc5c5e1cb3fb75235d63e884d47a8a573c
-size 135958
+oid sha256:909057dc00d687a1209195c2dc504d69c846974ce78a2fbea0887c085f8f5c41
+size 148568
--- a/test/testoutput/satwind_obs_2018041500.nc4
+++ b/test/testoutput/satwind_obs_2018041500.nc4
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:01f2791ede347e45c3361d319f2036e82f67f1de009d0e479e0ecf98ba28da21
-size 42528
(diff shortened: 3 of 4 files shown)


ADCollard (commented):
Question:  This functionality should be almost identical to QC_flag with the only difference being QC_flag=0 means that the ob is used.  Is this duplication really required?

fabiolrdiniz:
> Question: This functionality should be almost identical to QC_flag with the only difference being QC_flag=0 means that the ob is used. Is this duplication really required?

I believe that is not identical because GSI inflates errors without changing the QC flag.

ADCollard:
> > Question: This functionality should be almost identical to QC_flag with the only difference being QC_flag=0 means that the ob is used. Is this duplication really required?
> 
> I believe that is not identical because GSI inflates errors without changing the QC flag.

I'm not sure I understand.  If the errors are inflated, the observation is still used, so the flag should still be 0.  Unless you are suggestion a threshold on the obs error where the observation is no longer considered used?  How do we determine the threshold?

src/gsi-ncdiag/gsi_ncdiag.py
```diff
@@ -782,6 +782,12 @@ def toIODAobs(self, OutDir, clobber=True, platforms=None):
                             else:
                                 tmp[tmp > 4e8] = self.FLOAT_FILL
                             outdata[gvname] = tmp
+                    # create a GSI effective QC variable
+                    gsiqcname = outvars[o], 'GsiEffectiveQC'
+                    errname = outvars[o], 'GsiFinalObsError'
```
fabiolrdiniz, line 787: I believe it's, but certainly, the others can say about it better: is these criteria based on obs errors sufficient to encapsulate the conventional & ozone cases too (not only radiances)?

fabiolrdiniz (approved):
Thanks, @CoryMartin-NOAA. I left a question to everyone over the code.

fabiolrdiniz:
> > > Question: This functionality should be almost identical to QC_flag with the only difference being QC_flag=0 means that the ob is used. Is this duplication really required?
> > 
> > 
> > I believe that is not identical because GSI inflates errors without changing the QC flag.
> 
> I'm not sure I understand. If the errors are inflated, the observation is still used, so the flag should still be 0. Unless you are suggestion a threshold on the obs error where the observation is no longer considered used? How do we determine the threshold?

Apparently, from what I've seen in the changes in this PR, the obs errors are already coming as FLOAT_FILL (```gsiqc[outdata[errname] == self.FLOAT_FILL] = 1```).

ADCollard:
> > > > Question: This functionality should be almost identical to QC_flag with the only difference being QC_flag=0 means that the ob is used. Is this duplication really required?
> > > 
> > > 
> > > I believe that is not identical because GSI inflates errors without changing the QC flag.
> > 
> > 
> > I'm not sure I understand. If the errors are inflated, the observation is still used, so the flag should still be 0. Unless you are suggestion a threshold on the obs error where the observation is no longer considered used? How do we determine the threshold?
> 
> Apparently, from what I've seen in the changes in this PR, the obs errors are already coming as FLOAT_FILL (`gsiqc[outdata[errname] == self.FLOAT_FILL] = 1`). Above in the code, the threshold is being defined as 4e8 (`obserr[obserr > 4e8] = self.FLOAT_FILL`).

In that case is this equivalent to Analysis_Use_Flag as in the GSI that includes cases where the ob error is inflated to make the effective obs error effectively zero?
        obsqc = self.var('Analysis_Use_Flag').astype(int)

huishao-r:
> > > > > Question: This functionality should be almost identical to QC_flag with the only difference being QC_flag=0 means that the ob is used. Is this duplication really required?
> > > > 
> > > > 
> > > > I believe that is not identical because GSI inflates errors without changing the QC flag.
> > > 
> > > 
> > > I'm not sure I understand. If the errors are inflated, the observation is still used, so the flag should still be 0. Unless you are suggestion a threshold on the obs error where the observation is no longer considered used? How do we determine the threshold?
> > 
> > 
> > Apparently, from what I've seen in the changes in this PR, the obs errors are already coming as FLOAT_FILL (`gsiqc[outdata[errname] == self.FLOAT_FILL] = 1`). Above in the code, the threshold is being defined as 4e8 (`obserr[obserr > 4e8] = self.FLOAT_FILL`).
> 
> In that case is this equivalent to Analysis_Use_Flag as in the GSI that includes cases where the ob error is inflated to make the effective obs error effectively zero? obsqc = self.var('Analysis_Use_Flag').astype(int)

In the converted ioda obs files (from GSI diag), there is only "preQC". In Jan's code sprint, we found it was not consistent with obs inflation results (JEDI consider obs inflation as one of the QC as well). Is this preQC same as "analysis_use_flag" or something else?

ADCollard:
Having discussed this with @CoryMartin-NOAA , this appears to be a temporary fix for a specific purpose.  I am concerned about a proliferation of QC variables as has happened in the GSI, but I happy for this to proceed.

ADCollard (approved):

srherbener (approved):

srherbener:
There are several open questions. Are the reviewers okay with merging now? My preference is to get this merged and address any remaining issues in subsequent PRs. Thanks!

huishao-r (approved):

PR #642, opened 2021-10-22
Feature/v2profile

rhoneyager:
## Description

This PR ports marine/godae_profile2ioda.py and marine/hgodas_insitu2ioda.py to use the new ioda-v2 interface.


srherbener:
Are we waiting on this PR for the other converter upgrades since we used this as an example case?

CoryMartin-NOAA:
> Are we waiting on this PR for the other converter upgrades since we used this as an example case?

I think so, yes. Plus I think it still needs finished up.

CoryMartin-NOAA:
This PR/branch still needs completed as there are things like units, etc. missing. Any volunteers?

srherbener:
> This PR/branch still needs completed as there are things like units, etc. missing. Any volunteers?

I'm fixing this during the review sprint.

srherbener:
I believe that I have put in all of the missing variable attributes (units, _FillValue). However, I don't know what some of the units are supposed to be so I put in a value of '???' and the tests will fail.

Could someone please let me know (or replace the '???' values) what the units are supposed to be for all of the variables.

Thanks!

srherbener:
I've done a number of code changes in this branch. I should probably not be a reviewer because of that. Who is a good person to replace me as a reviewer? thanks

CoryMartin-NOAA:
> I've done a number of code changes in this branch. I should probably not be a reviewer because of that. Who is a good person to replace me as a reviewer? thanks

As a marine profile, I suggest perhaps @HamidehGMAO or @kbhargava or @lliu1220

src/marine/godae_profile2ioda.py
```diff
@@ -169,25 +169,29 @@ def __init__(self, filename, date, varDict, obsList):
             ("datetime", "string")
         ]
 
-        self.AttrData = {
+        self.GlobalAttrs = {
             'odb_version': 1,
             'date_time_string': self.date.strftime("%Y-%m-%dT%H:%M:%SZ")
```
kbhargava, line 174: The PR looks good to me. On minor thing: I think it was decided to remove the date_time _string here, not sure if it matters right now or not. I forgot what the reason was but @rhoneyager or @CoryMartin-NOAA remember may be.

kbhargava (approved):

src/marine/godae_profile2ioda.py
srherbener: @kbhargava thanks for the review! I'll remove the date_time_string attribute.

src/marine/godae_profile2ioda.py
```diff
@@ -254,11 +259,17 @@ def main():
         obsList.append(obs)
 
     varDict = {
-        'ob_tmp': 'sea_water_temperature',
-        'ob_sal': 'sea_water_salinity'
+        #              var name,             units
+        'ob_tmp': ['sea_water_temperature', '???'],
+        'ob_sal': ['sea_water_salinity', '???']
```
srherbener, line 264: Can someone please tell me what the units values should be for these variables? Thanks!

src/marine/hgodas_insitu2ioda.py
```diff
@@ -19,13 +19,14 @@
...
-import ioda_conv_ncio as iconv
+import ioda_conv_engines as iconv
 from orddicts import DefaultOrderedDict
 
 
 vName = {
-    2210: "sea_water_temperature",
-    2220: "sea_water_salinity",
+    #          var name,            units
+    2210: ["sea_water_temperature", "???"],
```
srherbener, line 28: Can someone please tell me what the units value should be for this variable? Thanks!

src/marine/hgodas_insitu2ioda.py
```diff
@@ -19,13 +19,14 @@
...
+    2220: ["sea_water_salinity", "PSU"]
```
srherbener, line 29: Is "PSU" the correct units value for this variable? (I simply copied from the ioda-v1 version of this script.) Thanks!

src/marine/godae_profile2ioda.py
CoryMartin-NOAA: Probably `K` and `PSU`, respectively, but I'm not a marine expert and will defer to others.

src/marine/hgodas_insitu2ioda.py
kbhargava: I thought you left placeholders on purpose to be imported from the input file in future . From looking at the data, it looks like Celsius for temperature and PSU for salinity.

src/marine/godae_profile2ioda.py
srherbener: Thanks!

src/marine/hgodas_insitu2ioda.py
srherbener: Thanks. I'll use "C" and "PSU".

I was just forcing the tests to fail so that would remind us that the units still needed to get fixed.

src/marine/godae_profile2ioda.py
```diff
@@ -214,14 +217,15 @@ def __init__(self, filename, date, varDict, obsList):
...
+                        self.data[locKey][valKey] = val
+                        self.data[locKey][errKey] = err
+                        self.data[locKey][qcKey] = qc
 
-        (ObsVars, LocMdata, VarMdata) = self.writer.ExtractObsData(self.data)
-        self.writer.BuildNetcdf(ObsVars, LocMdata, VarMdata, self.AttrData)
+        ObsVars, nlocs = iconv.ExtractObsData(self.data, self.locKeyList)
+        DimDict = {'nlocs': nlocs}
+        self.writer = iconv.IodaWriter(self.filename, self.locKeyList, DimDict)
+        self.varAttrs['depth', 'MetaData']['units'] = "???"
```
CoryMartin-NOAA, line 227: Is this `meters`? @kbhargava do you know?

src/marine/hgodas_insitu2ioda.py
```diff
@@ -64,21 +67,31 @@ def _read(self):
 
         base_date = datetime(1970, 1, 1) + timedelta(seconds=int(time[0]))
 
+        self.VarAttrs['depth', 'MetaData']['units'] = '???'
```
CoryMartin-NOAA, line 70: Same question as above.

CoryMartin-NOAA (commented):
Just need to figure out the units; update the test files; and then we are ready to merge I think.

srherbener:
> Just need to figure out the units; update the test files; and then we are ready to merge I think.

I agree. Once these units (and test files) are update, this PR should be ready to merge. Thanks everyone for your help with this.

CoryMartin-NOAA (approved):

PR #700, opened 2021-11-10
Add ability to convert GNSSRO data in WMO-formatted BUFR to IODA (v2) output

gthompsnJCSDA:
## Description

Add capability to convert WMO BUFR formatted GNSSRO data to IODA directly with more flexible bufr2ioda application instead of fortran converter.  One problem is the Quality Flags remain a 16-bit integer whereas the existing IODA file does need the derived value of ascending and descending pass (bit position 3).  Therefore more processing of the quality flags will be needed within UFO.  This PR also adds documentation for BUFR mnemonics and sample obs into the doc subdirectory for additional data usage if needed.

### Issue(s) addressed

This item in particular does not have a corresponding issue number but moving towards BUFR-2-IODA in general needs to include as many data types as possible.

Successful completion of ctest and useful IODA (v2) file for use within UFO (QC and subsequent steps).

## Dependencies

No dependencies.

## Impact

No impacts to other repos.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -154,7 +154,7 @@ if( iodaconv_bufr_ENABLED )
     testoutput/NC005066.nc
     testoutput/airep_multi.nc
     testoutput/amdar_single.nc
-    testinput/gnssro_2020-306-2358C2E6.nc
+    testoutput/gnssro_2020-306-2358C2E6.nc
   )
 endif()
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -136,6 +136,8 @@ if( iodaconv_bufr_ENABLED )
     testinput/airep_wmo_multi.bufr
     testinput/amdar_wmoBUFR2ioda.yaml
     testinput/amdar_wmo_single.bufr
+    testinput/gnssro_wmoBUFR2ioda.yaml
+    testinput/gnssro_2020-306-2358C2E6.bufr
   )
 
   list( APPEND test_output
@@ -152,6 +154,7 @@ if( iodaconv_bufr_ENABLED )
     testoutput/NC005066.nc
     testoutput/airep_multi.nc
     testoutput/amdar_single.nc
+    testinput/gnssro_2020-306-2358C2E6.nc
   )
 endif()
 
@@ -777,6 +780,14 @@ if(iodaconv_bufr_ENABLED)
                     ARGS    netcdf
                     "${CMAKE_BINARY_DIR}/bin/bufr2ioda.x testinput/amdar_wmoBUFR2ioda.yaml"
                     amdar_single.nc ${IODA_CONV_COMP_TOL_ZERO}
+                    DEPENDS bufr2ioda.x )
+
+  ecbuild_add_test( TARGET  test_iodaconv_bufr_gnssro_wmo_bufr
+                    TYPE    SCRIPT
+                    COMMAND ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                    ARGS    netcdf
+                    "${CMAKE_BINARY_DIR}/bin/bufr2ioda.x testinput/gnssro_wmoBUFR2ioda.yaml"
+                    gnssro_2020-306-2358C2E6.nc ${IODA_CONV_COMP_TOL_ZERO}
                     DEPENDS bufr2ioda.x )
 endif()
 
--- a/doc/bufr_obs_samples/gnssro_bufr_sample.txt
+++ b/doc/bufr_obs_samples/gnssro_bufr_sample.txt
@@ -1,0 +1,226 @@
+
+Found BUFR message #      1
+  
+        Message length:       17278
+      Section 0 length:           8
+          BUFR edition:           4
+  
+      Section 1 length:          22
+          Master table:           0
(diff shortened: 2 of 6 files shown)


srherbener:
@gthompsnJCSDA, the error message:
```
2021-11-09 19:14:37.195752 -0700 ERROR ../../src/nccmp_state.c:156 No such file or directory
```
from the test failure is saying that one of the input files to the nccmp command is not found.

test/CMakeLists.txt
```diff
@@ -152,6 +154,7 @@ if( iodaconv_bufr_ENABLED )
     testoutput/NC005066.nc
     testoutput/airep_multi.nc
     testoutput/amdar_single.nc
+    testinput/gnssro_2020-306-2358C2E6.nc
```
srherbener, line 157: I think this is the cause of the test failure. Need to change "testinput" to "testoutput" on this line.

srherbener (changes requested):

test/CMakeLists.txt
gthompsnJCSDA: > I think this is the cause of the test failure. Need to change "testinput" to "testoutput" on this line.

Yup, dumb mistake and got confused by the error message regarding a missing source code file, not an error message about a missing comparison file.  Is this cause for a different issue to be created?  Thanks for catching my error.

hailingz (commented):
Is gnssro_2020-306-2358C2E6.nc the final output?

I took a look at this file. i have some questions as below,
1. some MetaData are missing, for example:
        occultation instrument id, which was occulting_sat_is in the fortran output
        process_center
        record_number

2. qualityFlags seems to be undecoded, the sample output is "16448", which is a 16 bit indicator representing 16 QC flags. do we expect to decode this in UFO?

3. do we need some basic preQC check in the converter step, e.g.,  excluding bad reports (non-nominal data).

4. I am not sure where do you get obsvalues for "airTemperature" and "specificHumidity". do we need them?

test/CMakeLists.txt
srherbener: If you are talking about the not-so-helpful error message, it would be nice if nccmp would list the file name in the message but there doesn't seem to be an option to do that. We could issue a bug report with the nccmp developers but the message is good enough to narrow it down to the two file arguments. Not sure if it's worth pursuing this.

srherbener (approved):
Looks good to me, but I'm not expert on the required variables for GNSSRO. Thanks!

gthompsnJCSDA:
> Is gnssro_2020-306-2358C2E6.nc the final output?

Yes.
 
> 1. some MetaData are missing, for example:
>    occultation instrument id, which was occulting_sat_is in the fortran output

This was in the file already - you must have missed it.

>    process_center

This was missing but easy for me to add because I was capturing the mnemonic - so I just added it.

>    record_number

This cannot be created from a single BUFR record - so we have to consider how to re-generate this variable.

> 2. qualityFlags seems to be undecoded, the sample output is "16448", which is a 16 bit indicator representing 16 QC flags. do we expect to decode this in UFO?

This was already mentioned in the "description" section of the PR.  This is a pure data-input => data-output, so if the value of a "short" integer must be converted into a series of 16 bits and the 3rd bit represents the ascending/descending flag.  This will have to be done now by UFO code, not ioda-converters.

> 3. do we need some basic preQC check in the converter step, e.g.,  excluding bad reports (non-nominal data).

This is part of QC and isn't really relevant to the converter being discussed for this PR.

> 4. I am not sure where do you get obsvalues for "airTemperature" and "specificHumidity". do we need them?

These are available variables from the BUFR - see the additions in the "doc/bufr_sample_tables" for all possible mnemonics in the input data file.  Since others might want to use them directly, I thought it wise to include in the converter.  If no one uses them, obviously they can be skipped.  I included them for completeness and because DA systems almost always want to assimilate temperature and moisture values.


It is my opinion that all items you commented upon have been resolved.  This isn't meant as the **replacement** to the fortran converter yet.  More work will be needed to accomplish that considering your bullet-2 in particular.  This is just a step towards getting there.

hailingz:
Hi @gthompsnJCSDA 
The one you have is **occulting_sat_id**, which is the low earth orbit (LEO) RO satellite identifier. For example, 5 is MeTop-Cc, 750 is one of the COSMIC2.  The missing one I asked is **occulting_sat_is**, which is the instrument identifier. For example, 98 is GPS receiver, 103 is the Integrated GPS and Occultation Receiver (IGOR).

hailingz:
I understood what you said. There are 16 bits of QC in the bufr file. some of them are used in PreQC. 
I am not sure if all of them should be done in UFO stage. They are PreQC not regular QC.
What do you think? @fcvdb @huishao-r @rusbec 
------
qualityFlags seems to be undecoded, the sample output is "16448", which is a 16 bit indicator representing 16 QC flags. do we expect to decode this in UFO?
This was already mentioned in the "description" section of the PR. This is a pure data-input => data-output, so if the value of a "short" integer must be converted into a series of 16 bits and the 3rd bit represents the ascending/descending flag. This will have to be done now by UFO code, not ioda-converters.

hailingz:
@gthompsnJCSDA 
another question is just naming conventional.
You put "airTemperature" and "specificHumidity" in obsvalue group. What I understand is that those are not observables. Rather, they are derived by additional process using some model background.

hailingz (approved):

PR #718, opened 2021-12-03
Feature/add atms ingest scripts

BenjaminRuston:
## Description

This adds a bash script to pull data for a day or series of days for the ATMS sensor from the NASA archive.

Script can pull data from any machine HPC or laptop to local drive


fabiolrdiniz:
Hi @BenjaminRuston, thanks for adding me here. Please, see my comments below.

I believe it's needed to have a username & password from https://disc.gsfc.nasa.gov to get access to this data, correct? If so, should this information be added somewhere? Perhaps inside a `README.md` file living on the same folder as the script.

I noticed that there is no `CMakeLists.txt` under the `tools/ingest` folder and no mention about this new `ingest` folder inside the `tools/CMakeLists.txt` file, so I believe this script is not being "installed". Is this intentional to not save it under the `bin` directory?

There is a typo in the script name `tools/ingest/retreive_atms_from_nasa.sh` ---> `tools/ingest/retrieve_atms_from_nasa.sh`.

srherbener (approved):
Code looks good. I'm not an expert on these data sources, so it would be good to have a review from that perspective. Thanks!

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

PR #719, opened 2021-12-06
Feature/generic gnss ro decoder to ioda

BenjaminRuston:
## Description

add a python based GNSS-RO converter BUFR to IODA v2

Successfully write a IODAv2 file from the WMO BUFR GNSS-RO data

## Dependencies

ecCodes and ecCodes python interface


srherbener:
I got the Infinity issue fixed. The first nine values of ObsValue variables and lat, lon are now marked as missing values. 

I also updated the test reference file. Before doing that, I carefully checked the test output with the prior test reference file and verified that all the variables matched. The prior test reference file was missing the MetaData group and had all variables declared as double instead of float.

The output of this converter looks like it might be ready to try in an H(x) flow.

srherbener:
CI tests are failing because eccodes is not yet installed in the CI test containers. I'll work on getting this remedied.

huishao-r (approved):

PR #724, opened 2021-12-07
Feature/add noaa nomads ingest script

### LINKED ISSUE #723: add a NOAA NOMADS ingest script
## Description
some data is available through NOAA NOMADS for the most recent 10 days from real time
create an ingest script with default types ingested currently:

        1bamua
        1bmhs
        esamua
        esmhs
        gpsro
        mtiasi

optional user input will allow specification of a data type

## Requirements
this depends on wget 

can reliably pull data from NOAA NOMADS these are 6-hourly windowed data files

BenjaminRuston:
## Description

add an ingest script for NRT data from NOAA NOMADS server
this is 6-hourly binned data used by GFS

### Issue(s) addressed

- fixes #723 

files can be successfully pulled from NOAA server


tools/ingest/retrieve_data_from_noaa_nomads.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/sh
```
rickgrubin, line 1: Suggestion: `/bin/bash`

You reference the use of `bash` on line 19 in the _usage_ statement setting.

Also: set the exec bits for the script and obviate the need for exec'ing a (sub) shell when running the script.

BenjaminRuston: will change line 1

try to always set scripts with the exec bits on initial commit will make sure they are set, but still usually from old habits launch as a subshell when running scripts

will leave the recommended _usage_ statement unless large objection

rickgrubin: If you set the script to be executable, then it can be launched as

`script args`

and not as

`bash script args` (which starts a sub-shell)

This would necessarily change your `usage` statement.

These are only suggestions; I'm not a reviewer.

BenjaminRuston:
Rick,

I follow your comment,  and understand 'bash script' starts a subshell, and
if executable this preface of 'bash' isn't needed.

It's just my convention, and generally do this in my usage/example
statements.  Again, as far as I can tell the executable bit has been set on
the file correctly so it will work either way.

This is a very vanilla script so will work equally well in a subshell and
does not rely on any non-shell items except wget.  so figure it's pretty
safe.

Is there an advantage to not running in a sub-shell?

btw, and your comments are welcome,,, when I went back over the script
found some more junk and other small cleanup to do.

definitely always feel welcome to drop comments if you have time.

best
ben

rickgrubin:
> Is there an advantage to not running in a sub-shell? 

Within a subshell (which is a separate instance of a 'parent' shell), variables outside of a block of code may not accessible by the parent process -- they are, in effect, local variables to the child process. If you're relying on values of certain variables, they may not be what's expected. As well, changes to variables within a child process do not impact the parent process.

For `bash`, if you really want subshells, the variables `BASHPID` and `SHLVL` are helpful to consult, to figure out where you might be (parent or child). These are not necessarily portable.

If you want portability, then an idiom like:

```
if [ "$(exec sh -c 'echo "$PPID"')" != "$$" ]; then
    echo you\'re in a subshell
fi
```

is useful.

My approach is that with a script (or any code), you never how how it may get used in the end, so I try to write with that (seemingly inevitable) result in mind.

tools/ingest/retrieve_data_from_noaa_nomads.sh
```diff
@@ -0,0 +1,72 @@
...
+#         will default to AMSU-A, MHS, GNSS-RO and IASI
+#
+#      NOTE: NOMADS keeps ~10days of data online
+#
+#   REQUIREMENTS:
+#       wget
+#
+##################################################################
+
+nomads_address='https://nomads.ncep.noaa.gov/pub/data/nccf/com/gfs/prod'
```
CoryMartin-NOAA, line 17: `gfs` == early cycle and `gdas` == late cycle. Suggest either 1) changing to `gdas` everywhere or 2) making `dump` another option for the user.

tools/ingest/retrieve_data_from_noaa_nomads.sh
```diff
@@ -0,0 +1,72 @@
...
+        esamua
+        esmhs
+        gpsro
+        mtiasi
+        )
+fi
+
+get_files() {
+    # retrieve the files
+    gfile=${nomads_address}/gfs.${dtg:0:8}/${dtg:8:2}/atmos/gfs.t${dtg:8:2}z.${atype}.tm00.bufr_d
```
CoryMartin-NOAA, line 55: Same as above `gdas` vs `gfs`

tools/ingest/retrieve_data_from_noaa_nomads.sh
BenjaminRuston: excellent,,, I never knew about this

could have it try `gdas` first and only try `gfs` if no data is found?

tools/ingest/retrieve_data_from_noaa_nomads.sh
BenjaminRuston: understood

tools/ingest/retrieve_data_from_noaa_nomads.sh
CoryMartin-NOAA: So I misspoke. This URL should be `gfs` regardless, but the below in `get_files` can be either `gfs` or `gdas`.

I think for super near real time, `gfs` will exist before `gdas` (there is about a 3 hour difference between the two). In terms of observations, `gdas` will have everything in `gfs` plus all late arriving obs. So I would say yes, have it try `gdas` first and then fail to `gfs`.

CoryMartin-NOAA: For example: `https://nomads.ncep.noaa.gov/pub/data/nccf/com/gfs/prod/gdas.20211210/12/atmos/` vs `https://nomads.ncep.noaa.gov/pub/data/nccf/com/gfs/prod/gfs.20211210/12/atmos/`

fabiolrdiniz (commented):
Thanks, @rusbec! Minor suggestion.

fabiolrdiniz:
I don't know what is the idea to approve/merge this PR, but I was able to retrieve all the default and non-default observation types for a given day (e.g., `1bamua` `1bhrs4` `1bmhs` `adpsfc` `adpupa` `atms` `atovs` `bathy` `crisf4` `eshrs3` `esmhs` `gome` `gpsipw` `gpsro` `mtiasi` `osbuv8` `satwnd` `sevcsr` `sfcshp` `ssmisu` `tesac` `trkob`).

tools/ingest/retrieve_data_from_noaa_nomads.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+##################################################################
+# 
+#  PURPOSE:
+#       retrieve BUFR data from NOAA NOMADS server
+#         optionally provide a data type
+#         will default to AMSU-A, MHS, GNSS-RO and IASI
+#
+#      NOTE: NOMADS keeps ~10days of data online
```
lliu1220, line 10: script works well if only retrieve **most recent 10 days** of data.
thank you.

lliu1220 (approved):
scripts works fine for most recent 10 days of data download

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

PR #731, opened 2021-12-14
Feature/avhrr radiance2ioda

HamidehGMAO:
## Description

this is a ioda-converter to convert the AVHRR radiance data to ioda-v2 fprmat. there are several sources of data for AVHRR, this converter is for patmosx data in nc format "https://www.ncei.noaa.gov/data/avhrr-hirs-reflectance-and-cloud-properties-patmosx/access/"

### Issue(s) addressed

- fixes #<730>


ShastriPaturi (approved):
Ctest passes.
Looks good.

src/marine/avhrr_radiance2ioda.py
```diff
@@ -0,0 +1,286 @@
+#!/usr/bin/env python3
+
+#
+# (C) Copyright 2019-2021 UCAR
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+
+from __future__ import print_function
```
srherbener, line 10: This isn't required anymore. It was originally put in to be able to use python3 print syntax while running python2.

src/marine/avhrr_radiance2ioda.py
```diff
@@ -0,0 +1,286 @@
...
+    # (needed because we are bypassing ExtractObsData within BuildNetcdf)
+    VarDims = {
+        'brightness_temperature': ['nlocs', 'nchans']
+    }
+
+    nchans = len(chan_number)
+    nlocs = len(obs_data[('longitude', 'MetaData')])
+    DimDict = {
+        'nlocs': nlocs,
+        'nchans': nchans
```
srherbener, line 270: This is resulting in the nchans variable being set to all zeros, where it should be set to the list of channel numbers. The API will accept a list of the channel numbers which will then assign those channel numbers to the nchans variable. You should be able to fix this by changing this line to:

`'nchans': list(chan_number)`

srherbener (changes requested):
Thanks for contributing this! Everything looks great except that the nchans variable is getting set to all zeros. Please see the comment for a recommended fix. Thanks!

src/marine/avhrr_radiance2ioda.py
HamidehGMAO: Ah I see, Thanks for the information. I'll delete it

src/marine/avhrr_radiance2ioda.py
HamidehGMAO: I see. Thanks

srherbener (approved):
thanks for addressing my comments! Sorry about the delay getting back to this.

PR #745, opened 2022-01-18
Add BUFR-to-IODA support for Aircraft data: gdas.aircar files (similar to amdar_wmoBUFR2ioda.yaml)

gthompsnJCSDA:
## Description

This is rather similar to what is performed with amdar_wmoBUFR2ioda.yaml, except that the input data are reprocessed NCEP-EMC GDAS BUFR files available from NOMADS and also NCAR-RDA (ds351.0) in files with reference name of gdas.aircar.(date).bufr in place of WMO-formatted BUFR.  The mnemonics are somewhat different with these files compared to WMO BUFR data.  The mnemonics are listed/described as part of ```doc/bufr_table_samples/aircft.nc004004.bufr.table.txt``` because it is the subset that is referred to with NC004004 only.  Various other aircraft data (AIREPs, PIREPs, etc.) are found in other subsets and the gdas.aircft.(date).bufr files, which would need to be split out into separate chunks for different yaml files because of using different mnemonics depending on the subset of data.

This PR should permit the use of some variables (```air_temperature``` and ```specific_humidity```) while others need variable transformations to get into JEDI (wind direction and speed need to be changed to ```eastward_wind``` and ```northward_wind```).

### Issue(s) addressed

A separate issue was not created for this

When all tests pass.

## Dependencies

There are no dependencies.

## Impact

No impacts to other repos.

### DIFF
--- a/test/testinput/amdar_wmoBUFR2ioda.yaml
+++ b/test/testinput/amdar_wmoBUFR2ioda.yaml
@@ -6,7 +6,8 @@
 observations:
   - obs space:
       name: bufr
-      obsdatain: "./testinput/amdar_wmo_multi.bufr"
+#     obsdatain: "./testinput/amdar_wmo_multi.bufr"
+      obsdatain: "/home/gthompsn/2021_07_31T18_00_00Z.P6H.amdar.BUFR"
       isWmoFormat: true
       tablepath: "./testinput/bufr_tables"
 
@@ -62,7 +63,8 @@ observations:
 
     ioda:
       backend: netcdf
-      obsdataout: "./testrun/amdar_wmo_multi.nc"
+#     obsdataout: "./testrun/amdar_wmo_multi.nc"
+      obsdataout: "./testrun/2021_07_31T18_00_00Z.P6H.amdar.nc"
 
       dimensions:
         - name: "nlocs"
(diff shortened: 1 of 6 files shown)


test/testinput/amdar_wmoBUFR2ioda.yaml
```diff
@@ -152,7 +152,7 @@ observations:
           coordinates: "longitude latitude"
           source: variables/waterVaporMixingRatio
           dimensions: [ "nlocs" ]
-          longName: "Dewpoint Temperature"
+          longName: "specific humidity"
```
BenjaminRuston, line 155: is this mixing ratio and not specific humidity?  we should keep track of these as they can be added to the derived variables list.   Even though the change is very small, impact for the small amounts of vapor in the stratosphere may be important.

gthompsnJCSDA: Well, the line in question itself is just a fix for the ```longName``` that I had left over from some yaml where I copied-pasted.  But your point is valid that I am just making a super fast fix to get **any** moisture data from aircraft into UFO, even though it will be slightly off due to the difference with water vapor mixing ratio and specific humidity, which is about 0.01-0.03% different.  I know this needs fixing with a VariableTransform in the future; just want to use moisture rather than skip it due to the other problems with derived versus native variables in IODA.

BenjaminRuston (approved):

nicholasesposito (approved):
Looks good to me

rhoneyager (approved):

PraveenKumar-NOAA:
Looks good.

PR #746, opened 2022-01-19
 add non-nominal check for processing near real time files

BenjaminRuston:
## Description

When using raw GNSS-RO observations particularly in a near real-time feed occultation profiles deemed non-nominal are generally removed.  

Successful run through UFO comparing number of occultations from this pathway with that from an alternate feed where non-nominal profiles have been removed.


fcvdb (approved):
The changes look good. With a little bit of time, I can run some tests on the 2022/01/10 case to see the impact before and after the changes.

hailingz (approved):

PR #761, opened 2022-01-25
Switch from using string-format datetime variable to the newer dateTime (seconds since epoch) in bufr2ioda.x application

gthompsnJCSDA:
## Description

For efficiency improvements, change the bufr2ioda.x application to switch from prior string-formatted ```datetime``` variable to the newer integer-based ```dateTime``` variable.  Hard-coded at this time, all observations will be based on seconds since 1970-01-01T00:00:00Z (typical UNIX epoch).  In the future, the units attribute could be used to set a different epoch date.

A positive outcome of this change is that any BUFR message that has missing date information will be properly handled and given to IODA as missing values.

Rather importantly, any YAML that previously had a ```datetime``` output variable needs to be changed to ```dateTime``` and must now be assigned the units of ```seconds since 1970-01-01T00:00:00Z```.  As such, there will be many files touched in ```testinput``` as well as ```testoutput``` that simply have this piece of data changed.

### Issue(s) addressed

TBD: will create one for tracking purposes

When the tests all pass along with reviews.

## Dependencies

None.

## Impact

No impacts to other repos are expected.


src/bufr/DataObject/Int64VecDataObject.cpp
```diff
@@ -0,0 +1,61 @@
+/*
+ * (C) Copyright 2022 UCAR
```
rmclaren, line 2: please update copyright to  * (C) Copyright 2022 NOAA/NWS/NCEP/EMC

src/bufr/DataObject/Int64VecDataObject.h
```diff
@@ -0,0 +1,61 @@
+/*
+ * (C) Copyright 2022 UCAR
```
rmclaren, line 2: copyright

src/bufr/BufrParser/Exports/Export.cpp
```diff
@@ -30,7 +30,7 @@ namespace
 
         namespace Variable
         {
-            const char* Datetime = "datetime";
+            const char* Datetime = "dateTime";
```
rmclaren, line 33: Why are we changing the name of datetime?

rmclaren: This will require changes in every YAML file in existents...

rmclaren (changes requested):
Don't forget to update files associated with Unit tests.

src/bufr/BufrParser/Exports/Export.cpp
gthompsnJCSDA: Every single YAML file already has to change - since we must have a units string attribute to make the data work in UFO or elsewhere.  And the new variable name is ```dateTime``` in place of ```datetime``` so I figured I would just do it all as needed and be consistent.

gthompsnJCSDA:
> Don't forget to update files associated with Unit tests.


I think I got them all, but waiting for auto-testing to complete to see if I missed any.

gthompsnJCSDA:
Switched from Draft to ```ready to review``` because I think I got all the testoutput files updated with newer versions so I believe the auto-testing will pass (fingers crossed).

src/bufr/BufrParser/Exports/Export.cpp
rmclaren: The datatime tag, in this context, just refers to the class of object you are creating and doesn't have anything to do with what you might decide to call it in the end (by convention or otherwise). We should not conflate the two. I'm going to be very resistant to making arbitrary changes like this to the publicly facing interface.

gthompsnJCSDA: @rmclaren   Here is my rationale:  I've done the work of changing about 25 yaml files already.  Only those in branches not yet merged to develop would need changing and those would already need changes because we must enforce the proper use of ```units``` to accept this change in its entirety without backward compatibility.  Furthermore, a complete sweeping change with one letter capitalized will help to enforce that the variable name in the ioda output section must be changed and cannot be forgotten.  So, while unnecessary, I believe it is helpful for anyone in future either modifying or creating new yamls to get this right and consistent; otherwise, I can see confusion by forgetting with letter is capitalized.

jamiebresch: From a user's point-of-view, it makes better sense to me to keep general `datetime` in the `exports:` part. `dateTime` is a specific defined quantity to ioda-v2 convention.

rmclaren: " I can see confusion by forgetting with letter is capitalized."

Yes exactly. `datetime` is like a built in keyword. Adding capitol letters not only makes it more likely that a user might spell it incorrectly, but it also makes it look like a user defined thing (usually camalCased in our case) as apposed to a built in thing which are lowercase words or conjunctions of words...

rmclaren (approved):

rmclaren:
Thanks Greg!  Sorry for being such a pain about things...

srherbener (approved):
I looked at a couple test reference files and they look good. I think the code changes look good too. Thanks for this contribution!

PR #795, opened 2022-02-11
In order for these bufr2ioda converters to work in UFO, there must be a height variable.

gthompsnJCSDA:
## Description

While these bufr2ioda converters work fine as-is, the output IODA file is missing the ```height``` variable which is needed for Obs Operator of SfcPCorrected.  Unfortunately, the original BUFR data does not contain the proper variable, so we just fake it for now by setting the height equal to station elevation.  To be proper, the height of the measurement should be above the height of the station, but we need a quick fix for right now.

### Issue(s) addressed

Passes tests.

## Dependencies

None

## Impact

No impacts to other repos.


fabiolrdiniz (approved):
Thanks, @gthompsnJCSDA! I don't know if comments in the code are welcome, but I believe that would be nice to have something mentioning about this approximation.

gthompsnJCSDA:
> Thanks, @gthompsnJCSDA! I don't know if comments in the code are welcome, but I believe that would be nice to have something mentioning about this approximation.


Agreed about the point, but what I am creating is all temporary and will be resolved pretty quickly.

srherbener (approved):

PR #798, opened 2022-02-17
add cleaner exit using curl to check for file on server

BenjaminRuston:
## Description

Currently the script passes back a non-zero exit from wget if the file does not exist

This happens because there are 4 files names checked for each data type for simplicity though only 2 typically exist.  Rather than explicitly trying to predict the specific 2 names just always check for 4 makes it much simpler

### Issue(s) addressed

- fixes #1890

Scripts runs and retrieves data

## Dependencies

wget

### DIFF
--- a/tools/ingest/retrieve_data_from_noaa_nomads.sh
+++ b/tools/ingest/retrieve_data_from_noaa_nomads.sh
@@ -21,7 +21,7 @@
 
 nomads_address='https://nomads.ncep.noaa.gov/pub/data/nccf/com/gfs/prod'
 
-usage="usage bash ${0} yyyymmdd [user_input_data_type]"
+usage="usage bash ${0} yyyymmddhh [user_input_data_type]"
 example="example:  bash ${0} 2021091500 [adpupa]"
 
 if (( ${#@} == 0 ))  ||  (( ${#@} > 2 )) || [[ $1 == [hH]elp ]]; then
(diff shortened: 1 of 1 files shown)


fabiolrdiniz (approved):
Thanks, @BenjaminRuston! Just tested to retrieve `gpsro` with:
`retrieve_data_from_noaa_nomads.sh 2022021700 gpsro` and it retrieved:
- `gdas.t00z.gpsro.tm00.bufr_d.nr`
- `gfs.t00z.gpsro.tm00.bufr_d.nr`

tools/ingest/retrieve_data_from_noaa_nomads.sh
```diff
@@ -62,23 +62,23 @@ get_files() {
...
-            if [[ ! -s ${out_file} ]]; then
-                rm -f ${out_file}
+        # check both "normal" name and one with restricted data stripped ( appended with "nr" )
+        for afile in ${gfile} ${nfile}; do
+            out_file=${afile##*/}
+            # optional rename
+            # out_file="${out_file%.bufr_d*}.bfr"
+            # does the file exist on the server
+            if curl --output /dev/null --silent --head --fail "${afile}"; then
+                wget -nc ${afile} -O ${out_file}
```
fcvdb, line 72: Why using curl and then get? could curl be used in both commands?

fcvdb (approved):

tools/ingest/retrieve_data_from_noaa_nomads.sh
rhoneyager: We should add both programs to the "stack" as required utilities. Currently we only deploy libraries. curl is omnipresent on Linux systems, but wget is occasionally missing.

PR #807, opened 2022-02-28
Python script to convert WMO-formatted BUFR messages of AMDARs (aircraft obs) to IODA

gthompsnJCSDA:
## Description

This PR allows AMDAR aircraft observations direct off NCAR-RAL's LDM feed of live data to be converted to IODA for use in UFO/JEDI.  This is an alternative to using bufr2ioda.x application (Ron McLaren's) because it also does conversion of winds from speed/direction into ```eastward_wind``` and ```northward_wind``` for immediate usage while waiting for larger changes to UFO for variable transforms.  This PR also converts from water vapor mixing ratio to ```specific_humidity``` variable needed in JEDI.  Simple constant assumptions are made for starting ```ObsError``` values that can easily be overridden or replaced by UFO filters (i.e., ```Perform Action``` and ```assign error``` found in ```ufo/ewok/skylab/amdar.yaml```) as well as setting the ```PreQC``` value to 2, which EMC uses for default, not yet quality controlled.

Near the top of the Python script, the long variable descriptions corresponding to WMO BUFR tables give a way to capture the variables, including either singular variables (e.g., ```tail_number```) or more complex mixtures such as ```height``` that could be 3 different BUFR variable names (taken with precedence in left-to-right order).

Perhaps with relatively lightweight changes, this same python script could be used for various other sources of similar aircraft data (TAMDAR, AIREP, etc.), but this is yet to be tested and implemented.

### Issue(s) addressed

A specific issue was not yet created.

Two or more positive reviews plus working ctests and fully useful output for next steps in UFO/JEDI

## Dependencies

Is this dependent on addition of another PR for the CMakeLists.txt file related to ```eccodes_FOUND``` from ```find_package```

## Impact

No impacts to other repos expected.


src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,505 @@
...
+
+    optional = parser.add_argument_group(title='optional arguments')
+    optional.add_argument('--debug', action=argparse.BooleanOptionalAction,
+                          help='enable debug messages')
+    optional.add_argument('--verbose', action=argparse.BooleanOptionalAction,
+                          help='enable verbose debug messages')
+
+    args = parser.parse_args()
+
+    debug = True if args.debug else False
```
climbfuji, line 497: If you set default values for debug and verbose, then you can simply assign the args.* values without any if statements.

climbfuji (commented):
I would recommend looking into the Python logging module.

It seems to me that your default print statements all correspond to the WARNING level, your verbose print statements to INFO, and your debug print statements to DEBUG. You can also replace things like lines 308-309 with a single logging.critical line and configure the logger to raise an exception for critical messages.

When using the Python logging module, you can just have one cmd line argument `-v`, which can be repeated to give more output (one -v changes the default WARNING level to INFO, two -vv increase it to DEBUG).

In general, I would not catch exceptions, print an error message manually and then use sys.exit, but rather throw the exception as-is or catch it and throw an exception using logging.critical.

src/conventional/amdar_bufr2ioda.py
gthompsnJCSDA: @climbfuji  Not sure I understand except to say that what's the difference here?  So I set False as one line, then I set True if args.debug is true?  Seems worse than just these simple one-liners to me.  I am still very new to Python, so I might be mis-interpreting your thought here.

Meanwhile, I have changed a bunch of my print statements to utilize the logging module - great piece of advice, thanks.

climbfuji: Here is one way (not the only way) to implement this, with some additional format specifiers for the log output:
```
#!/usr/bin/env python3

import argparse
import logging

parser = argparse.ArgumentParser()
parser.add_argument('-v', '--verbose', action='count', default=0)
args = parser.parse_args()

levels = [logging.WARNING, logging.INFO, logging.DEBUG]
level = levels[min(len(levels)-1,args.verbose)]  # capped to number of levels

logging.basicConfig(level=level,
                    format="%(asctime)s %(levelname)s %(message)s")

logging.debug("a debug message")
logging.info("a info message")
logging.warning("a warning message")
```
This gives:
```
JCSDA-L-18146:Desktop heinzell$ mate logging_test.py
JCSDA-L-18146:Desktop heinzell$ python logging_test.py
2022-02-28 17:16:38,941 WARNING a warning message
JCSDA-L-18146:Desktop heinzell$ python logging_test.py -v
2022-02-28 17:16:42,460 INFO a info message
2022-02-28 17:16:42,460 WARNING a warning message
JCSDA-L-18146:Desktop heinzell$ python logging_test.py -vv
2022-02-28 17:16:44,778 DEBUG a debug message
2022-02-28 17:16:44,778 INFO a info message
2022-02-28 17:16:44,778 WARNING a warning message
```

Another way would be to stick with your definitions and use:
```
#!/usr/bin/env python3

import logging
import argparse

parser = argparse.ArgumentParser(
     description=(
         'Read aircraft (AMDAR/TAMDAR/ACAR/etc.) BUFR file and convert into IODA output file'))

optional = parser.add_argument_group(title='optional arguments')
optional.add_argument('--debug', action=argparse.BooleanOptionalAction,
                      help='enable debug messages', default = False)
optional.add_argument('--verbose', action=argparse.BooleanOptionalAction,
                      help='enable verbose debug messages', default = False)

args = parser.parse_args()

if args.debug:
    logging.basicConfig(level=logging.DEBUG)
elif args.verbose:
    logging.basicConfig(level=logging.INFO)
# This is the default
#else:
#    logging.basicConfig(level=logging.WARNING)

logging.debug("a debug message")
logging.info("a info message")
logging.warning("a warning message")
```
which gives
```
JCSDA-L-18146:Desktop heinzell$ python3 logging_test2.py
WARNING:root:a warning message
JCSDA-L-18146:Desktop heinzell$ python3 logging_test2.py --verbose
INFO:root:a info message
WARNING:root:a warning message
JCSDA-L-18146:Desktop heinzell$ python3 logging_test2.py --verbose --debug
DEBUG:root:a debug message
INFO:root:a info message
WARNING:root:a warning message
```

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,506 @@
...
+import dateutil.parser
+import os
+from pathlib import Path
+import sys
+import time
+import logging
+
+import numpy as np
+import netCDF4 as nc
+from eccodes import *
```
CoryMartin-NOAA, line 13: this (`import *`) is usually very frowned upon because then you don't know where the functions/classes/etc. are coming from.

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,506 @@
...
+    data['specific_humidity'] = []
+    data['air_temperature'] = []
+    for key in meta_keys:
+        data[key] = []
+
+    varDict = defaultdict(lambda: DefaultOrderedDict(dict))
+    varAttrs = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
+
+    for fname in file_names:
+        logging.debug("INFO: Reading file: " + fname)
```
CoryMartin-NOAA, line 128: Is it necessary to include `INFO`? I thought the `logging` module would handle that depending on which stream is called (debug, info, etc.)?

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,506 @@
...
+        varAttrs[iodavar, qcName]['coordinates'] = 'longitude latitude'
+        varAttrs[iodavar, obsValName]['units'] = obsvars_units[n]
+        varAttrs[iodavar, obsErrName]['units'] = obsvars_units[n]
+        varAttrs[iodavar, qcName]['units'] = 'unitless'
+        n += 1
+
+    # Set units of the MetaData variables and all _FillValues.
+    for key in meta_keys:
+        varAttrs[(key, metaDataName)]['units'] = locationKeyList[meta_keys.index(key)][2]
+        if (locationKeyList[meta_keys.index(key)][1] == "string"):
```
CoryMartin-NOAA, line 162: Suggest using two dictionaries to get the fill value and dtype rather than a bunch of `elif`:
something like
```
missing_vals = {'string' : string_missing_value}
dtypes = {'string': object}
varAttrs[(key, metaDataName)]['_FillValue'] = missing_vals['string']
obs_data[(key, metaDataName)] = np.array(data[key], dtype=dtypes['string'])
``` 
(also what is `object`?)

CoryMartin-NOAA (commented):
Half reviewed with some comments, but had to leave for meeting.

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,506 @@
...
+    # Force longitude into space of -180 to +180 only. Reset to missing if either lat or lon not on earth
+    meta_data['longitude'][np.logical_or(meta_data['longitude']<-180.0, meta_data['longitude']>360.0)] = float_missing_value
+    meta_data['latitude'][np.logical_or(meta_data['latitude']<-90.0, meta_data['latitude']>90.0)] = float_missing_value
+    meta_data['latitude'][np.logical_or(meta_data['longitude']<-180.0, meta_data['longitude']>180.0)] = float_missing_value
+    meta_data['longitude'][np.logical_or(meta_data['latitude']<-90.0, meta_data['latitude']>90.0)] = float_missing_value
+    n = 0
+    for longitude in meta_data['longitude']:
+        if (meta_data['longitude'][n] != float_missing_value and meta_data['longitude'][n] > 360):
+            meta_data['longitude'][n] = 360.0 - meta_data['longitude'][n]
+        n+=1
```
fabiolrdiniz, line 409:
```suggestion
    for n, longitude in enumerate(meta_data['longitude']):
        if (meta_data['longitude'][n] != float_missing_value and meta_data['longitude'][n] > 360):
            meta_data['longitude'][n] = 360.0 - meta_data['longitude'][n]
```

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,506 @@
...
+        if (wind_direction != float_missing_value and vals['windSpeed'][n] != float_missing_value):
+            uwnd[n], vwnd[n] = met_utils.dir_speed_2_uv(wind_direction, vals['windSpeed'][n])
+        n+=1
+
+    spfh = np.full(len(year), float_missing_value)
+    n = 0
+    for mixing_ratio in vals['mixingRatio']:
+        if (mixing_ratio>0 and mixing_ratio<25.E-3):
+            spfh[n] = mixing_ratio / (1.0 + mixing_ratio)
+        n+=1
```
fabiolrdiniz, line 453:
```suggestion
    for n, mixing_ratio in enumerate(vals['mixingRatio']):
        if (mixing_ratio>0 and mixing_ratio<25.E-3):
            spfh[n] = mixing_ratio / (1.0 + mixing_ratio)
```

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,506 @@
...
+                          action='store', default=None, required=True,
+                          help='output file')
+
+    optional = parser.add_argument_group(title='optional arguments')
+    optional.add_argument('--debug', action=argparse.BooleanOptionalAction,
+                          help='enable debug messages')
+    optional.add_argument('--verbose', action=argparse.BooleanOptionalAction,
+                          help='enable verbose debug messages')
+
+    args = parser.parse_args()
```
fabiolrdiniz, line 492:
```suggestion
    args = parser.parse_args()
    args.output_file = os.path.abspath(args.output_file)
```
I suggest this because IODA will complain if one is running to generate a file locally (not specifying a full path, e.g., path+filename). For example, running like this will crash:
```
amdar_bufr2ioda.py -i input.bufr -o output.nc4
```
with the following message:
```
OSError: output.nc4 does not specify a valid path to a file
```
While a `./output.nc4` would need to be passed instead. With the above suggestion that wouldn't be needed.

fabiolrdiniz (commented):
Hi, @gthompsnJCSDA. Thanks for adding me here. A few suggestions...

src/conventional/amdar_bufr2ioda.py
srherbener: @fabiolrdiniz, FYI the issue with the output file specification has been fixed in: https://github.com/JCSDA-internal/ioda/pull/601.

src/conventional/amdar_bufr2ioda.py
gthompsnJCSDA: Should this be solely ```import eccodes``` with nothing after it?  I based mine on Ben Ruston's python script for raob.

src/conventional/amdar_bufr2ioda.py
fabiolrdiniz: Thank you very much, @srherbener! Sorry, I wasn't aware. @gthompsnJCSDA, I'll mark it as resolved.

src/conventional/amdar_bufr2ioda.py
gthompsnJCSDA: I must be dense, because I am not seeing how your proposed solution provides the right missing and type for ints or floats separately without still using a if-elif block.

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,508 @@
...
+
+    if args.debug:
+        logging.basicConfig(level=logging.INFO)
+    elif args.verbose:
+        logging.basicConfig(level=logging.DEBUG)
+    else:
+        logging.basicConfig(level=logging.ERROR)
```
fabiolrdiniz, line 501: @gthompsnJCSDA, I believe @climbfuji's idea was to set like this (with default printing only warnings, and `--debug` & `--verbose` triggering verbose and info printing, respectively).

```suggestion
    optional.add_argument('--debug', action=argparse.BooleanOptionalAction,
                          help='enable debug messages', default=False)
    optional.add_argument('--verbose', action=argparse.BooleanOptionalAction,
                          help='enable verbose debug messages', default=False)

    args = parser.parse_args()

    if args.debug:
        logging.basicConfig(level=logging.DEBUG)
    elif args.verbose:
        logging.basicConfig(level=logging.INFO)
    else:
        logging.basicConfig(level=logging.WARNING)
```

src/conventional/amdar_bufr2ioda.py
CoryMartin-NOAA: You can change lines 162-176 to:
```
dtypestr = locationKeyList[meta_keys.index(key)][1]
varAttrs[(key, metaDataName)]['_FillValue'] = missing_vals[dtypestr]
obs_data[(key, metaDataName)] = np.array(data[key], dtype=dtypes[dtypestr])
```

And then at like line 97 do something like:
```
missing_vals = {'float': nc.default_fillvals['f4'], 'string': '_'}
```

Does that make sense?

src/conventional/amdar_bufr2ioda.py
CoryMartin-NOAA: what actually are you using from `eccodes`? The problem with `import *` is it just makes everything available in this space and can overwrite things. After doing this, then you need to change whatever is used in eccodes to be `eccodes.thing` later in this script

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,508 @@
...
+            obs_data[(key, metaDataName)] = np.array(data[key], dtype=np.int64)
+        elif (locationKeyList[meta_keys.index(key)][1] == "float"):
+            varAttrs[(key, metaDataName)]['_FillValue'] = float_missing_value
+            obs_data[(key, metaDataName)] = np.array(data[key], dtype=np.float32)
+        elif (locationKeyList[meta_keys.index(key)][1] == "double"):
+            varAttrs[(key, metaDataName)]['_FillValue'] = double_missing_value
+            obs_data[(key, metaDataName)] = np.array(data[key], dtype=np.float64)
+
+    # Transfer from the 1-D data vectors and ensure output data (obs_data) types using numpy.
+    iodavar = 'air_temperature'
```
CoryMartin-NOAA, line 179:
```
errlist = [1.2, 0.75E-3, 1.7, 1.7]
for n, iodavar in ['air_temperature', 'specific_humidity', 'eastward_wind', 'northward_wind']:
    obs_data[(iodavar, obsValName)] = np.array(data[iodavar], dtype=np.float32)
    obs_data[(iodavar, obsErrName)] = np.full(nlocs, errlist[n], dtype=np.float32)
    obs_data[(iodavar, qcName)] = np.full(nlocs, 2, dtype=np.int32)

Comment (CoryMartin-NOAA) [reply src/conventional/amdar_bufr2ioda.py:179]: This above to replace 179-194.

File: src/conventional/amdar_bufr2ioda.py [hunk 10]
```diff
@@ -0,0 +1,508 @@
...
+    writer = iconv.IodaWriter(output_file, locationKeyList, DimDict)
+
+    # write everything out
+    writer.BuildIoda(obs_data, VarDims, varAttrs, AttrData)
+
+    logging.info("--- {:9.4f} total seconds ---".format(time.time() - start_time))
+
+
+# Replace BUFR missing value indicators with IODA missings.
+def assign_values(data, key):
```
CoryMartin-NOAA, line 208: similar suggestion to above to reduce ifs here

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,508 @@
...
+            data[np.abs(data) >= np.abs(bufr_missing_value)] = int_missing_value
+            return np.array(data, dtype=np.int32)
+        elif data.dtype.kind in {'U', 'S'}:
+            #data[data == ""] = string_missing_value
+            return np.array(data, dtype=object)
+    logging.critical("ABORT, no matching datatype found for key: " + key)
+
+
+# Assign the correct IODA missing value to MetaData element.
+def assign_missing_meta(key, mimic):
```
CoryMartin-NOAA, line 246: Same/similar suggestion again.

src/conventional/amdar_bufr2ioda.py
fabiolrdiniz: I believe that @CoryMartin-NOAA is suggesting to use something like this:
```suggestion
from eccodes import CODES_MISSING_LONG, codes_bufr_new_from_file, codes_set, codes_get_array, KeyValueNotFoundError, codes_release
```

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,508 @@
...
+    call_fail = False
+    if start_pos == f.tell():
+        return data, count, None
+    start_pos = f.tell()
+    count[0] += 1
+
+    met_utils = meteo_utils.meteo_utils()
+
+    try:
+        bufr = codes_bufr_new_from_file(f)
```
CoryMartin-NOAA, line 307: I assume this is one example where this *should* be `eccodes.codes_bufr_new_from_file(f)`.

test/CMakeLists.txt
```diff
@@ -549,6 +550,21 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_metar
                           -d 2020100106"
                           2020100106_metars_small.nc ${IODA_CONV_COMP_TOL_ZERO})
 
+#===============================================================================
+# Conventional data, aircraft Obs - AMDAR BUFR (WMO) converter
+#===============================================================================
+
+ecbuild_add_test( TARGET  test_${PROJECT_NAME}_amdar
```
CoryMartin-NOAA, line 557: This needs to be *optional* because of the eccodes dependence.

CoryMartin-NOAA (commented):
More comments/suggestions. Only hard requirement is the `eccodes` dependence means this test *must* be optional IMO.

src/conventional/amdar_bufr2ioda.py
gthompsnJCSDA: Yup, of course, gotcha and Thanks!

src/conventional/amdar_bufr2ioda.py
gthompsnJCSDA: In this particular method, I must use the IF-testing for data type, because the outgoing data type can sometimes change from the input type.  The example use-case I found is the "height" variable in aircraft-AMDAR data is coming into BUFR as an integer, but we must have it as a float for use in UFO/JEDI.  Otherwise, I followed your advice in the other places.

test/CMakeLists.txt
```diff
@@ -1062,31 +1063,50 @@ ecbuild_add_test( TARGET  test_iodaconv_atms_nc4
                           -i testinput/SNDR_SNPP_ATMS_SAMPLE.nc4
                           -o testrun/2021120600_atms_sdr.nc4" 2021120600_atms_sdr.nc4 ${IODA_CONV_COMP_TOL_ZERO} )
 
+if( iodaconv_eccodes_ENABLED )
```
gthompsnJCSDA, line 1066: @rhoneyager   I just added this ```iodaconv_eccodes_ENABLED``` since we should probably protect against the usage of scripts contributed by Ben Ruston and myself until eccodes can be installed on as many platforms as possible.  Meanwhile, I don't know how to modify the cmake codes to create this new variable.  So I am seeking  your help to enable this macro please.

gthompsnJCSDA:
Message to previous reviewers:  THANKS for all the terrific suggestions.  The code is now modified basically how each person advised and therefore greatly improved.  Please re-review when you can spare some time.

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,514 @@
+#!/usr/bin/python
```
CoryMartin-NOAA, line 1: Aren't these supposed to all be `python3` now?

CoryMartin-NOAA:
```suggestion
#!/usr/bin/env python3
```

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,514 @@
...
+import dateutil.parser
+import os
+from pathlib import Path
+import sys
+import time
+import logging
+
+import numpy as np
+import netCDF4 as nc
+from eccodes import CODES_MISSING_LONG, codes_bufr_new_from_file, codes_set, codes_get_array, KeyValueNotFoundError, codes_release
```
CoryMartin-NOAA, line 13: is this line not too long for coding norms? Alternative to this approach is to do:
`import eccodes as ecc` and then each of these calls can be `ecc.CODES_MISSING_LONG` for example.

CoryMartin-NOAA (changes requested):
Small thing, and then I think it is good to go.

gthompsnJCSDA:
@CoryMartin-NOAA   Incorporated your suggestions.  I still need to handle the coding_norms check.

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
good to merge once coding norms are dealt with

test/CMakeLists.txt
rhoneyager: The eccodes find_package script should report whether it is available of not. Checking that variable should be fine. It's something like ```eccodes_FOUND```.

```find_package(eccodes .......``` should also be added in the top-level CMakeLists.txt file.

Comment (rhoneyager) [reply test/CMakeLists.txt:1066]: Pinging @climbfuji here since we just added eccodes Python in addition to eccodes C, and we need to develop a find_package script in jedi-cmake for this.

Comment (climbfuji) [reply test/CMakeLists.txt:1066]: Duly notes, thanks for the heads up.

**Review by climbfuji — APPROVED** [review]
I only want to confirm that the logging.warning statements are intentional, because this log level isn't used at all. You have error/info/debug, which means that the warning messages will only show up for info/debug, just with a "WARNING" prefix instead of an "INFO" prefix. If that is the intention, then I am ok with it.

File: test/CMakeLists.txt (hunk 14 above)
Comment (srherbener) [reply test/CMakeLists.txt:1066]: @gthompsnJCSDA You could do something similar to what is done for pygrib in the top level CMakeLists.txt file. The following suggestion is code to be placed in the top level CMakeLists.txt file.

```
option(USE_ECCODES "Use Python eccodes package" ON)
if ( USE_ECCODES )
    message(STATUS "Adding python tests that import eccodes")
    set( iodaconv_eccodes_ENABLED True)
else()
    message(STATUS "Skipping python tests that import eccodes")
    set( iodaconv_eccodes_ENABLED False)
endif()
```

What this will do is set USE_ECCODES to ON by default. If you need to shut these off, then you add `-DUSE_ECCODES=OFF` to the ecbuild command line.

**Review by srherbener — COMMENTED** [review]
Looks good. Thanks for adding this!

File: test/CMakeLists.txt (hunk 14 above)
Comment (srherbener) [reply test/CMakeLists.txt:1066]: Sorry, when writing my review I didn't see @rhoneyager's response. @rhoneyager has a good point about using find_package. I had forgotten that eccodes comes in two pieces, a compiled library (of which the find_package is needed to check for its existence) and a python package (of which the option is needed to get the hint from the user). I'm not sure if there is a way to automatically detect the existence of the python eccodes package. Perhaps cmake has a way to do that.

For now you could add the find package and change the if statement to something like:
```
find_package( eccodes QUIET )
option(USE_ECCODES "Use Python eccodes package" ON)
if ( eccodes_FOUND AND USE_ECCODES )
...
```

Comment (srherbener) [reply test/CMakeLists.txt:1066]: After today's EMC IODA Hour discussion, I think we should make the default for USE_ECCODES be OFF instead of ON. Not all operational centers will use eccodes, and therefore eccodes should be an optional part of jedi-stack (as well as the eccodes python package being optional).

**Comment by gthompsnJCSDA** [comment]
> 

This was done intentionally.  I might change the logging levels at a later time, but currently it's how I wanted to do things.

File: CMakeLists.txt [hunk 17]
```diff
@@ -53,6 +53,22 @@ else()
...
+        set( iodaconv_eccodes_ENABLED True)
+    else()
+        message(STATUS "Not-found: eccoodes - eccodes disabled")
+        set( iodaconv_eccodes_ENABLED False)
+    endif()
+else()
+    message(STATUS "Skipping python tests that import eccodes")
+    set( iodaconv_eccodes_ENABLED False)
+endif()
+
```
srherbener, line 71: This section looks great! Thanks for adding it in. I tried it on my Mac and it works just fine.

test/CMakeLists.txt
```diff
@@ -1062,31 +1063,50 @@ ecbuild_add_test( TARGET  test_iodaconv_atms_nc4
...
+                           "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/decode_bufr_LDM_raob.py
+                           -i testinput/sonde_wmo_double.bufr testinput/sonde_wmo_double.bufr
+                           -o testrun/wmo_raob_quadruple.nc4"
+                           wmo_raob_quadruple.nc4 ${IODA_CONV_COMP_TOL_ZERO})
+
+#===============================================================================
+# Conventional data, aircraft Obs - AMDAR BUFR (WMO) converter
+#===============================================================================
+
+  ecbuild_add_test( TARGET  test_${PROJECT_NAME}_amdar
```
srherbener, line 1098: CI testing defaults to `USE_ECCODES=OFF` so this test is not getting run. I tried `USE_ECCODES=ON` on my mac and I get this error:
```
Start testing: Mar 03 13:53 MST
----------------------------------------------------------
591/695 Testing: test_iodaconv_amdar
591/695 Test: test_iodaconv_amdar
Command: "/usr/local/bin/bash" "/Users/steveherbener/projects/DEVELOP/ioda-bundle/build/bin/iodaconv_comp.sh" "netcdf" "/usr/local/Frameworks/Python.framework/Versions/3.9/bin/python3.9 /Users/steveherbener/projects/DEVELOP/ioda-bundle/build/bin/amdar_bufr2ioda.py
                            -i testinput/amdar_wmo_multi.bufr
                            -o testrun/amdar_wmo_multi2.nc" "amdar_wmo_multi2.nc" "0.0"
Directory: /Users/steveherbener/projects/DEVELOP/ioda-bundle/build/iodaconv/test
"test_iodaconv_amdar" start time: Mar 03 13:53 MST
Output:
----------------------------------------------------------
DIFFER : VALUES OF GLOBAL ATTRIBUTE : source_files : "testinput/amdar_wmo_multi.bufr" <> "/work/noaa/da/gthompsn/src/ioda-bundle/iodaconv/test/testinput/amdar_wmo_multi.bufr"
<end of output>
Test time =   1.02 sec
----------------------------------------------------------
Test Failed.
"test_iodaconv_amdar" end time: Mar 03 13:53 MST
"test_iodaconv_amdar" time elapsed: 00:00:01
----------------------------------------------------------

End testing: Mar 03 13:53 MST
```

The global attribute "source_files" doesn't match and I think the fix is to update the test reference file.

srherbener (changes requested):
All the code looks great, and thanks for adding in the CMake configuration for eccodes. I'm requesting changes due to a test failure I saw on my Mac. Once that's addressed, it's good to go. Thanks!

srherbener:
@climbfuji, @mer-a-o after some discussions today it seems that eccodes should be treated as an optional jedi-stack library and python package (ie, some centers want to use it, others don't). 

This PR contains the cmake configuration to handle eccodes as an optional feature, which is disabled by default. To enable you need to have the eccodes library installed and find-able by cmake, plus have the eccodes python package installed, plus add `-DUSE_ECCODES=ON` to the ecbuild command line.

I believe that the eccodes library and python package are available in the test containers already, and all that is left is to add `-DUSE_ECCODES=ON` to the ecbuild. Could you help with getting this added? Thanks!

gthompsnJCSDA:
Sorry for this rant, but this coding norms check is so annoying.  The following lines of code fail with W503 warning even when I attempt to bypass with a ```# noqa``` (no quality assurance) bypass.

```
            logging.warning("Key called " + k + " contains only " + str(len(meta_data[k])) +
                            " elements, whereas " + str(target_number) + " were expected.")  # noqa
```

Whether I put the + sign at end of 1st line or begin on 2nd line, Pycodestyle complains. What a nuisance.  So, am I supposed to waste the time of putting into a dumb temporary variable the pieces of the string that I want to build just to get rid of a useless warning message?  Does anyone else have an idea why the noqa comment is not causing the warning to be bypassed?  (and, I have also tried it on the end of both lines besides what is shown above).

src/conventional/amdar_bufr2ioda.py
```diff
@@ -0,0 +1,529 @@
...
+
+    # For any of the MetaData elements that were totally lacking, fill entire vector with missing.
+    empty = []
+    for k, v in metaDataKeyList.items():
+        if not any(meta_data[k]):
+            meta_data[k] = assign_missing_meta(empty, k, target_number, 0)
+        if (len(meta_data[k]) == 1):
+            meta_data[k] = np.full(target_number, meta_data[k][0])
+        elif (len(meta_data[k]) < target_number):
+            logging.warning("Key called " + k + " contains only " + str(len(meta_data[k])) +
```
CoryMartin-NOAA, line 345:
```suggestion
            logging.warning(f"Key called {k} contains only {len(meta_data[k])} "
                                        f" elements, wheras {target_number} were expected.")
```

CoryMartin-NOAA: above should work?

src/conventional/amdar_bufr2ioda.py
```diff
@@ -440,9 +440,8 @@ def read_bufr_message(f, count, start_pos, data):
         try:
             avals = ecc.codes_get_array(bufr, variable)
             if (len(avals) != target_number):
-                logging.warning("Caution, for " + variable + "a length mismatch exists: " +
-                                str(len(avals)) + " found, when expecting " +
-                                str(target_number) + ".  Skpping this BUFR msg.")            # noqa
+                logging.warning(f"Variable called {k} contains only {len(avals)} "
```
CoryMartin-NOAA, line 443: shouldn't `k` be `variable` here?

src/conventional/amdar_bufr2ioda.py
gthompsnJCSDA: Very good suggestion, thanks @CoryMartin-NOAA.  I find it really weird syntax with this f character to precede something in double-quotes and so forth.

CoryMartin-NOAA: https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings formatted string literals

gthompsnJCSDA:
@srherbener   I made a comment in the top section under ```Dependencies``` that I request your assistance.  Otherwise, I finally have this code sorted out for all tests and reviews are all good now.

srherbener:
> @srherbener I issued a PR and added `-DUSE_ECCODES=ON` to CI: #810

Thanks @mer-a-o for fixing this!

srherbener:
> @srherbener I made a comment in the top section under `Dependencies` that I request your assistance. Otherwise, I finally have this code sorted out for all tests and reviews are all good now.

@gthompsnJCSDA I think the CMake configuration is working now. Is there something else you need help with? Thanks

srherbener (approved):
Thanks for taking care of the unexpected issues that came up. Your patience is greatly appreciated.

PR #812, opened 2022-03-04
Python script to convert WMO-formatted BUFR messages of SYNOPs (surface obs) to IODA

gthompsnJCSDA:
## Description

This PR allows SYNOP surface observations direct off NCAR-RAL's LDM feed of live data to be converted to IODA for use in UFO/JEDI. This is an ```eccodes``` alternative to using bufr2ioda.x application (Ron McLaren's) because it also does conversion of winds from speed/direction into eastward_wind and northward_wind for immediate usage while waiting for larger changes to UFO for variable transforms. This PR also converts from dewpoint temperature to specific_humidity variable needed in JEDI. Simple constant assumptions are made for starting ObsError values that can easily be overridden or replaced by UFO filters (i.e., Perform Action and assign error found in ufo/ewok/skylab/synop.yaml) as well as setting the PreQC value to 2, which EMC uses for default, not yet quality controlled.

Much of the code here overlaps what is found in the ```amdar_bufr2ioda.py``` script except the needed variable changes that differ between the two.

### Issue(s) addressed

A specific issue was not yet created

Two or more positive reviews plus working ctests and fully useful output for next steps in UFO/JEDI

## Dependencies

Waiting on the following PRs:
- [x] waiting on #807 

## Impact

No impacts to other repos expected.


CoryMartin-NOAA (approved):

gthompsnJCSDA:
> I have one question around these scripts. I see that there are three of these right now, SYNOPS, SHIP, BUOY. And it seems that there is a substantial amount of code duplication between them. Would it make sense to create a common module to share between them? Doesn't have to be right now, I think it's ok to merge all three PRs and do this as a follow-up. If this is something we should do, then there should be an issue so that we don't forget.


Yes, I do have a lot of overlap and had considered how to modularize.  Something for the TO-DO list, agreed.

climbfuji:
> > I have one question around these scripts. I see that there are three of these right now, SYNOPS, SHIP, BUOY. And it seems that there is a substantial amount of code duplication between them. Would it make sense to create a common module to share between them? Doesn't have to be right now, I think it's ok to merge all three PRs and do this as a follow-up. If this is something we should do, then there should be an issue so that we don't forget.
> 
> Yes, I do have a lot of overlap and had considered how to modularize. Something for the TO-DO list, agreed.

I looked at the repository and there are many more Python scripts. I think this consolidation should be part of a larger effort to reduce code duplication in the many Python scripts.

BenjaminRuston:
Yes agree as well @climbfuji ,, I'm somewhat to blame as well so work with @gthompsnJCSDA to consolidate these.

climbfuji:
I learned that @srherbener had started an effort in this direction a few years ago, I think it would be worthwhile considering what was done back then.

PR #834, opened 2022-03-09
Feature/viirs_bias :  First attempt at bias file writer for VIIRS AOD observations. 

andytangborn:
## Description

This is a draft PR for a VIIRS bias writer.   It is a python code that writes a netcdf file with just a single global bias value for VIIRS AOD (and it's uncertainty).  The resulting NetCDF file has been successfully tested with hybrid DA in JEDI. 

But it is not clear if it is in the right directory. It's currently in iodaconv/compo because this is where all the python NetCDF IODA writers are created, but it probably needs to be in iodaconv/satbias.   And while the ctest for this converter passes successfully, it fails in the validator step. This is likely because the validator is looking for items that are needed for IODA obs files, rather than satbias files.   

### Issue(s) addressed

This is connected to the issue #1889, but this issue should remain open after this PR is closed so  that it can be tied to a new PR. 

What does it mean for this PR to be finished?  This PR will be closed and another issued once the details of where the code should be located, and how the 

## Dependencies

None

### DIFF
--- a/src/compo/CMakeLists.txt
+++ b/src/compo/CMakeLists.txt
@@ -6,6 +6,7 @@ list(APPEND programs
   modis_aod2ioda.py
   aeronet_aod2ioda.py
   aeronet_aaod2ioda.py
+  VIIRS_BiasWriter.py
 )
 
 file( RELATIVE_PATH SCRIPT_LIB_PATH ${CMAKE_BINARY_DIR}/bin ${PYIODACONV_BUILD_LIBDIR} )
--- a/src/compo/VIIRS_BiasWriter.py
+++ b/src/compo/VIIRS_BiasWriter.py
@@ -1,0 +1,46 @@
+import sys
+import argparse
+import netCDF4
+from netCDF4 import Dataset, chartostring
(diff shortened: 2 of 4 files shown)


CoryMartin-NOAA:
@andytangborn hmm, so does the validator run on the satbias files or not (I guess a question for @rhoneyager maybe)? It looks like the `glob` includes all files but maybe I'm missing something in the `CMakeLists.txt` file.

andytangborn:
My question is: Does the validator look for different items in satbias
files and ioda obs files?

andytangborn:
@CoryMartin-NOAA , @rhoneyager 
Looking at the CMakeLists.txt file in iodaconv/test, I find that the validator uses the yaml: ObsSpace.yaml in ioda-bundle/ioda/share/ioda/yaml/validation/ObsSpace.yaml

And this is accessed for all of the files in test/testoutput that end in *.nc and *.nc4.  So at first glance, it looks like this is run on all of the NetCDF files, which would include both Obs files and satbias files.   But I could be wrong.

PR #856, opened 2022-03-10
Sprint: Upgrade converters that use bufr2ioda.x to follow the JEDI Data Conventions

### LINKED ISSUE #852: Sprint:  Upgrade converters that use bufr2ioda.x to follow the JEDI Data Conventions
The list of yaml below is a partial list of those using ```bufr2ioda.x``` application to convert BUFR to IODA of various conventional data.  Each of these need to be update to follow JEDI data conventions.  Also a ```units``` attribute was considered required in the earlier version of ```budfr2ioda.x``` but it must now be optional since a variable with no units at all should simply skip this attribute, so very small changes have been made to ```src/bufr/IodaEncoder/IodaEncoder.cpp```

Starting list of simple converters (another issue will be created for those not included here):

- aircar_BUFR2ioda.yaml
- airep_wmoBUFR2ioda.yaml
- amdar_wmoBUFR2ioda.yaml
- buoy_wmoBUFR2ioda.yaml
- gnssro_wmoBUFR2ioda.yaml
- rass_wmoBUFR2ioda.yaml
- satwind_EUMet_wmoBUFR2ioda.yaml
- satwind_Himawari_wmoBUFR2ioda.yaml
- satwind_Insat_wmoBUFR2ioda.yaml
- ship_wmoBUFR2ioda.yaml
- synop_wmoBUFR2ioda.yaml
- vadwinds_wmoBUFR2ioda.yaml

## Notifications

@rhoneyager @rmclaren

gthompsnJCSDA:
## Description

This PR updates a batch of 11 yaml files (and testoutput reference files) to use the new JEDI Data Conventions; all of which are using the bufr2ioda.x application for converting BUFR data to IODA.  For the list of yaml files changed, please see the issue number below.

### Issue(s) addressed

closes #852 

Reviews and working ctests

## Dependencies

Waiting on the following PRs:
- [x] waiting on JCSDA-internal/ioda/pull/638

## Impact

Within ```feature/sprint-ioda-converters``` there should be no impacts to other repos.


gthompsnJCSDA:
@rmclaren   Just tagging Ron McLaren here, because I modified IodaEncoder.cpp to permit the usage of a string of ```none``` or an empty double-quote string as the ```units``` because we want some variables in IODA to **SKIP** the units attribute in the case that units do not make sense.  So I wanted to clear my technique with you on this solution.

test/testinput/buoy_wmoBUFR2ioda.yaml
```diff
@@ -37,117 +37,117 @@ observations:
...
           source: variables/stationElevation
-          dimensions: [ "nlocs" ]
+          dimensions: [ "Location" ]
           longName: "Station Elevation"
           units: "m"
 
         - name: "MetaData/height"
           source: variables/stationElevation
-          dimensions: [ "nlocs" ]
+          dimensions: [ "Location" ]
```
CoryMartin-NOAA, line 102: This and the previous variable are identical? Is that correct? I guess so since a buoy should be `0`?

test/testinput/gnssro_wmoBUFR2ioda.yaml
```diff
@@ -48,170 +48,151 @@ observations:
...
-          dimensions: [ "nlocs" ]
+          dimensions: [ "Location" ]
           longName: "Geopotential height"
           units: "m"
 
-        - name: "MetaData/airPressure"
-          coordinates: "longitude latitude"
-          source: variables/airPressure
-          dimensions: [ "nlocs" ]
+        - name: "MetaData/pressure"
```
CoryMartin-NOAA, line 112: I know this is not really part of the PR, but why is this changed? I don't like this. What about `water pressure` or `vapor pressure`. `pressure` is too ambiguous IMO.

CoryMartin-NOAA (approved):
Mostly looks good, just a couple comments/questions.

test/testinput/buoy_wmoBUFR2ioda.yaml
gthompsnJCSDA: @CoryMartin-NOAA   Agree this is a total HACK, but there is no choice really.  All of the treatment of conventional data for pressure (and probably temp/wind/moisture) require a station elevation and height of ob separately.  Buoy data rarely reports both.  This hack provides a way to fill the height variable which should generally be a meter or so higher than so-called station elevation.  This is a kinda problem with Ron's converter in general.  If the data are lacking in the first place, then a straight bufr-to-ioda is often useless.  Very frustrating that BUFR allows for rich content, but countries making those data have to fill things in (and they don't do enough of it).

test/testinput/gnssro_wmoBUFR2ioda.yaml
gthompsnJCSDA: > I know this is not really part of the PR, but why is this changed? I don't like this. What about `water pressure` or `vapor pressure`. `pressure` is too ambiguous IMO.


I could not agree more!  This is a definite issue/problem, but when I reviewed the choices in WMO bufr tables, there really isn't anything separated for ```pressure```.  I debated just diverting from BUFR tables, but didn't.  We can still diverge if we want.  But, at this point, I would need to change another 10-15 files (again).

CoryMartin-NOAA: maybe something to discuss today with Ryan/the group? I'm fine with leaving as is for now, but really think airPressure is much better than pressure, especially since the long name is still 'air pressure'

src/bufr/IodaEncoder/IodaEncoder.cpp
```diff
@@ -155,7 +155,11 @@ namespace Ingester
 
 
                 var.atts.add<std::string>("long_name", { varDesc.longName }, {1});
-                var.atts.add<std::string>("units", { varDesc.units }, {1});
+
+                if (!(varDesc.units == "none" || varDesc.units == ""))
```
srherbener, line 159: @rmclaren just want you to be aware of this change that @gthompsnJCSDA needs for this PR. For now this change will remain in the feature/sprint-ioda-converters branch which may not get merged into develop for a couple months. If you are okay with this change it might be worthwhile to add it to your feature branch (feature/query ?) and develop.

srherbener (approved):
Looks good. My one comment is there to help keep track of the update you made to the bufr converter.

PR #887, opened 2022-03-24
SPRINT:  upgrade radiosonde python-eccodes bufr (WMO) converter to JEDI Data Conventions

### LINKED ISSUE #876: Upgrade converter for (decode_bufr_LDM_raob.py) to follow the JEDI Data Conventions
## Description

In addition to updating the existing converter for JEDI Data Conventions, also changing its name to be more consistent with others in the ```conventional``` subdirectory to become ```sonde_bufr2ioda.py```.  Also making numerous updates to streamline the code to be more similar to the others.

## Notifications

@rhoneyager and @BenjaminRuston

gthompsnJCSDA:
## Description

Update the python-eccodes radiosonde bufr-to-ioda converter to align with JEDI Data Conventions.  This is also a nearly complete rewrite of the previous code named ```decode_bufr_LDM_raob.py``` to make it more flexible.  All of the specialized BUFR mnemonics being captured and IODA variable names are found near the very top - easily modified if later additions are needed.  There are also 2 documented files in the ```doc/bufr_obs_samples``` and ```doc/bufr_table_samples``` to find other elements encoded in the BUFR.

### Issue(s) addressed

Closes #876 

Passes tests and positive reviews

## Dependencies

Depends on
- [ ] [IODA 667](https://github.com/JCSDA-internal/ioda/pull/667)

## Impact

No impacts to other repos provided this remains part of ```feature/sprint-ioda-converters```


rhoneyager:
From AWS:
```
-- Enabled Component: pbfortran
439 | CMake Error: File /jcsda/ioda-bundle/iodaconv/src/conventional/decode_bufr_LDM_raob.py does not exist.
440 | CMake Error at iodaconv/cmake/iodaconv_extra_macros.cmake:32 (configure_file):
441 | configure_file Problem configuring file
442 | Call Stack (most recent call first):
443 | iodaconv/src/conventional/CMakeLists.txt:11 (conf_targets_deps)
444 |  
445

```

srherbener:
> From AWS:
> 
> ```
> 439 | CMake Error: File /jcsda/ioda-bundle/iodaconv/src/conventional/decode_bufr_LDM_raob.py does not exist.
> 440 | CMake Error at iodaconv/cmake/iodaconv_extra_macros.cmake:32 (configure_file):
> 441 | configure_file Problem configuring file
> 442 | Call Stack (most recent call first):
> 443 | iodaconv/src/conventional/CMakeLists.txt:11 (conf_targets_deps)
> ```
> 
> You deleted this file in this PR. The CMake code needs to be updated to reflect this.

I think this is the line that needs to be removed to fix this:

https://github.com/JCSDA-internal/ioda-converters/blob/63b067d813276b32fa69ce5f01d7fbaa7eebf59a/src/conventional/CMakeLists.txt#L5

src/conventional/sonde_bufr2ioda.py
```diff
@@ -0,0 +1,639 @@
...
+        second = vals[0]
+        if (second < 0 or second > 59):
+            second = 0
+    except ecc.KeyValueNotFoundError:
+        logging.info("Caution, no data for second")
+
+    if (year < 1900 or year>2499 or month < 1 or month > 12 or day < 1 or day > 31 or hour < 0 or hour > 23 or minute < 0 or minute > 59):
+        bad_date = True
+
+    if bad_date:
```
srherbener, line 373: Sorry if I'm not getting the logic right in this section of the code.

It looks like, when bad_data is False, time_offset is set to the number of seconds from the epoch to the first datetime value from the bufr file. Then all subsequent datetime values from the bufr file are adjusted to offsets from the epoch (via the time_offset value). This part looks good.

It seems a bit risky to silently "repair" bad datetime values (ie out of the range 0 to 6 hours). Would it be better to put in missing values?

When bad_data is False, it looks like to me all of the datetimes from the bufr file are getting translated to being offsets from Jan 1, 1900 0Z instead of the more current time you would likely have if the first datetime was valid.

Was this done to force the ioda reader to throw away all of the obs? If so, would it be better to just remove this bufr message from the output file altogether? Perhaps you could check the first datetime value at the start of the read_bufr_message function and skip the processing of the message if it is invalid.

src/conventional/sonde_bufr2ioda.py
```diff
@@ -0,0 +1,639 @@
...
+    mask_lon = np.logical_or(meta_data['longitude'] < -180.0, meta_data['longitude'] > 360.0)
+    meta_data['latitude'][mask_lat] = float_missing_value
+    meta_data['longitude'][mask_lon] = float_missing_value
+    meta_data['latitude'][mask_lon] = float_missing_value
+    meta_data['longitude'][mask_lat] = float_missing_value
+    for n, longitude in enumerate(meta_data['longitude']):
+        if (meta_data['longitude'][n] != float_missing_value and meta_data['longitude'][n] > 360):
+            meta_data['longitude'][n] = 360.0 - meta_data['longitude'][n]
+
+    # It is NOT IDEAL, but if missing a lat or lon, fill with prior known value for now.
```
srherbener, line 523: Would this be another case to skip this message or put in missing values? It seems a bit risky to silently "repair" bad lat, lon values.

srherbener (commented):
For the most part the code looks great. I've got a couple comments about the handling of invalid values for the datetime, lat, and lon data. Thanks!

src/conventional/sonde_bufr2ioda.py
gthompsnJCSDA: Actually this check is enforcing that the longitudes are not permitted to go from zero to 360, but must go per usual convention of -180 to +180 degrees.

However, the point you are making about lines below 523 is correct. I decided to auto-correct missing lat/lon values from the most recently known values due to the fact that the observations are entirely pointless without a lat/lon and the potential error from how far a radiosonde moves in 1 or 2 seconds (of balloon ascent) is pretty small.  In a single example file, I found a few sites that had infrequent spots of missing lat/lon that I felt was worthy of keeping location constant.  This is debatable to fix like this versus using a specialized function to interpolate or use wind info to get a better true estimate.  However, at this time, I would like to move forward as-is with possibility to write a specialized UFO function to fix it there.

src/conventional/sonde_bufr2ioda.py
gthompsnJCSDA: The point you're making is valid.  Here is the problem:  there are actually no quality checks at all in the creation of BUFR messages.  Quite sad but true.  So one sounding I found has a year of "22" instead of "2022."  It's actually ridiculous in this modern age how poorly quality controlled the *inputting* of data is in the first place.  Another example that is possible is specifying the 31st of February - there is almost nothing to prevent this insanity.  The logic behind my choice of setting a default bad date of 1900-01-01T00:00:00Z is simply that we do not have any such data back then yet the datetime object in Python will permit the calculation of time offset from epoch of that date.  Once in UFO, a year of 1900 will be rejected anyway.

My attempt here is to try and prevent the ioda-converter from failing given as many possible pathological inputs as possible - yet I do know that if someone codes a BUFR message with month=2 and day=31, the python datetime object will choke on it since I am not using anything like try-except.

src/conventional/sonde_bufr2ioda.py
```diff
@@ -0,0 +1,639 @@
...
+    'pressure': ['pressure', 'nonCoordinatePressure'],
+    'geopotentialHeight': ['nonCoordinateGeopotentialHeight', 'geopotentialHeight'],
+    'vertSignificance': ['extendedVerticalSoundingSignificance'],
+    'latDisplacement': ['latitudeDisplacement'],
+    'lonDisplacement': ['longitudeDisplacement'],
+    'timeDisplacement': ['timePeriod'],
+    'wmoBlockNumber': ['blockNumber'],
+    'wmoStationNumber': ['stationNumber'],
+    'stationWMO': ['Constructed'],
+#   "stationLongName": 'shipOrMobileLandStationIdentifier',
```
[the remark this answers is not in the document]
gthompsnJCSDA: Yes, I envisioned that in a not-so-distant update of the converter, some of these other useful pieces of info need to be conveyed into the IODA output file for later usage in UFO.  So they are here as reminders to myself.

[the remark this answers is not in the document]
srherbener: Sounds good. I'm okay with leaving as-is.

src/conventional/sonde_bufr2ioda.py
srherbener: Thanks for the explanation. I see the point you are making about the lat, lon not changing much between successive readings. There still is a possibility that you could have a long series of bogus lat, lon values resulting in larger errors, but that might be rare, and it seems that the potential function you mention could see that and do a somewhat reasonable job of remedying the situation.

I'm okay with deferring the debate about how/where to correct these values to after this PR.

src/conventional/sonde_bufr2ioda.py
```diff
@@ -0,0 +1,665 @@
...
+    # Force longitude into space of -180 to +180 only. Reset both lat/lon missing if either absent.
+    mask_lat = np.logical_or(meta_data['latitude'] < -90.0, meta_data['latitude'] > 90.0)
+    mask_lon = np.logical_or(meta_data['longitude'] < -180.0, meta_data['longitude'] > 360.0)
+    meta_data['latitude'][mask_lat] = float_missing_value
+    meta_data['longitude'][mask_lon] = float_missing_value
+    meta_data['latitude'][mask_lon] = float_missing_value
+    meta_data['longitude'][mask_lat] = float_missing_value
+    for n, longitude in enumerate(meta_data['longitude']):
+        if (meta_data['longitude'][n] != float_missing_value and meta_data['longitude'][n] > 360):
+            meta_data['longitude'][n] = 360.0 - meta_data['longitude'][n]
```
srherbener, line 541: My apologies if I'm misreading/misinterpreting.

It looks like the code on lines 534-538 is going to put in a missing value if the longitude is < -180 or > 360 (ie the longitude is out of range for either 0-360 or +/- 180 schemes).

If I've got that right, then Is the code on lines 540-541 meant to translate the 0-360 scheme to the +/- 180 scheme?

If so, then shouldn't the logic be like this:
```
# Test if longitude is greater than 180 and if so subtract 360 to shift to +/- 180 value
if (meta_data['longitude'][n] != float_missing_value and meta_data['longitude'][n] > 180.0):
            meta_data['longitude'][n] = meta_data['longitude'][n] - 360.0
```
Eg, if longitude is 270, then the current code leaves it as 270, and the suggested code would change it to -90; while for either code if the longitude is 90, then it remains 90.

gthompsnJCSDA: Thank you Steve for being thorough.  You are absolutely correct and I made an error here for certain.  Now correcting it.

src/conventional/sonde_bufr2ioda.py
srherbener: Thanks for the feedback @gthompsnJCSDA. I think what you are saying makes sense and sounds reasonable to me. I like the aspect of deferring the choice to "rescue" the data to UFO. Thanks for your patience with this.

srherbener (approved):
Looks great! Thank you for your thoughtful discussions with my comments.

rhoneyager:
There are several tests that now fail the reference check because the number of processed observations have changed. See test_iodaconv_bufr_filtering / http://cdash.new.jcsda.org:8080/test/199689 for an example.

gthompsnJCSDA:
> There are several tests that now fail the reference check because the number of processed observations have changed. See test_iodaconv_bufr_filtering / http://cdash.new.jcsda.org:8080/test/199689 for an example.

Ugh. I figured this could happen.  One of the testinput files I was using is also used for another ctest.  When I cut back the size of the file, this now changed an older ctest.  So I got some fixing to do on the ones that ended up changing from me cutting back the size of the input file.

CoryMartin-NOAA (approved):

PR #895, opened 2022-03-31
Sprint:  Update the GODAE (marine) converters listed below to adopt the JEDI Data Conventions

### LINKED ISSUE #890: Upgrade converters for Marine: GODAE to follow the JEDI Data Conventions
## Description

Update the python converters (listed below) for marine godae ship, track, profile, and argo float to follow the JEDI Data Conventions.  These converters are part of the ```src/marine``` directory.

- [x] godae_ship2ioda.py
- [x] godae_trak2ioda.py
- [x] godae_profile2ioda.py
- [x] godae_bgc_argo2ioda.py

gthompsnJCSDA:
## Description

This PR updates the python scripts for four GODAE marine converters to adopt the JEDI Data Conventions.  The four converters are:

godae_ship2ioda.py
godae_trak2ioda.py
godae_profile2ioda.py
godae_bgc_argo2ioda.py

In addition the test reference files are updated and should properly validate using ```ioda-validate.x```

### Issue(s) addressed

Closes #890 

Working ctests for conversion and matching test reference files and ioda-validate results plus reviews.

## Dependencies

I need to create a few modifications to ObsSpace.yaml in ioda repo (I will return and edit this PR info once I create the PR over there)
- [x] waiting on JCSDA-internal/ioda#667
- [x] waiting on JCSDA-internal/ioda#673

## Impact

No impacts to other repos provided this is kept in ```feature/ioda-sprint-converters```


gthompsnJCSDA:
Tagging @travissluka  here because I need to clarify the units of salinity and possibly any other variables in these four converters.  The original units were PSU (practical salinity units) which I hope are the same as g/kg, but I need to adjust if this is incorrect.  A Wikipedia page I found shows salinity in g/kg which I believe seems more logical to be some mass of salt divided by mass of salt water.

And, FYI, I greatly unified the look of these four converters to have the same basic structure throughout that includes names of variables to write to IODA files and their units and so forth.  I think the 4 are now going to be much easier to maintain.  Please have a look when time permits.

srherbener:
I'm on leave now until April 11th. You may want to remove/replace me as a reviewer. Thanks!

src/marine/godae_profile2ioda.py
```diff
@@ -160,72 +217,103 @@ def __init__(self, filename, date, varDict, varDims, obsList):
...
+
+        # Set units and FillValue attributes for groups associated with observed variable.
+        for key in varDict.keys():
+            value = varDict[key][0]
+            units = varDict[key][1]
+            varAttrs[(value, obsValName)]['units'] = units
+            varAttrs[(value, obsErrName)]['units'] = units
+            varAttrs[(value, obsValName)]['_FillValue'] = float_missing_value
+            varAttrs[(value, obsErrName)]['_FillValue'] = float_missing_value
+            varAttrs[(value, qcName)]['_FillValue'] = int_missing_value*100
```
CoryMartin-NOAA, line 247: Why not just change the missing value above?

src/marine/godae_trak2ioda.py
```diff
@@ -113,75 +169,61 @@ def __init__(self, filename, date, varDict, varDims, obsList):
...
-            self.varAttrs[value, iconv.OqcName()]['units'] = 'unitless'
+
+        # Set units and FillValue attributes for groups associated with observed variable.
+        for key in varDict.keys():
+            value = varDict[key]
+            self.varAttrs[(value, obsValName)]['units'] = unitDict[key]
+            self.varAttrs[(value, obsErrName)]['units'] = unitDict[key]
+            self.varAttrs[(value, obsValName)]['_FillValue'] = float_missing_value
+            self.varAttrs[(value, obsErrName)]['_FillValue'] = float_missing_value
+            self.varAttrs[(value, qcName)]['_FillValue'] = int_missing_value*100
```
CoryMartin-NOAA, line 190: same comment

CoryMartin-NOAA (approved):
Looks good. Curious about why 2 of them have `int_missing_value*100` rather than just making `int_missing_value = 99900`, is it so they all have `int_missing_value = 999`?

travissluka (changes requested):
@gthompsnJCSDA Salinity is a wonderfully confusing variable., I had to refresh my memory on this stuff. Technically, `PSU` and `g/kg` are both wrong. It should instead be unitless. 

There are two standards for salinity 1) practical salinity and 2) absolute salinity.
- Practical salinity is what is used for pretty much everything we currently do. It is defined as a ratio of conductivity, and is _close_ to g/kg, but not exactly. Everyone tends to use "PSU" as the unit, even though that is technically wrong. 
- Absolute salinity is newer (past decade), and is properly defined as g/kg. I have never actually seen absolute salinity used. 

In practice, the difference between practical salinity and absolute salinity is small enough that it usually falls within the range of observation sensor error, so people tend to ignore the difference. Where it does really matter is in the model variables, where there could be big differences in vertical density gradient if the wrong type of salinity is used. MOM6 uses practical salinity.

Short answer: to be clear and correct, and it should be unitless, and noted somewhere that it is practical salinity, not absolute salinity. 

@guillaumevernieres can double check me on everything I just said.

gthompsnJCSDA:
@travissluka   Are you OK with me making the salinity units=1 which we are using for things like a ratio?  This would be our standard treatment instead of unitless.  I will go ahead under the assumption this is OK and you chime in with more info if needed.

YoulongXia-NOAA:
> @travissluka Are you OK with me making the salinity units=1 which we are using for things like a ratio? This would be our standard treatment instead of unitless. I will go ahead under the assumption this is OK and you chime in with more info if needed.

I used unitless as @travissluka suggested. Do I need to change it into "1"?

gthompsnJCSDA:
> > @travissluka Are you OK with me making the salinity units=1 which we are using for things like a ratio? This would be our standard treatment instead of unitless. I will go ahead under the assumption this is OK and you chime in with more info if needed.
> 
> I used unitless as @travissluka suggested. Do I need to change it into "1"?


The choice of explicit ```unitless``` is not possible due to how we decided things in the Data Conventions.  It must be left out entirely or it should be "1"  I gave my opinion that it should be "1" because it is technically a ratio.  That makes it very similar to relative humidity or many other variables whose units are 1.

travissluka:
@gthompsnJCSDA `1` sounds fine with me.

gthompsnJCSDA:
Arghhh!!!  I am so lost about the cdash results saying this...

```
DIFFER : LENGTHS OF RECORDS : Location (76) <>  (0)
```

as one example of the four converters (which all four are doing the similar thing).  How is this possible.   I uploaded the right working files to Git.  My testoutput dir matches the repo in the branch.  What is going on here someone?  The ioda-validate failures will resolve themselves when the two dependencies get merged, but these 4 additional failures are a total mystery (and frustrating) to me.

YoulongXia-NOAA (approved):
Look good for me. I approved this PR.

gthompsnJCSDA:
> Look good for me. I approved this PR.

Youlong:  Would you please use the Review procedure just to check the box for approved?  That is better than adding a comment to the PR.  Thank you!

travissluka (approved):

YoulongXia-NOAA (approved):

YoulongXia-NOAA:
> > Look good for me. I approved this PR.
> 
> Youlong: Would you please use the Review procedure just to check the box for approved? That is better than adding a comment to the PR. Thank you!

Now I did. Thank you.

rhoneyager (approved):

PR #901, opened 2022-04-07
New python converter for satwinds from nesdis ascii to ioda

haydenlj:
## Description

New converter for satwinds from nesdis ascii to ioda with test reference input/output files. 

### Issue(s) addressed

- fixes #<issue_number>

passes tests and positive reviews

## Dependencies
none

## Impact
no impact to other repos expected


BenjaminRuston (approved):

src/conventional/amv_ssec_ascii2ioda.py
```diff
@@ -0,0 +1,320 @@
...
+    logging.info("--- {:9.4f} total seconds ---".format(time.time() - start_time))
+
+
+def read_file(file_name, data):
+
+    with open(file_name, newline='') as f:
+        reader = csv.DictReader(f, skipinitialspace=True, delimiter=' ')
+        keyerr = False
+        freqerr = False
+        saterr = False
```
srherbener, line 151: It looks like the purpose of `freqerr` and `saterr` is to print the warning message once (the first time) when encountering these errors (ie, so you don't get thousands of warning messages). Is this correct? If so, one issue I can think of is that if you encounter more than one unknown channel/satellite type you only get warned about the first one that you encounter. To address that issue you could instead create top-level dictionaries for unknown channels and another for unknown satellites. Then in the `get_frequency` and `get_id` functions, when an unknown type is encountered do something like: `unknown_chan[obs_type] = 1`. It doesn't matter what the value you assign is, the idea is that you get one key for each unknown channel/satellite id and you can then print out all the unknown ids at the end of the script.

src/conventional/amv_ssec_ascii2ioda.py
```diff
@@ -0,0 +1,320 @@
...
+            except KeyError as e:
+                if not keyerr:
+                    logging.warning(file_name + ' is missing variable ' + e.args[0])
+                keyerr = True
+
+                if (e.args[0] == 'dir') | (e.args[0] == 'spd'):
+                    data['eastward_wind'] = np.append(data['eastward_wind'], float_missing_value)
+                    data['northward_wind'] = np.append(data['northward_wind'], float_missing_value)
+                else:
+                    outname, missing = get_outname(e.args[0])
```
srherbener, line 197: the `get_outname` function returns uninitialized values if the e.args[0] string doesn't match what the function is testing for. Should a check be put in the function for a non-match and a flag or special `outname` be returned so you can avoid putting a bad entry in the `data` dictionary? Or is this getting checked somewhere else so when you are here, you know you have a match?

srherbener (commented):
Thanks for adding this. The code looks great. I just have a couple comments.

src/conventional/amv_ssec_ascii2ioda.py
haydenlj: I'm not sure I understand. Do you mean if the file has a variable that's not one of the known ones in the test file? I'm not sure what would be best to do in that situation. get_outname contains all the known variables besides spd and dir

src/conventional/amv_ssec_ascii2ioda.py
haydenlj: Thanks, the change has been made

src/conventional/amv_ssec_ascii2ioda.py
```diff
@@ -0,0 +1,323 @@
...
+            logging.warning(file_name + ' contains unknown satellite ID ' + s)
+
+    return data
+
+
+def get_frequency(obs_type, unk_freq):
+
+    if obs_type == 'IR':
+        freq = 2.99792458E+14/10.7
+    elif (obs_type == 'WVCA') | (obs_type == 'WVCT') | (obs_type == 'WV'):
```
gthompsnJCSDA, line 213: I believe my change here makes more sense to test simply if "WV" in obs_type rather than 3 specific choice letter combinations.

src/conventional/amv_ssec_ascii2ioda.py
```diff
@@ -0,0 +1,323 @@
...
+    elif (obs_type == 'VIS'):
+        freq = 2.99792458E+14/0.65
+    else:
+        unk_freq.append(obs_type)
+        freq = float_missing_value
+
+    return freq, unk_freq
+
+
+def get_id(sat_name, unk_sat):
```
gthompsnJCSDA, line 226: Is it not easier just to make a top-level dictionary variable near the top of code rather than having this number/name assignment embedded down lower in the code where it isn't very clear to add new sat name sources?

src/conventional/amv_ssec_ascii2ioda.py
```diff
@@ -0,0 +1,323 @@
...
+    else:
+        unk_sat.append(unk_sat)
+        satid = int_missing_value
+
+    return satid, unk_sat
+
+
+def get_outname(key):
+
+    if key == 'type':
```
gthompsnJCSDA, line 247: Similar to the prior comment, I believe we can just list the missing data type together with the variable name near the very top of the code in the Dictionary that is already present at the top.  Otherwise, it is too easy to miss changing in more than one place.  I spent considerable time in editing the 1st version of code to keep all JEDI Data Conventions information at the very top of the code - making it super simple to locate and change rather than multiple places in code.

src/conventional/amv_ssec_ascii2ioda.py
srherbener: I am concerned that if you encounter a variable from the file that isn't in the list that get_outname checks, you might end up with some garbage in the output ioda file and more importantly the user won't be aware that an unknown file variable was encountered.

If it's just a matter of keeping things consistent within this script I think get_outname is okay as is. What I mean is that there is a list of variables that are pulled from the file during the read, and that list is in sync with the list that get_outname checks. If this is the case, I think it would be helpful to put in comments where the variables are being extracted from the file and in get_outname pointing out that these two sections of code need to remain in sync. If a new variable is added to the section reading from the file, then it has to also be added in get_outname for example.

However, if a new variable is introduced by simply adding it to the file, then I think get_outname (or a new check) needs to handle unknown variables (in a similar fashion as get_frequency and get_id) so that the user becomes aware that a new variable is not being handled properly.

src/conventional/amv_ssec_ascii2ioda.py
srherbener: If you decide to keep the three conditional tests (as code is now), please change the bitwise or operators ('|') to logical or operators ('or'). Ie,
```suggestion
    elif (obs_type == 'WVCA') or (obs_type == 'WVCT') or (obs_type == 'WV'):
```

haydenlj: I have removed this check in favor of a known frequency dictionary

src/conventional/amv_ssec_ascii2ioda.py
haydenlj: known frequency and known satellite dictionaries have been made and the functions have been removed

src/conventional/amv_ssec_ascii2ioda.py
haydenlj: I understand now. I have added a dictionary of known variables and section to print out any unknown variables in the file. The except block is similar since it would only reach this section of code while trying to write out one of the known variables that wasn't present (ie in the event of a KeyError in the code above)

srherbener (approved):
Changes look good - thanks for addressing  my comments!

srherbener:
@gthompsnJCSDA have all your concerns been addressed? If so, I can merge. Thanks!

BenjaminRuston:
@gthompsnJCSDA this is ready to go no?

gthompsnJCSDA (approved):
At some future time, I would like to consolidate the ```known_var``` dictionary into the higher up dictionary and reduce the number of places where the IODA variable name appears.

PR #921, opened 2022-04-22
Replaced "land_type_index" by "land_type_index_NPOESS"

gmao-jjin3:
Replaced "land_type_index" by "land_type_index_NPOESS" in "gsi_ncdiag.py".  Current UFO requires the land surface data source to be included in its variable name: land_type_index_IGBP, land_type_index_USGS, or land_type_index_NPOESS.  See lines 150-160 in:
https://github.com/JCSDA-internal/ufo/blob/develop/src/ufo/crtm/ufo_radiancecrtm_mod.F90    

UFO tests fail for IR data made from GSI outputs  with a land type variable "land_type_index". Current CRTM uses NPOESS as a default source. Therefore, the Python program "gsi_ncdiag.py" is edited in this PR.


gmao-jjin3:
For some radiance data outputs such IASI from GSI, the type "**Forecast_adjusted**" represents simulated brightness temperature after bias correction and adjustments because of correlated errors are considered in GSI.  Therefore,  "GsiHofXBc" which is saved from this data type does not represent GSI forward processing after bias correction. In this new  update, GsiHofXBc is saved from "ObsValue" subtracted by "Obs_Minus_Forecast_adjusted", and correctly represents GSI forward processing after bias correction.

danholdaway (approved):

src/gsi-ncdiag/gsi_ncdiag.py
```diff
@@ -1395,12 +1389,12 @@ def toIODAobs(self, OutDir, ObsBias, QCVars, TestRefs, clobber=True):
         for key, value2 in chan_metadata_dict.items():
             try:
                 if value2 in chan_metadata_int:
-                    outdata[(value2, 'MetaData')] = self.var(key).astype(np.int32)
+                    outdata[(value2, 'VarMetaData')] = self.var(key).astype(np.int32)
```
CoryMartin-NOAA, line 1392: Are we positive we want `VarMetaData` in here? I thought that was a deprecated group?

CoryMartin-NOAA (commented):
Sorry we have allowed this to fall through the cracks. One comment/question for discussion and then we can approve/merge.

src/gsi-ncdiag/gsi_ncdiag.py
srherbener: The nvars dimension and the VarMetaData group have been deprecated. These come from an attempt to organize meta data long before the OBS Conventions were developed.

All meta data variables should be placed in the MetaData group and each variable should be dimensioned by a dimension of whatever is being described. Ie, channel frequencies, channel polarization etc should be dimensioned by nchans; latitude, longitude dimensioned by nlocs, etc.

The name of the dictionary `chan_metadata_dict` seems to imply that these variables are MetaData describing channels. If that is the case please change these back to MetaData and nchans; otherwise let's discuss further and figure out what dimensions to use (we still want these variables in the MetaData group). 

Thanks!

srherbener (changes requested):

src/gsi-ncdiag/gsi_ncdiag.py
gmao-jjin3: Thank you both! The Discover is down today. I cannot remember why I made this change. Have to check once the computer is up.

gmao-jjin3: This VarMetaData is rename back to the MetaData in a new PR https://github.com/JCSDA-internal/ioda-converters/pull/1038 . The new PR converts GNSS diagnostic variables and other data in GSI .

CoryMartin-NOAA:
@gmao-jjin3 what about modifying this PR to just include the NPOESS change?

gmao-jjin3:
> @gmao-jjin3 what about modifying this PR to just include the NPOESS change?

Done.

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
@srherbener you requested changes on this which I think are now no longer part of this PR, so I re-requested your review.

srherbener (approved):
Looks good now. Thanks for removing the `VarMetaData` group.

PR #923, opened 2022-04-26
Handling missing values of TROPICS observed data

fabiolrdiniz:
## Description
This DRAFT PR is touching the TROPICS converter. It's adding a mapping of missing values from the input file into the ones used by the output file (IODA) through a masking procedure. Although this addition doesn't check for negative brightness temperatures, it helps to avoid encoding IODA files with -999.0 as actual observed values whenever they happen (-999.0 is the missing value from the input file).

### Issue(s) addressed
- fixes TO-BE-CREATED

The converter for TROPICS data still works as expected.

## Dependencies
None

## Impact
None

### DIFF
--- a/src/hdf5/tropics_2ioda.py
+++ b/src/hdf5/tropics_2ioda.py
@@ -27,7 +27,6 @@ import ioda_conv_engines as iconv
 from orddicts import DefaultOrderedDict
 
 # globals
-TROPICS_WMO_sat_IDs = [691, 895, 896, 897, 898]
 TROPICS01_WMO_sat_ID = 691
 TROPICS02_WMO_sat_ID = 895
 TROPICS03_WMO_sat_ID = 896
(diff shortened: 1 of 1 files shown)


srherbener:
@fabiolrdiniz this test is failing due to a defect in the validator that is causing it to crash (jcsda-internal/ioda/issues/690). For the develop branch, we have configured the tests to only run the validator and ignore the warnings and errors that the validator produces. It's done this way since the develop branch doesn't have the JEDI Conventions coded in yet (of which the validator is checking).

In this case you don't need to repair the file that is produced by the converter since this is work based on the develop branch.

We have a PR (jcsda-internal/ioda/pull/691) under review that will fix the validator crash and allow this test to pass again. @rhoneyager and I are working on getting this PR merged in soon.

BenjaminRuston (approved):

fcvdb (commented):
The code changes look good to me.

fcvdb (approved):

Issue #932, opened 2022-05-03
Critical slowdown in src/convential/amv_ssec_ascii2ioda.py for multiple files


PatNichols:
## Description

The python script src/convential/amv_ssec_ascii2ioda.py is incredibly slow when reading multiple files. Instead of a near linear increase in time with each new file that is added to the run, the run time increases by a factor >> 1.

## Requirements

Expected behavior is that the runtime should be proportional to the sum total of file sizes. No new software dependencies.

## Acceptance Criteria (Definition of Done)

The runtime should scale roughly with sum of the sizes for files.

## Dependencies

None

symoore90:
Closing due to lack of activity.

PR #933, opened 2022-05-03
amv_ssec performance bug

PatNichols:
## Description

The python script src/convential/amv_ssec_ascii2ioda.py was incredibly slow when reading multiple files. This change improved performance a factor of 5 on our test by creating a local_data for each file then appending that to the global data variable rather than appending to global_data directly. Since np.append reallocates an array with each append this saves memory from thrashing. A read of six files went down from 6min to 65 seconds.
  Only the file in question was changed.

### Issue(s) addressed

The above issue was outlined in issue #932 

The changes are done and ready to be merged after review. The time is nearly linear with # of files accounting for file sizes.

## Dependencies

None

## Impact

No impact on other projects or in other parts of ioda-coverters.


PatNichols:
the only real change is in the file mentioned

BenjaminRuston:
did you commit from your build directory?  

there are 668 files changed,, so think this did not go as intended?

fabiolrdiniz:
Hi, @PatNichols. Thanks for adding me here. Nice that you worked on it. I believe that the `bld` folder has been added to your commit. Usually, people create a folder name as `build` inside the cloned copy of the repository, because the `.gitignore` file will ignore it. Please see this line: https://github.com/JCSDA-internal/ioda-converters/blob/a0bc40bdfbbbe897c2a68417c11abbff612adb71/.gitignore#L2

PatNichols:
This branch is broken will create new pr.

PR #934, opened 2022-05-04
fix for performance bug in sat winds 

PatNichols:
## Description
The python script src/convential/amv_ssec_ascii2ioda.py was incredibly slow when reading multiple files. This change improved performance a factor of 5 on our test by creating a local_data for each file then appending that to the global data variable rather than appending to global_data directly. Since np.append reallocates an array with each append this saves memory from thrashing. A read of six files went down from 6min to 65 seconds. Only the file in question was changed.

## issues addressed

The above issue was outlined in issue #933

The changes are done and ready to be merged after review. The time is nearly linear with # of files accounting for file sizes.

## Dependencies

None 

## Impact

None. No changes in interfaces from develop.

### DIFF
--- a/src/conventional/amv_ssec_ascii2ioda.py
+++ b/src/conventional/amv_ssec_ascii2ioda.py
@@ -169,6 +169,11 @@ def main(file_names, output_file, datetimeRef):
 
 
 def read_file(file_name, data):
+    local_data = {}              # Before assigning the output types into the above.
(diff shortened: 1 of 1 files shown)


haydenlj (approved):

fabiolrdiniz (approved):
Thanks, @PatNichols!

PR #943, opened 2022-05-10
improve and clean viirs aod

jeromebarre:
## Description

Cleaned the code on the error specification using the nesdis option. 
Removed useless variables.
Removed a for loop that was problematic given the size and number of viirs files.
The code (within those lines) is about 50x faster now in term of process time. I did the tests on my mac... 

Avoid as much as possible for loops with python (I don't think this language has been designed for that....)! Use list comprehensions

Tagged colleagues:
@bhuang95, @mpagowski, @mer-a-o, @rhoneyager

### DIFF
--- a/src/compo/viirs_aod2ioda.py
+++ b/src/compo/viirs_aod2ioda.py
@@ -124,29 +124,10 @@ class AOD(object):
                 obs_time = obs_time[mask_thin]
 
             # defined surface type and uncertainty
-            sfctyp = 0*qcall
-            uncertainty = 0.0*errs
-            uncertainty1 = 0.0*errs
-            uncertainty2 = 0.0*errs
-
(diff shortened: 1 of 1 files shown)


andytangborn (commented):
Is sfctyp needed? It looks like you have taken it out. Otherwise, it's good to remove the loops.

jeromebarre:
> Is sfctyp needed? It looks like you have taken it out. Otherwise, it's good to remove the loops.

I don't think this variable is needed. The assignment line was weird and probably a relic from prior uncleaned devs. I didn't changed the testoutputs so the results are unchanged.

jeromebarre:
@andytangborn if you are OK with those changes would please accept this review so this can be merged? Thanks

andytangborn (approved):

PR #953, opened 2022-05-26
Fix for ncdiag converter for conv surface type

FanHan-NOAA:
## Description

This PR adds a temporary fix for conversion of conv surface t & q, which are 2 meter temperature and moisture observations. In the current conversion, the height of the 2 meter temperature and moisture is computed from the station height of surface pressure plus 2m. Because of this, 2 meter temperature and moisture observations are not properly combined with surface pressure observation. These 2m variables are used in the UFO sfcpcorrected GSI scheme. A temporary fix is to comment out the section that computes the height of 2m variables as station elevation plus 2m, but instead directly use the station elevation for the 2m variables. 

### Issue(s) addressed

N/A

Tests pass and approval by two reviewers

## Dependencies

N/A

## Impact

N/A

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -867,11 +867,14 @@ class Conv(BaseGSI):
                             tmp[tmp == 10009.] = self.FLOAT_FILL  # for u,v sfc Height values that are 10+9999
                             # GSI sfc obs are at 0m agl, but operator assumes 2m agl, correct output to 2m agl
                             # this is correctly 10m agl though for u,v obs
-                            if lvar == 'Height' and self.obstype in ['conv_t', 'conv_q']:
-                                elev = self.var('Station_Elevation')[idx]
-                                hgt = elev + 2.
-                                hgt[hgt > 9998.] = self.FLOAT_FILL
-                                tmp = hgt
+                            # --- temporarily comment out the following so 2m_t & 2m_q can be properly combined
+                            # --- with surface_pressure ioda obs as single record because 2m_t and 2m_q are used
+                            # --- in UFO surface pressure correction scheme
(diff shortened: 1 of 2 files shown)


src/gsi-ncdiag/gsi_ncdiag.py
```diff
@@ -867,11 +867,14 @@ def toIODAobs(self, OutDir, clobber=True, platforms=None):
...
-                                hgt = elev + 2.
-                                hgt[hgt > 9998.] = self.FLOAT_FILL
-                                tmp = hgt
+                            # --- temporarily comment out the following so 2m_t & 2m_q can be properly combined
+                            # --- with surface_pressure ioda obs as single record because 2m_t and 2m_q are used
+                            # --- in UFO surface pressure correction scheme
+                            # if lvar == 'Height' and self.obstype in ['conv_t', 'conv_q']:
+                            #     elev = self.var('Station_Elevation')[idx]
+                            #     hgt = elev + 2.
+                            #     hgt[hgt > 9998.] = self.FLOAT_FILL
```
CoryMartin-NOAA, line 876: Can we keep the fill values for where it is sufficiently large? Or is it not needed?

FanHan-NOAA: The variable `tmp` on the lines 866-867 directly above the comments already had fill values, so I think it is not needed, unless you know there are cases where the missing value is neither 9999 nor 10009.

CoryMartin-NOAA: I'm not 100% sure, we can assume it is fine for now though.

CoryMartin-NOAA (approved):

frolovsa (approved):
@FanHan-NOAA thank you for adding this so quickly. This agrees with what we discussed in the meeting this week.

srherbener (approved):

Issue #384, opened 2020-12-17
Bufr Converter: Add ability to read BUFR sequences.


rmclaren:
Currently the BUFR converter uses the NCEPLib BUFR calls **ufbrep** and **ufbint** to read BUFR mnemonic data. There is however a seperate class of data, sequences, which we do not handle (read using **ufbseq**).

YAML wise we could extend the section mnemonicSets with a new type, **_sequence_**.

rmclaren:
Currently thinking about the following YAML format changes to the BUFR description section.

- Remove mnemonicSets section
- Add sequences section (I'm going to try to eliminate the need for this section).
- Bufr export variables now indicate a path to the Mnemonic so it lists the sequences.
- Bufr export variables now list the indices that are desired to be exported for a specific key.

Here is an example how this might look:

![Screenshot from 2021-02-10 09-03-03.png](https://images.zenhubusercontent.com/5efcacfc3b91075e014181ba/c56529d8-e25e-4e6a-9443-39a2de6caf47)

gthompsnJCSDA:
@rmclaren  So is a sequence the same as what debufr.x reports as a 'replication?'  If I interpret your screen capture correctly, I assume that a sequence of __root__ would be any unchanging header-like data for station ID/name, position info, a radiosonde launch time; but any named sections below root are referring to the repeated info such as a air_temperature at distinct heights of a sonde while it ascends.  Is this correct?

Also, are the final three vars in your exports doing mathematics such as value of ROSEQ3 divided by PCCF?

rmclaren:
Hi Greg,

Let me start by posting a part of a BUFR table as an example:

![image.png](https://images.zenhubusercontent.com/5efcacfc3b91075e014181ba/5912745e-cd6a-4d25-b984-8d047f8812c3)

So in BUFR there are basically two types of mnemonics (basically data field ID's). **Data** mnemonics such as CLATH which are straight forward mnemonics that represent a data field (via a "Table B" descriptor), and **sequence** mnemonics that represent an ordered set of sub-elements (defined by a "Table D" descriptor) where each element can be described by its own own mnemonic either **sequence** or **data** (sequences can be arbitrarily nested).

A simplistic way of thinking about it is that an **data** mnemonic is just a data field, and a **sequence** mnemonic is a list of other mnemonics.

[...]

PR #1006, opened 2022-07-12
Feature/query cxx

### LINKED ISSUE #425: NOAA EMC Observation Data (BUFR/PrepBUFR) to IODA Conversion Software Development
### **What is it?**
- C++ tool that converts BUFR file to IODA ObsGroup object (saving to as NetCDF if desired).
- YAML file is used to describe the mapping between the input and the output.
- Supports simple transformations of “export” data so that the units of the data can be converted to the desired representation.
- Compiles to a library that can be used by other applications.
- Creates convenient executable bufr2ioda.x <yaml path>.
- Source found in [ioda-converters/src/bufr](https://github.com/JCSDA-internal/ioda-converters/tree/develop/src/bufr)

### **Why do we need it?**

- Data observation processing gathers data from various sources and generates BUFR encoded data (like PREPBUFR files). In the past the target applications (models and simulations) would use these files directly. This meant each endpoint needed to write code to get the data they needed.
- The JEDI application framework takes care of reading and filtering of the source data for the endpoints (using IODA and UFO components). It handles its internal data as ioda::ObsGroup objects, but only has the capability to read these from NetCDF files. It is extensible with new data backends, however.
- Obs Processing needs to have the ability to provide these JEDI based applications with the data they need, but all the infrastructure is built around BUFR. Up until recently this was handled with python programs the translated the data to the NetCDF format. It’s a lot of work to create a program for each data type, however, and python is relatively slow to run.
- The BUFR converter is a fast format converter which only needs some configuration to work. In the long term it can be integrated into JEDI IODA as another data backend removing the need for a pre-processing step.

### **Documentation**

[BUFR Converter explained](https://github.com/JCSDA-internal/ioda-converters/blob/develop/src/bufr/README.md)

**How-to Documentation**
- How to run BUFR Converter on [Hera](https://docs.google.com/document/d/18j78LwUe_nYMWCHZUmpjP0FDO6gYO4CEN9JIjSJBqNU/edit?usp=sharing) and [Orion](https://docs.google.com/document/d/13u88bY6jTrnwRDYy4MQe4ml2luNV6itpzktdoRYC8_w/edit?usp=sharing)

- [How to build/install NCEPLIB-bufr and IODA-Converters with feature/query](https://docs.google.com/document/d/1_AvC3Ebt3R5EeMtHgl1nwKLEPXy-2J9nIqpDJvCL7ME/edit?usp=sharing) - updated on July 30, 2021

**Presentations by Ron Mclaren:**

- [BUFR Converter - IODA-Engines based Ingester @ JEDI Discussion](https://drive.google.com/file/d/1Qir_GVwLyihu0RjUaP7sOxOy3Gv8nT5o/view?usp=sharing) - - Feb 03, 2021

- [BUFR Converter - IODA-Engines based Ingester @ MDAB-DAQC Group Meeting](https://drive.google.com/file/d/1E5w-kHB6fC8llEre-c1mzHPIPW_dRG22/view?usp=sharing) - Feb 10, 2021

- [feature/query Introduction](https://docs.google.com/presentation/d/1q-_-pB5z3yBflwyDo6EWXn4CXE9tZlMC0TdMqIrcs08/edit?usp=sharing) - July, 20, 2021

### **Resources**
- [JEDI Documentation (latest)](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/)
- [ECMWF ODB Database](https://apps.ecmwf.int/odbgov/unionall/)
- [NOAA NCEP BUFR Tables](https://www.nco.ncep.noaa.gov/pmb/codes/nwprod/decoders/decod_shared/fix/)

### Task List 
[feature/query](https://docs.google.com/spreadsheets/d/1D68rWHCvuv3dOpZUGAGMqDdHqQD6c9VexMnJRf_QoXU/edit?usp=sharing)

---

*(Instructions: replace text in this and all sections with your own text)*

Provide a detailed description of this issue.

What problem needs to be fixed? What new capability needs to be added?

If this is a bug, describe the current behavior.

[Be sure to add a Pipeline, Label, Estimate, Assignees, and Epic](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/practices/issues.html)

## Requirements

If this is a new feature: What does the new code need to accomplish? Does it require new software dependencies (e.g. new jedi-stack components or new python modules?)

If this is a bugfix: What is the expected behavior?

What does it mean for this to be finished?

## Dependencies

What must be done before this can be done?
 - add issue dependencies in ZenHub as appropriate

Does this block progress on other issues?
 - add this issue as a dependency to other ZenHub issues as appropriate

rmclaren:
## Description

Modifies C++ based BUFR ingestion to use queries in order to retrieve specific BUFR data fields.

### Issue(s) addressed

- fixes #425, #384

Changes the way the the BUFR to IODA YAML files work. They now use queries. Please too the following slides for a more detailed explanation: https://docs.google.com/presentation/d/19faSUyeunw6M8oDhtm6luA_FIgLiPPUYTByJuChqYBA/edit#slide=id.p

## Dependencies

Will require once feature/query_cxx_types is merged
- [ ] waiting on NCEPLIBS-bufr/feature/query_enhancements
- [ ] merge with jcsda-internal/ioda/pull/779


tools/bufr/print_queries.cpp
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    closbf_f(FileUnit);
+    close_f(FileUnit);
+}
+
+
+int main(int argc, char** argv)
+{
```
rhoneyager, line 237: See above for try-catch blocks.

rhoneyager (commented):
Minor comments.

src/bufr/BufrParser/Exports/Splits/Split.h
```diff
@@ -8,17 +8,18 @@
...
 namespace Ingester
 {
     /// \brief Base class for all Split objects that split data into sub-parts
     class Split
     {
      public:
-        Split() = default;
+        explicit Split(const std::string& name);
 
         /// \brief Get set of sub categories this split will create
```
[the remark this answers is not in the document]
nicholasesposito: And what are the sub-parts?

src/bufr/BufrParser/Query/DataProvider.h
```diff
@@ -0,0 +1,87 @@
...
+        // Getters to get the raw data by idx. Since fortran indices are 1-based,
+        // we need to subtract 1 to get the correct c style index.
+        inline FortranIdx getIsc(FortranIdx idx) const { return isc_[idx - 1]; }
+        inline FortranIdx getLink(FortranIdx idx) const { return link_[idx - 1]; }
+        inline FortranIdx getItp(FortranIdx idx) const { return itp_[idx - 1]; }
+        inline FortranIdx getJmpb(FortranIdx idx) const { return jmpb_[idx - 1]; }
+        inline Typ getTyp(FortranIdx idx) const { return typ_[idx - 1]; }
+        inline std::string getTag(FortranIdx idx) const { return tag_[idx - 1]; }
+
+        inline FortranIdx getInode() const { return inode_; }
```
nicholasesposito, line 62: what is defined as a "node" in this case?

src/bufr/BufrParser/Query/DataProvider.h
```diff
@@ -0,0 +1,87 @@
...
+
+    /// \brief Responsible for exposing the data found in a BUFR file in a C friendly way.
+    class DataProvider
+    {
+     public:
+        DataProvider() = default;
+        ~DataProvider() = default;
+
+        /// \brief Read the data from the BUFR interface for the current subset and reset the
+        /// internal data structures.
```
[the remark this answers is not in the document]
rmclaren: So when you read BUFR data with NCEPLIB-bufr you always read the data one subset at a time. So you will have some kind of loop (see File.cpp::execute) that loops through each message and subset and calls `updateData`. updateData causes the DataProvider to load the data from the current subset into its internal data structures so that the data returned will always be for the subset we are currently on.

```cxx 
        auto dataProvider = DataProvider();  //Create the data provider
        auto resultSet = ResultSet(querySet.names());
        auto query = Query(querySet, resultSet, dataProvider);

        while (ireadmg_f(fileUnit_, subset, &iddate, SubsetLen) == 0)
        {
            while (ireadsb_f(fileUnit_) == 0)
            {
                status_f(fileUnit_, &bufrLoc, &il, &im);
                dataProvider.updateData(bufrLoc);  // Read the data for the current subset.
                query.query();
            }
```

src/bufr/BufrParser/Exports/Splits/Split.h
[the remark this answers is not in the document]
rmclaren: There is a feature that allows you to split the BUFR file into different subfiles according to data found in a BUFR element. *Side Note: Unfortunately it does have far reaching effects (makes the code more complicated) so I'm not sure it was such a good idea.* So for example the YAML file might contain the following export definition:

```yaml
        splits:
          satId:
            category:
              variable: satelliteIdentifier
              map:
                _3: metop-b
                _4: metop-a
                _5: metop-c
#               _206: noaa-15
                _209: noaa-18
                _223: noaa-19
```

So this YAML specifies one category of data **satID** which is to be split into separate datasets based on the BUFR variable called ***satelliteIdentifier***. The way these variables are intended to be split is defined in the map. So you will have one dataset (Subcategory) where satelliteIdentifier is 3, another where it equals 4, and so on.

So the Set of subcategories satId 
{{metop-b}, {metop-a}, {metop-c}, {noaa-15}, {noaa-18}, {noaa-19}}

It is also possible to have multiple categories defined as in the following yaml blurb:

```yaml
        splits:
          hour:
            category:
              variable: timestamp_hour
          minute:
            category:
              variable: timestamp_minute
              map: # Optional
                _5: five #can't use integers as keys so underscore
                _6: six
                _7: seven
```

Doing this creates subcategories (datasets) for each possible combination as follows:

So the Set of subcategories
{{0,          five},
{2,          five},
{3,          five},
...
{23,        seven}}

Each subcategory knows what split it is (ex hour:2, minute:five) and the user can then name the output file by specifying a templated string as follows.

```yaml
    obsdataout: "./testrun/gdas.t18z.1bmhs.tm00.{splits/hour}.{splits/minute}.filter_split.nc"
```

Its a little complicated unfortunately... Let me know if this makes sense.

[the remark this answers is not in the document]
rmclaren: You'll get an appropriately named file for each subcategory.

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -7,137 +7,141 @@
...
 
-    BufrParser::BufrParser(const eckit::Configuration& conf) :
-        description_(BufrDescription(conf)),
-        fortranFileId_(0),
-        table1FileId_(0),
-        table2FileId_(0)
+    BufrParser::BufrParser(const eckit::Configuration &conf) :
+            description_(BufrDescription(conf)),
+            file_(bufr::File(description_.filepath(),
+                             description_.isWmoFormat(),
```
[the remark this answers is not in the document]
rmclaren: NCEP BUFR is us, WMO BUFR is used by folks in Europe. The NCEP BUFR files have embedded meta data for the structure of the BUFR file (what you see when you do gettab). WMO BUFR files do not contain this meta data. The data must be known ahead of time before you read the file. So for WMO bufr files there are a bunch of extra table files you have to read before you can make sense of the BUFR file. NCEPLIB-bufr can handle both, but the process for opening the files is a little different (have to point it to where the extra files are stored).

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -7,137 +7,141 @@
...
+        std::cout << "Executing Queries" << std::endl;
+        const auto resultSet = file_.execute(querySet, maxMsgsToParse);
 
-        char subset[SubsetStringLength];
-        int iddate;
-
-        unsigned int messageNum = 0;
-        while (ireadmg_f(fortranFileId_, subset, &iddate, SubsetStringLength) == 0)
+        std::cout << "Building Bufr Data" << std::endl;
+        auto srcData = BufrDataMap();
```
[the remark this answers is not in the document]
rmclaren: `auto` is  just a c++ keyword that tells the compiler it should figure out a return type for you. It's a newish feature introduced in c++ 11. It makes the code look cleaner because it avoids a lot of redundant information, for example:

`std::string myString = std::string(dataBuffer, 20)` vs `auto myString = std::string(dataBuffer, 20)`

Also it makes it easier to change things if you want to say myString to be something other than a std::string... Some template tricks also depend on this feature as you can return `auto` and let the compiler figure out the return type from a templated function for you...

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -7,137 +7,141 @@
...
             }
-
-            if (maxMsgsToParse > 0 && ++messageNum >= maxMsgsToParse) break;
         }
 
-        return exportData(collectors.finalize());
+        std::cout << "Exporting Data" << std::endl;
+        auto exportedData = exportData(srcData);
+
+        auto timeElapsed = std::chrono::steady_clock::now() - startTime;
```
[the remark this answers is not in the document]
rmclaren: It just leaves it to the compiler to figure out what the return type is and act accordingly.

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -7,137 +7,141 @@
...
         // Split
         CategoryMap catMap;
-        for (const auto& splitPair : splitMap)
+        for (const auto &split : splits)
         {
             std::ostringstream catName;
-            catName << "splits/" << splitPair.first;
-            catMap.insert({catName.str(), splitPair.second->subCategories(dataCopy)});
+            catName << "splits/" << split->getName();
+            catMap.insert({catName.str(), split->subCategories(dataCopy)});
```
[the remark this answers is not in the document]
rmclaren: You can split on anything... I guess in practice we've used them for satellite ids.

src/bufr/BufrParser/Query/DataProvider.h
```diff
@@ -0,0 +1,87 @@
...
+
+        /// \brief Tells the Fortran BUFR interface to delete its temporary data structures that are
+        /// are needed to support this class instanc.
+        void deleteData();
+
+        /// \brief Get the subset string for the currently active message subset.
+        std::string getSubset() const { return subset_; }
+
+        // Getters to get the raw data by idx. Since fortran indices are 1-based,
+        // we need to subtract 1 to get the correct c style index.
```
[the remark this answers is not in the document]
rmclaren: We have different getters for different types of data from the BUFR tables which are related to each other by their `node Id` which is basically just an index into their data structure. The different types of data are used to understand the underlying structure of the data (mainly how things are nested with respect to each other).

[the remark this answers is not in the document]
rmclaren: Please let me know if I interpreted the question correctly :)

src/bufr/BufrParser/Query/DataProvider.cpp
```diff
@@ -0,0 +1,99 @@
...
+        int *intPtr = nullptr;
+        double *dataPtr = nullptr;
+
+        int strLen = 0;
+        char *charPtr = nullptr;
+
+        if (isc_.empty())
+        {
+            get_isc_f(&intPtr, &size);
+            isc_ = gsl::span<const int>(intPtr, size);
```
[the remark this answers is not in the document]
rmclaren: This is C style code used to interact with the NCEPLIB-bufr C interface. Temporary variables intPtr and size were previously declared as `int* intPtr;` (stores a pointer) `int size` (stores a value) which allocates them on the `stack`. I then send the bufr interface pointers to these blips of memory. The bufr interface then changes these values to point to the location in memory were it holds this data (intPtr) and also gives me the size (number of bytes) of the **isc** array. gal::span is just a C++ wrapper that adds a layer of convenience and safety to access the data stored in this raw buffer (no pun intended).

src/bufr/BufrParser/Query/DataProvider.h
rmclaren: Inode just gives you the starting Node Idx for the message subset that we are currently busy processing. This is what the call it in NCEPLIB-bufr so I was just trying to be consistent with them... Sorry its a little confusing.

guillaumevernieres:
Could you merge develop and resolve the conflict @rmclaren ?

rmclaren:
@guillaumevernieres. I just merged the latest from develop into feature/query_cxx. Sorry I got distracted with other things :)

srherbener:
@rmclaren sorry about the slow response. I'm taking a look at this now. 

Please consider removing the draft status of this PR. I think the main reason to keep this PR in draft status is that it still requires code changes before it makes sense to merge it into develop. Otherwise I think we should just put the stake in the ground and say this is ready as is (outside of code changes to address reviewer comments). Thanks!

rmclaren:
No problem Steve. Trying to figure out how to convert a draft PR back to a regular PR using the GitHub GUI...

rmclaren:
@srherbener OK I figured it out. You have to press the "Ready for Review" button at the bottom.

srherbener (commented):
In general I think the code looks great, and this is a really nice addition to the converters code. Thanks for all your hard work to create this!

I have a few minor comments, plus why are there changes in these source directories:
- src/gnssro
- src/lib-python
- src/ncar-bufr2nc-fortran

It seems that these are unrelated to the bufr converter. Perhaps there is a sync issue with the develop branch?

src/bufr/BufrParser/Query/Query.h
```diff
@@ -0,0 +1,128 @@
...
+        /// we execute a query.
+        /// \note Will be refactored to use the SubsetTable object.
+        struct Target {
+            std::string name;
+            std::string queryStr;
+            bool isString;
+            std::vector<int> seqPath;
+            std::vector<int> nodeIds;
+            std::vector<std::string> dimPaths;
+            std::vector<int> exportDimIdxs;
```
nicholasesposito, line 33: I assume this is just the index of the dimenision if there are more than 1 dimension?

src/bufr/BufrParser/Query/QueryParser.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+        if (working_str.substr(0, 1) == "[") {
+            if (working_str.substr(working_str.length() - 1, 1) != "]") {
+                std::cerr << "Query Parser: multi query is lacking closing brackets." << std::endl;
+                exit(1);
+            }
+
+            working_str = working_str.substr(1, working_str.length() - 2);
+
+            std::vector<size_t> comma_positions;
```
nicholasesposito, line 31: What's the benefit of using a "vectors" in a lot of these?

src/bufr/BufrParser/Query/QueryParser.cpp
```diff
@@ -0,0 +1,127 @@
...
+        subset = query.substr(0, slashPositions[0]);
+
+        std::vector<std::string> mnemonicStrings(slashPositions.size());
+
+        // Capture the sequence mnemonic strings
+        for (size_t mnemonicIdx = 0; mnemonicIdx < mnemonicStrings.size() - 1; ++mnemonicIdx)
+        {
+            mnemonicStrings[mnemonicIdx] =
+                    query.substr(slashPositions[mnemonicIdx] + 1,
+                                 slashPositions[mnemonicIdx + 1] - slashPositions[mnemonicIdx] - 1);
```
nicholasesposito, line 97: What is trying to be done here?

src/bufr/BufrParser/Query/QueryParser.cpp
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        // Get the last element
+        std::string lastElement = query.substr(slashPositions[slashPositions.size() - 1] + 1);
```
nicholasesposito, line 101: Why do you need the +1 at the end? Because it starts at 0?

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -7,137 +7,141 @@
 
 #include "BufrParser.h"
 
-#include <map>
 #include <ostream>
 #include <iostream>
+#include <chrono>  // NOLINT
```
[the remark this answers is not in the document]
rmclaren: Therewide code norm check complains "<chrono> is an unapproved C++11 header."

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -7,137 +7,141 @@
...
-        return exportData(collectors.finalize());
+        std::cout << "Exporting Data" << std::endl;
+        auto exportedData = exportData(srcData);
+
+        auto timeElapsed = std::chrono::steady_clock::now() - startTime;
+        auto timeElapsedDuration = std::chrono::duration_cast<std::chrono::milliseconds>
+                (timeElapsed);
+        std::cout << "Finished "
+                  << "[" << timeElapsedDuration.count() / 1000.0 << "s]"
+                  << std::endl;
```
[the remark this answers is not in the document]
rmclaren: Unfortunately casting to seconds rounds the result to the nearest second. The purpose of this timer is to aid in performance monitoring. For many samples we run we care to know the result with a higher granularity.

src/bufr/BufrParser/Query/Query.h
rmclaren: exportDimIdxs keeps track of which dimensions count. This is because "binary" repeats don't count (so they are excluded from this list. I should figure out a cleaner way to do this...

src/bufr/BufrParser/Query/QueryParser.cpp
rmclaren: Vectors are equivalent to Lists in python. You use them basically the same way. So every time you see a vector its just a List.

src/bufr/BufrParser/Query/QueryParser.cpp
rmclaren: This line just splits the memonics in a query string into a list of strings. Substr first argument is position and the second position is the length of the string. Slash positions holds the positions of the "/" characters.

src/bufr/BufrParser/Query/QueryParser.cpp
rmclaren: +1 to skip the slash "/" character. Slash positions holds the positions of the slashes in the query string.

src/bufr/BufrParser/Query/Query.h
```diff
@@ -0,0 +1,128 @@
...
+         private:
+            std::vector<T> data_;
+            size_t offset_;
+        };
+
+        /// \brief Masks used to make the processing of BUFR data more efficient. The aim is to skip
+        /// branches without data we care about.
+        struct ProcessingMasks {
+            std::vector<bool> valueNodeMask;
+            std::vector<bool> pathNodeMask;
```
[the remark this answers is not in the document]
rmclaren: Thanks. Think its ok.

src/bufr/BufrParser/Query/DataProvider.cpp
```diff
@@ -0,0 +1,99 @@
...
+                if      (typ == Subset)            typ_[wordIdx] = Typ::Subset;
+                else if (typ == DelayedRep)        typ_[wordIdx] = Typ::DelayedRep;
+                else if (typ == FixedRep)          typ_[wordIdx] = Typ::FixedRep;
+                else if (typ == DelayedRepStacked) typ_[wordIdx] = Typ::DelayedRepStacked;
+                else if (typ == DelayedBinary)     typ_[wordIdx] = Typ::DelayedBinary;
+                else if (typ == Sequence)          typ_[wordIdx] = Typ::Sequence;
+                else if (typ == Repeat)            typ_[wordIdx] = Typ::Repeat;
+                else if (typ == StackedRepeat)     typ_[wordIdx] = Typ::StackedRepeat;
+                else if (typ == Number)            typ_[wordIdx] = Typ::Number;
+                else if (typ == Character)         typ_[wordIdx] = Typ::Character;
```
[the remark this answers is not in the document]
rmclaren: As before: Current path returns is guaranteed to be very small (probably no more than about 5). But I can preallocate some memory with **reserve** if you like.

src/bufr/BufrParser/Query/Query.cpp
```diff
@@ -0,0 +1,389 @@
...
+        auto dataTable = __details::OffsetArray<NodeData>(
+                dataProvider_.getInode(),
+                dataProvider_.getIsc(dataProvider_.getInode()));
+
+        for (size_t dataCursor = 1; dataCursor <= dataProvider_.getNVal(); ++dataCursor) {
+            int nodeIdx = dataProvider_.getInv(dataCursor);
+
+            if (masks->valueNodeMask[nodeIdx]) {
+                auto &values = dataTable[nodeIdx].values;
+                values.push_back(dataProvider_.getVal(dataCursor));
```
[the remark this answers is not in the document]
rmclaren: The number of values and counts in the NodeData can be hard to know ahead of time unfortunately which is why I don't attempt to preallocate this memory. Most of the time the size of these arrays is very small, so if you guess a number you usually allocate too much degrading performance. The step of re-organizing the data would be unnecessary if we had a better implementation of a BUFR file reader.

src/bufr/BufrParser/Query/Query.cpp
```diff
@@ -0,0 +1,389 @@
...
+            }
+
+            if (masks->pathNodeMask[nodeIdx] && isQueryNode(nodeIdx)) {
+                if (dataProvider_.getTyp(nodeIdx) == Typ::DelayedBinary &&
+                    dataProvider_.getVal(dataCursor) == 0) {
+                    // Ignore the node if it is a delayed binary and the value is 0
+                } else {
+                    currentPath.push_back(nodeIdx);
+                    const auto tmpReturnNodeIdx = dataProvider_.getLink(nodeIdx);
+                    currentPathReturns.push_back(tmpReturnNodeIdx);
```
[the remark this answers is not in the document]
rmclaren: Same as the last one.

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
```diff
@@ -85,65 +85,98 @@ namespace Ingester
...
                 tm.tm_isdst = 0;
 
-                if (!secondKey_.empty())
+                if (!minuteQuery_.empty())
                 {
-                    if (map.at(secondKey_)(idx) >= 0 && map.at(secondKey_)(idx) < 60)
+                    minutes =
+                        static_cast<int>(map.at(getExportKey(ConfKeys::Minute))->getAsFloat(idx));
+
+                    if (minutes >= 0 && minutes < 60)
```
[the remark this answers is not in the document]
rmclaren: @gthompsnJCSDA Could you provide some insight?

src/bufr/BufrParser/Exports/Variables/QueryVariable.cpp
```diff
@@ -0,0 +1,63 @@
+/*
+ * (C) Copyright 2020 NOAA/NWS/NCEP/EMC
```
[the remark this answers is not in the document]
rmclaren: @srherbener Not sure what Ryans intentions were. The file in question was created in 2020 and hasn't changed that much.

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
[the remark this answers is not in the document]
rmclaren: In large part I think this is to exclude the missing value, values.

src/bufr/BufrParser/Query/File.cpp
```diff
@@ -0,0 +1,98 @@
...
+            fileUnitTable2_(nextFileUnit()),
+            isWmoFormat_(isWmoFormat),
+            wmoTablePath_(wmoTablePath)
+    {
+        open();
+    }
+
+    void File::open()
+    {
+        open_f(fileUnit_, filename_.c_str());
```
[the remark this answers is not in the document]
rmclaren: Unfortunately doing this would mean I have to expose another Fortran function in NCEPLIB-bufr so that I could enquire about the status of a file. So I might just write an issue to address this in the future.

rmclaren:
there is a test for test_iodaconv_bufr_adpupa_groupby.yaml, except that YAML file and supporting files no longer exist. I will delete that test.

rmclaren:
All iodaconv_bufr_* tests now pass.

src/bufr/BufrParser/BufrParser.cpp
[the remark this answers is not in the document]
srherbener: Thanks for the explanation. Somehow (probably through the use of CPPLINT.cfg files) we don't need to do this in the ioda repo. But this fine.

src/bufr/BufrParser/Query/Query.cpp
```diff
@@ -0,0 +1,389 @@
...
+                    returnNodeIdx = currentPathReturns[lastNonZeroReturnIdx];
+                }
+            }
+
+            if (masks->pathNodeMask[nodeIdx] && isQueryNode(nodeIdx)) {
+                if (dataProvider_.getTyp(nodeIdx) == Typ::DelayedBinary &&
+                    dataProvider_.getVal(dataCursor) == 0) {
+                    // Ignore the node if it is a delayed binary and the value is 0
+                } else {
+                    currentPath.push_back(nodeIdx);
```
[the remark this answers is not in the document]
srherbener: I don't think there is a need as long as the path is limited to a small number as you say. It's okay to leave this as is.

srherbener (approved):
Thanks for the responses to my comments. Looks good now.

src/ncar-bufr2nc-fortran/define_mod.f90
```diff
@@ -19,8 +19,7 @@ module define_mod
 integer(i_kind), parameter :: nvar_met          = 6
 integer(i_kind), parameter :: nvar_info         = 9  ! number of metadata
 integer(i_kind), parameter :: nsen_info         = 7  ! number of sensor metadata
-integer(i_kind), parameter :: ninst             = 12
-!integer(i_kind), parameter :: ninst             = 13 ! including airs
```
BenjaminRuston, line 23: was not aware you were looking at these routines as well. @rmclaren thank you for trying to keep these in sync were these tested?  There was an independent fork of these decoders but would be nice to make this set the standard ones

PatNichols:
rmclauren There are a few errors when compiling the code:

Building CXX object iodaconv/src/bufr/CMakeFiles/ingester.dir/BufrParser/Query/DataProvider.cpp.o
--
192 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:43:13: error: use of undeclared identifier 'get_isc_f'
193 | get_isc_f(&intPtr, &size);
194 | ^
195 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:46:13: error: use of undeclared identifier 'get_link_f'
196 | get_link_f(&intPtr, &size);
197 | ^
198 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:49:13: error: use of undeclared identifier 'get_itp_f'
199 | get_itp_f(&intPtr, &size);
200 | ^
201 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:52:13: error: use of undeclared identifier 'get_typ_f'
202 | get_typ_f(&charPtr, &strLen, &size);
203 | ^
204 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:56:35: error: no template named 'unordered_map' in namespace 'std'
205 | static const std::unordered_map<std::string, Typ> TypMap =
206 | ~~~~~^
207 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:72:13: error: use of undeclared identifier 'get_tag_f'
208 | get_tag_f(&charPtr, &strLen, &size);
209 | ^
210 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:80:13: error: use of undeclared identifier 'get_jmpb_f'
211 | get_jmpb_f(&intPtr, &size);
212 | ^
213 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:84:9: error: use of undeclared identifier 'get_inode_f'
214 | get_inode_f(bufrLoc, &inode_);
215 | ^
216 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:85:9: error: use of undeclared identifier 'get_nval_f'
217 | get_nval_f(bufrLoc, &nval_);
218 | ^
219 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:87:9: error: use of undeclared identifier 'get_val_f'
220 | get_val_f(bufrLoc, &dataPtr, &size);
221 | ^
222 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:90:9: error: use of undeclared identifier 'get_inv_f'
223 | get_inv_f(bufrLoc, &intPtr, &size);
224 | ^
225 | /jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider.cpp:98:9: error: use of undeclared identifier 'delete_table_data_f'
226 | delete_table_data_f();
227 | ^

These need to be fixed before we merge (maybe a header file that wasn't included?) alos there are also some warnings due to technical points.

rmclaren:
@PatNichols The errors mean you are trying to compile against an old version of NCEPLIB-bufr. You need to be either on develop or at least tag version 11.7 for these functions to be present.

rmclaren:
@PatNichols I also noticed it didn't find std::unordered_map with your compiler for some reason. I went ahead and added an include to the relevant header file to that cpp file (please update).

PatNichols:
We think (SteveH,Ben and myself) that perhaps the wrong version of ncep-bufrlib is in the aws test container. We are going to investigate that. On my MAC it seems to build when the missing header file was included. Can you confirm that it is version of 11.7 ?

rmclaren:
@PatNichols Yes, I believe the exact name for the NCEPLIB-bufr tag is bufr_v11.7.0.

srherbener:
I just submitted the PRs in spack-stack (https://github.com/NOAA-EMC/spack-stack/pull/322) and spack (https://github.com/NOAA-EMC/spack/pull/164) that upgrade NCEPLIBS-bufr to version 11.7.1.

src/ncar-bufr2nc-fortran/define_mod.f90
rmclaren: These changes were unintentional, I',m not sure how they happened. I went ahead and reverted this change and other changes like this.

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
```diff
@@ -85,65 +86,98 @@ namespace Ingester
...
+        std::vector<int64_t> timeOffsets;
+        timeOffsets.reserve(map.at(getExportKey(ConfKeys::Year))->size());
+
+        for (unsigned int idx = 0; idx < map.at(getExportKey(ConfKeys::Year))->size(); idx++)
         {
+            int year = static_cast<int>(map.at(getExportKey(ConfKeys::Year))->getAsFloat(idx));
+            int month = static_cast<int>(map.at(getExportKey(ConfKeys::Month))->getAsFloat(idx));
+            int day = static_cast<int>(map.at(getExportKey(ConfKeys::Day))->getAsFloat(idx));
+            int hour = static_cast<int>(map.at(getExportKey(ConfKeys::Hour))->getAsFloat(idx));
+            int minutes = 0;
```
CoryMartin-NOAA, line 98: Why are mins/secs different here? Not 100% sure what this is, but if it is related to offset from UTC to localtime, some timezones are 30 mins off, not on the hour

src/bufr/BufrParser/Exports/Variables/Variable.h
```diff
@@ -11,19 +11,52 @@
...
+        inline void initQueryMap() { queryList_ = makeQueryList(); }
+
+        /// \brief Make a map of name and queries
+        virtual QueryList makeQueryList() const = 0;
+
+     private:
+        /// \brief Name used to export this variable
+        std::string exportName_;
+
+        /// \brief The query map for all the queries this vairable needs
```
CoryMartin-NOAA, line 58: typo

src/bufr/BufrParser/Query/File.h
```diff
@@ -0,0 +1,57 @@
...
+    class File
+    {
+     public:
+        File() = delete;
+
+        File(const std::string& filename,
+             bool isWmoFormat = false,
+             const std::string& wmoTablePath = "");
+
+        /// \brief Execute the queryies given in the query set over the BUFR file and accumulate the
```
CoryMartin-NOAA, line 28: typo

src/bufr/DataObject.cpp
```diff
@@ -0,0 +1,48 @@
+/*
+ * (C) Copyright 2021 NOAA/NWS/NCEP/EMC
```
CoryMartin-NOAA, line 2: 2022

src/bufr/DataObject.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Ingester
+{
+    std::shared_ptr<DataObjectBase>
+        DataObjectBase::fromResult(const std::shared_ptr<bufr::ResultBase>& resultBase,
+                                   const std::string& query)
+    {
+        std::shared_ptr<DataObjectBase> dataObject = nullptr;
+
+        if (auto result = std::dynamic_pointer_cast<bufr::Result<float>>(resultBase))
```
CoryMartin-NOAA, line 23: Only float and str? What about int?

src/bufr/DataObject.h
```diff
@@ -0,0 +1,368 @@
+/*
+ * (C) Copyright 2020 NOAA/NWS/NCEP/EMC
```
CoryMartin-NOAA, line 2: 2022

src/bufr/DataObject.h
```diff
@@ -0,0 +1,368 @@
...
+
+        /// \brief Get the data at the Location as an string.
+        /// \return String data.
+        virtual std::string getAsString(const Location& loc) const = 0;
+
+        /// \brief Get the size of the data
+        /// \return Data size.
+        virtual size_t size() const = 0;
+
+        /// \brief Makes an ioda::Variable and ads it to the given ioda::ObsGroup
```
CoryMartin-NOAA, line 82: typo: adds

src/bufr/DataObject.h
```diff
@@ -0,0 +1,368 @@
...
+        /// \param obsGroup Obsgroup were to add the variable
```
CoryMartin-NOAA, line 83: were -> where?

src/bufr/DataObject.h
```diff
@@ -0,0 +1,368 @@
...
+                   const Dimensions& dimensions,
+                   const std::string& query,
+                   const std::vector<std::string>& dimPaths) :
+            DataObjectBase(field_name, group_by_field_name, dimensions, query, dimPaths),
+            data_(data)
+        {};
+
+        ~DataObject() = default;
+
+        /// \brief Makes an ioda::Variable and ads it to the given ioda::ObsGroup
```
CoryMartin-NOAA, line 130: ad and were again

src/bufr/DataObject.h
```diff
@@ -0,0 +1,368 @@
...
+                                      int compressionLevel) const final
+        {
+            auto params = makeCreationParams(chunks, compressionLevel);
+            auto var = obsGroup.vars.createWithScales<T>(name, dimensions, params);
+            var.write(data_);
+            return var;
+        };
+
+        /// \brief Print data to stdout for debug purposes.
+        void print(std::ostream &out) const final
```
CoryMartin-NOAA, line 149: Is this needed still? Should it use OOPS if so?

src/bufr/DataObject.h
```diff
@@ -0,0 +1,368 @@
...
+        /// \return Float data.
+        float getAsFloat(const Location& loc) const final { return _getAsFloat(loc); }
+
+        /// \brief Get the data at the location as a string.
+        /// \return String data.
+        std::string getAsString(const Location& loc) const final { return _getAsString(loc); }
+
+        /// \brief Get the data at the location as a float.
+        /// \return Float data.
+        float getAsFloat(size_t idx) const final { return _getAsFloat(idx); }
```
CoryMartin-NOAA, line 207: Index vs location? Why is there a second float one?

src/bufr/IodaEncoder/IodaDescription.cpp
```diff
@@ -83,21 +83,34 @@ namespace Ingester
...
+                {
+                    dim.paths = dimConf.getStringVector(ConfKeys::Dimension::Paths);
+                }
+                else if (dimConf.has(ConfKeys::Dimension::Path))
+                {
+                    dim.paths = {dimConf.getString(ConfKeys::Dimension::Path)};
+                }
+                else
+                {
+                    throw eckit::BadParameter(R"(ioda::dimensions section must have either "path" or "paths".)");
```
CoryMartin-NOAA, line 109: R? I guess that allows for the " ?

test/CMakeLists.txt
```diff
@@ -849,14 +893,6 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_viirs_aod
                           -n 0.0"
                           viirs_aod.nc ${IODA_CONV_COMP_TOL_ZERO})
 
-ecbuild_add_test( TARGET  test_${PROJECT_NAME}_satbias_viirs_aod
```
CoryMartin-NOAA, line 852: why are this one and the next one removed?

test/CMakeLists.txt
```diff
@@ -927,6 +953,17 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_airnow
                           -o testrun/airnow_2020081306.nc"
                           airnow_2020081306.nc ${IODA_CONV_COMP_TOL_ZERO})
 
+ecbuild_add_test( TARGET  test_iodaconv_generic_gnssro_bufr
```
CoryMartin-NOAA, line 956: Why is this added?

test/CMakeLists.txt
```diff
@@ -984,179 +1021,238 @@ if(iodaconv_bufr_ENABLED)
...
-                            netcdf
-                    "${CMAKE_BINARY_DIR}/bin/bufr2ioda.x testinput/vadwinds_wmoBUFR2ioda.yaml"
-                    vadwinds_wmo_multi.nc ${IODA_CONV_COMP_TOL_ZERO}
+                    COMMAND ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                    ARGS    netcdf
+                    "${CMAKE_BINARY_DIR}/bin/bufr2ioda.x testinput/bufr_ncep_sevcsr.yaml"
+                    gdas.t00z.sevcsr.tm00.nc ${IODA_CONV_COMP_TOL_ZERO}
                     DEPENDS bufr2ioda.x )
+
+#  ecbuild_add_test( TARGET  test_iodaconv_bufr_aircar
```
CoryMartin-NOAA, line 1186: Why are all these commented out?

CoryMartin-NOAA (commented):
A lot of stuff here... great work! A few comments from my quick review

srherbener:
I've found the problem compiling on Clang which is that the C++ standard is not being set, ie the `-std=gnu++14` compiler option is not being used. The rest of JEDI is currently using the C++14 standard..

I'm working on a PR that I will submit soon with this fix.

I've also noticed that a number of little bug fixes and changes keep creeping into the `feature/query_cxx` branch.

My recommendation is to reset the reviews (which I've done for me) because of the creeping changes, and only allow changes that are necessary to get the ctests to pass. Please put everything not related to the ctests into PRs based on `feature/query_cxx`. Doing this will allow us to "freeze" the `feature/query_cxx` branch and stay focused on getting it merged in. After the merge, we can then proceed with appropriate bug fixes and feature as necessary.

Adding the C++14 standard, and using NCEPLIBS-bufr version 11.7.0 on my Mac resulted in a successful build. I suspect after adding in the C++14 standard, that any build issues in the CI tests will be due to an older version of the NCEPLIBS-bufr in the test containers. Work is in progress to get the test containers updated.

Once we get the CI build functioning again, then we can tackle any test failures and get this PR merged in.

srherbener:
Also we need to reduce the number of reviewers - I'll do that now.

rmclaren:
@srherbener Sorry Steve, I'll avoid merging in any more bug fix type changes and focus just on the PR comments.

srherbener:
> @srherbener Sorry Steve, I'll avoid merging in any more bug fix type changes and focus just on the PR comments.

No worries @rmclaren. All the hard work you've been doing to make the major contribution is very much appreciated!

rmclaren:
@srherbener Thanks for your time and figuring out the `-std=gnu++14` flag issue. I wasn't really aware it existed...

YoulongXia-NOAA:
> Also we need to reduce the number of reviewers - I'll do that now.

This will help to speed up the merging procedure. Good suggestion, thank you @srherbener

src/bufr/BufrParser/BufrParser.cpp
```diff
@@ -7,137 +7,142 @@
...
     }
 
-    std::shared_ptr <DataContainer> BufrParser::parse(const size_t maxMsgsToParse)
+    std::shared_ptr<DataContainer> BufrParser::parse(const size_t maxMsgsToParse)
     {
-        const unsigned int SubsetStringLength = 25;
+        auto startTime = std::chrono::steady_clock::now();
 
-        if (fortranFileId_ <= 10)
+        oops::Log::info() << "Start" << std::endl;
```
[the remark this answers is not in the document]
rmclaren: Removing this message as its kinda pointless anyways.

src/bufr/DataObject.cpp
rmclaren: **query_cxx** only supports float and string. The other branch, **query_cxx_types** (merging later) actually refactors a lot of this (this function disappears) and has the flexibility to support basically any type.

src/bufr/DataObject.h
rmclaren: The objective was to print the contents of the data object out for debug purposes. I don't think its currently used, just a convenience method for debugging.

srherbener:
> I've found the problem compiling on Clang which is that the C++ standard is not being set, ie the `-std=gnu++14` compiler option is not being used. The rest of JEDI is currently using the C++14 standard..
> 
> I'm working on a PR that I will submit soon with this fix.
> 
> I've also noticed that a number of little bug fixes and changes keep creeping into the `feature/query_cxx` branch.
> 
> My recommendation is to reset the reviews (which I've done for me) because of the creeping changes, and only allow changes that are necessary to get the ctests to pass. Please put everything not related to the ctests into PRs based on `feature/query_cxx`. Doing this will allow us to "freeze" the `feature/query_cxx` branch and stay focused on getting it merged in. After the merge, we can then proceed with appropriate bug fixes and feature as necessary.
> 
> Adding the C++14 standard, and using NCEPLIBS-bufr version 11.7.0 on my Mac resulted in a successful build. I suspect after adding in the C++14 standard, that any build issues in the CI tests will be due to an older version of the NCEPLIBS-bufr in the test containers. Work is in progress to get the test containers updated.
> 
> Once we get the CI build functioning again, then we can tackle any test failures and get this PR merged in.

Sorry, I should have said let's limit the changes to `feature/query_cxx` to those that fix ctests, address reviewer comments, and get the build to complete successfully.

climbfuji:
Please note that we started rolling out the new bufr library 11.7.1 on the various spack-stack systems. You can monitor the progress here, and there are also some basic instructions in the issue on how to load the new module: https://github.com/NOAA-EMC/spack-stack/issues/328

src/bufr/IodaEncoder/IodaDescription.cpp
rmclaren: R stands for Raw String. Basically special characters are ignored (not processed) between "( and )".

src/bufr/DataObject.h
rmclaren: That's a really good question. 

So data object is able to store data of any dimension. It stores all data internally in a 1d vector.

`getAsFloat(const Location& loc)`: The Location is a vector<int> which points to the coordinate of the data point. So if I have 2 dimensional data the location {2, 4} would get me the value at that coordinate. This version of `getAsFloat` computes the correct offset for the data point (inside that 1d vector) based off the coordinate that is given and what it knows about the data's dimensional properties (size of each dimension).

`getAsFloat(size_t idx)`: This function basically allows you to get a data point based on the idx in the 1d array (it allows for direct access). In general it is not as safe, but computationally a lot cheaper than the previous example. I added it to get date time data making the assumption that this data is always 1d... In retrospect this was probably a stupid thing to do, however. We might want to refactor some of this code later and remove this function.

rmclaren: I updated some of the doc strings to make this a little clearer

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
rmclaren: Minutes and seconds are treated differently here because they are optional values where month, day, hour are required. There are BUFR files that don't include information about seconds or minutes... The timestamps aren't always that granular.

src/bufr/DataObject.h
rmclaren: the argument lets you print the data to any stream you want. So you could direct it to the oops log stream if you desired.

test/CMakeLists.txt
rmclaren: Was moved but probably to the wrong section (was in bufr section but is not part of that effort as it uses python for its processing. Dunno were it should go.

test/CMakeLists.txt
CoryMartin-NOAA: Yeah I'm with you, 'unit tests' should only be testing code, for science eval, these things should be done offline. JEDI as a whole has way too many unit tests already.

test/CMakeLists.txt
CoryMartin-NOAA: yeah I agree this is getting to be a large file

test/CMakeLists.txt
srherbener: I second the concern about test organization. Unit testing is for code coverage (not science verification) and should consist of a set of tiny, fast executing tests (since we run unit tests after every commit push). 

A set of tests for science verification should be outside the set for unit testing. There can be many small or large tests, but we would then run these nightly or on the weekend so that we can accommodate longer runtimes. I think roughly this is what test tier2 is for. @mer-a-o would you agree?

We also need a "system integration" set of tests. This would be more in the vein of running experiments in EWOK and other workflow systems that our partners are using. Ie, validating that all the pieces play together as intended. This can cover science verification, but it also serves the purpose of making sure the interfaces are working properly and for flushing out (and fixing) memory and runtime performance issues. I think this is roughly what test tier3 is for.

test/CMakeLists.txt
rmclaren: I wrote a python script that helped me find tests that were dropped by accident and added them back in.

```python
import re

path1 = "/Users/rmclaren/Work/ioda-bundle/iodaconv/test/CMakeLists.txt"
path2 = "/Users/rmclaren/Temp/develop_CMakeLists.txt"

dict1 = {}
dict2 = {}

expr = re.compile(".*TARGET\s+(?P<testFile>[^\s]+)")

with open(path1) as f:
    for line in f.readlines():
        if expr.match(line):
            m = expr.search( line )
            dict1[m.group('testFile')] = ''
            
with open(path2) as f:
    for line in f.readlines():
        if expr.match(line):
            m = expr.search( line )
            dict2[m.group('testFile')] = ''

print("Tests added.")
for key in dict1.keys():
    if not key in dict2.keys():
        print(key)

print("Missing Tests.")
for key in dict2.keys():
    if not key in dict1.keys():
        print(key)
```

test/CMakeLists.txt
rmclaren: Moved this test back to where it was before as I don't think it belongs in this section.

PatNichols (approved):

CoryMartin-NOAA (approved):
Thanks Ron for a lot of hard work!!

nicholasesposito (approved):

rmclaren:
@ShastriPaturi No problem :)

rmclaren:
Fixed an issue that was introduced with a code review change. Basically the exceptions were being sliced and we were loosing the descriptions, so the user has no idea what happened...

srherbener (approved):
Looks good @rmclaren! Thank you for your patience while the test containers get prepared, and thank you for all your hard work to get the bufr reader implemented!

srherbener:
I just noticed that we were missing a companion PR in ioda with its feature/query_cxx branch. So, I submitted that PR: jcsda-internal/ioda/pull/779 and updated the comments to note that these PRs need to be merged together.

I think that should cover what is needed for this PR, but can anyone think of anything else that needs to be prepared? Thanks!

srherbener:
On my Mac using NCEPLIBS-bufr v11.7.0, and using the feature/query_cxx branch from ioda, I see the build complete successfully, and all ioda-converter ctests pass except for these three:
```
603:test_iodaconv_ears_mhs_conv
605:test_iodaconv_ears_amsua_conv
606:test_iodaconv_iasi_conv
```

Can someone help fix these? The 3 test failures are all bufr2nc_fortran tests, which is the NCAR bufr converter. Source code is in `src/ncar-bufr2nc-fortran`.

In addition to the Mac, you should be able to reproduce these results on Orion.

Thanks!

rmclaren:
@emilyhcliu Told me to comment out the offending tests for now, so I did.

emilyhcliu (approved):

srherbener:
> @emilyhcliu Told me to comment out the offending tests for now, so I did.

Sounds good. I'm seeing all tests pass on my Mac now! Thanks for fixing this.

srherbener:
The Intel build had these 2 test failure, and the CDASH url didn't get built correctly (see below for error messages).

```
...
99% tests passed, 2 tests failed out of 744
--
938 |  
939 | Label Time Summary:
940 | executable           =  15.40 sec*proc (127 tests)
941 | fortran              =   0.17 sec*proc (1 test)
942 | ioda                 =  43.87 sec*proc (245 tests)
943 | iodaconv             =  92.69 sec*proc (235 tests)
944 | iodaconv_validate    =  10.78 sec*proc (125 tests)
945 | mpi                  =  26.69 sec*proc (44 tests)
946 | oops                 =  74.44 sec*proc (262 tests)
947 | openmp               =  21.31 sec*proc (51 tests)
948 | script               = 188.06 sec*proc (505 tests)
949 |  
950 | Total Test time (real) = 211.80 sec
951 |  
952 | The following tests FAILED:
953 | 566 - test_iodaconv_tropomi_no2_tropo (Failed)
954 | 569 - test_iodaconv_aeronet_aod (Failed)
955 | Errors while running CTest
...

[Container] 2022/09/21 15:23:09 Running command if [ "$BUILD_STATUS" = "1" ]; then echo "Build & tests passed, find cdash url"; url=$(bash /jcsda/ioda-bundle/iodaconv/CI/cdash-url.sh /home/nonroot/iodaconv/Testing $CDASH_URL); echo $url; echo ${url} > /jcsda/artifacts/cdash-url.txt; cat /jcsda/artifacts/cdash-url.txt; else echo "Build failed"; fi
--
983 | Build & tests passed, find cdash url
984 | head: cannot open '/home/nonroot/iodaconv/Testing/TAG' for reading: No such file or directory
985 | cat: /home/nonroot/iodaconv/Testing//Done.xml: No such file or directory
986 | http://cdash.jcsda.org:8080/buildSummary.php?buildid=
987 | http://cdash.jcsda.org:8080/buildSummary.php?buildid=
```

All tests pass on my Mac.

I think we are down to the same failures (one less) that we saw with #1056. I believe that these are not related and came about because they are in develop. I am going to go ahead and merge, and we can fix the test failures with new PRs.

srherbener:
Thank you everyone for all of your hard work to get this PR ready for merging. You help is greatly appreciated!

rmclaren:
@srherbener Thanks so much Steve!

YoulongXia-NOAA:
Thank you @srherbener @rmclaren, big achievement, congratulations!

nicholasesposito:
Congrats and great job!

PR #1013, opened 2022-07-26
Fix for test failure (test_iodaconv_ghcn_snod) with spack-stack build

srherbener:
## Description

This PR fixes an issue that surfaced with the `test_iodaconv_ghcn_snod` test when using a spack-stack build. spack-stack builds the pandas python package using the python installation that is found when spack-stack install is running. The `test_iodaconv_ghcn_snod` test runs a python script `ghch_snod2ioda.py` which uses the pandas read_csv command to read a text file containing station names. The read_csv is called with '\n' as a delimiter, but this was failing on my Mac (and possibly other platforms). This issue is that read_csv with it's built-in parser does not recognize '\n' as a delimiter, so it switches to the 'python' parser which can vary depending on the version that is found when spack-stack runs.

I'm not expert with pandas, so there might be a much better way to solve this. Please treat this PR as a proposal, and feel free to suggest a better way to fix this. I changed the delimiter from '\n' to '\t' (tab) which is recognized by the read_csv built-in parser and appears to work (since the intent appears to be to get read_csv to treat the entire line as a single entry in the output dataframe.

### Issue(s) addressed

None

The `test_iodaconv_ghcn_snod` test passes on the Mac (and all platforms).

## Dependencies

None

## Impact

None

### DIFF
--- a/src/land/ghcn_snod2ioda.py
+++ b/src/land/ghcn_snod2ioda.py
@@ -102,7 +102,7 @@ class ghcn(object):
         df30 = df30[df30["DATETIME"] == parse(startdate).date()]
         # Read station files
         cols = ["ID", "LATITUDE", "LONGITUDE", "ELEVATION", "STATE", "NAME", "GSN_FLAG", "HCNCRN_FLAG", "WMO_ID"]
-        df10all = pd.read_csv(self.fixfile, header=None, sep='\t')
+        df10all = pd.read_csv(self.fixfile, header=None, sep='\r\n')
         df10all = df10all[0].str.split('\\s+', expand=True)
         df10 = df10all.iloc[:, 0:4]
         sub_cols = {0: "ID", 1: "LATITUDE", 2: "LONGITUDE", 3: "ELEVATION"}
--- a/src/land/ghcn_snod2ioda.py
+++ b/src/land/ghcn_snod2ioda.py
@@ -102,7 +102,7 @@ class ghcn(object):
         df30 = df30[df30["DATETIME"] == parse(startdate).date()]
         # Read station files
         cols = ["ID", "LATITUDE", "LONGITUDE", "ELEVATION", "STATE", "NAME", "GSN_FLAG", "HCNCRN_FLAG", "WMO_ID"]
-        df10all = pd.read_csv(self.fixfile, header=None, sep='\n')
+        df10all = pd.read_csv(self.fixfile, header=None, sep='\t')
         df10all = df10all[0].str.split('\\s+', expand=True)
         df10 = df10all.iloc[:, 0:4]
         sub_cols = {0: "ID", 1: "LATITUDE", 2: "LONGITUDE", 3: "ELEVATION"}


CoryMartin-NOAA:
This is odd... `\n` is newline, no? and `\t` is tab? Those aren't really interchangeable are they? I'll let @YoulongXia-NOAA comment as he presumably wrote this, I don't know the specifics of the input file. I can't imagine why `\n` would ever be used as a delimiter, because that would just be 'read each line'.

YoulongXia-NOAA:
@tsga, I am taking AL for today and tomorrow. I used your python code to work for ioda-converter. There is one sentence using \n and Stephen wants to change \t, and Cory thought either is correct. You have more experience about GHCN snowD csv file. Could you please see how to fix this issue and give some explanations first?
 
Click File changed will see the modification from Stephen. I will take a look after I am back from AL.

YoulongXia-NOAA:
@tsga , either is NOT correct.

srherbener:
> This is odd... `\n` is newline, no? and `\t` is tab? Those aren't really interchangeable are they? I'll let @YoulongXia-NOAA comment as he presumably wrote this, I don't know the specifics of the input file. I can't imagine why `\n` would ever be used as a delimiter, because that would just be 'read each line'.

@CoryMartin-NOAA it looks like the code where the change is made is intending to build a pandas dataframe out of a text file (not a csv file). The text file has a different number of fields as you go line-by-line so the code (I believe) is reading the entire line as one field, and then in subsequent commands splits it up in an even (same number of fields per row in the data frame) manner. newline and tab delimiters of course are not the same, however the input file does not contain tabs so in this case they have the same result (ie, the entire line is treated as one field). Using a tab instead of a newline avoids the error I saw on my MacBook with my python3 (3.9.13) installation (when using spack-stack to build pandas).

I admit that this is a hack fix, so if someone more familiar with pandas has a better way to address this I am all for that. I think the idea is to transfer a text file that contains space separated fields, where there can be a different number of fields line-by-line, into a pandas dataframe.

tsga:
@srherbener, yes that was the reason, but I only tested on Windows. 

For now (until/unless someone comes up with a better way to address it in Pandas) a quick fix would be to use  sep='\r\n'  which I think should work on Mac as well as on Windows (tested on Windows). The downside will be it will be slower and we'll get a parser warning about using regex as delimiter.

tsga (changes requested):
Change line 105 to:
df10all = pd.read_csv(self.fixfile, header=None, sep='\r\n')

srherbener:
> Change line 105 to: df10all = pd.read_csv(self.fixfile, header=None, sep='\r\n')

@tsga thank you for your review! Your suggestion works on my Mac, so I applied it.

CoryMartin-NOAA (approved):
Yes this should handle both 'return' cases on all platforms (I think)

YoulongXia-NOAA:
I tested it on Hera and it works.

src/land/ghcn_snod2ioda.py
```diff
@@ -102,7 +102,7 @@ def assignValue(colrowValue, df400):
         df30 = df30[df30["DATETIME"] == parse(startdate).date()]
         # Read station files
         cols = ["ID", "LATITUDE", "LONGITUDE", "ELEVATION", "STATE", "NAME", "GSN_FLAG", "HCNCRN_FLAG", "WMO_ID"]
-        df10all = pd.read_csv(self.fixfile, header=None, sep='\n')
+        df10all = pd.read_csv(self.fixfile, header=None, sep='\r\n')
```
YoulongXia-NOAA, line 105: Look good for me. I tested on Hera for ioda-bundle, ctest passed.

YoulongXia-NOAA (approved):

YoulongXia-NOAA:
By the way, thank you very much @tsga.

tsga (approved):

PR #1025, opened 2022-08-15
Changes in L2 trace gas retrievals for column operator generalization

jeromebarre:
This is a first step for the column operator generalization.

This generalization aim to compute H(x) for any column product:
total columns
partial columns
with or without averaging kernel and with or without apriori contribution

for this the IODA structure needs to be changed a little to accomodate for this:

- adding upper most vertice in pressure coordinate. This was assumed to be top of the atmosphere (TOA) in UFO. This is not always the case, e.g. partial columns without averaging kernels are passed and needs H(x) from 800hPa to 200hPa.
- moving the AMF tropospheric scaling for the AK in DOAS retrievals to the ioda-converter. A proper AMF scaling in UFO should ideally use the model information and not observation information @christophkeller
- (optional) adding a pre-filtering capability in the ioda converter in the tropomi converter as it was done in the mopitt converter.

A UFO PR will be issued soon to account for those changes.

### DIFF
--- a/src/compo/tropomi_no2_nc2ioda.py
+++ b/src/compo/tropomi_no2_nc2ioda.py
@@ -84,6 +84,9 @@ class tropomi(object):
             qa_value = ncd.groups['PRODUCT'].variables['qa_value'][:]  # 2D
             times = np.empty_like(qa_value, dtype=object)
             qa_value = qa_value.ravel()
+            # adding ability to pre filter the data using the qa value
+            # documentation recommends using > 0.75
+            flg = qa_value > 0.75
             qc_flag = ncd.groups['PRODUCT'].groups['SUPPORT_DATA'].groups['DETAILED_RESULTS']\
                 .variables['processing_quality_flags'][:]
             qc_flag = qc_flag.ravel().astype('int32')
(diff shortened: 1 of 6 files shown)


src/compo/tropomi_no2_nc2ioda.py
```diff
@@ -180,7 +196,7 @@ def main():
     # get command line arguments
     parser = argparse.ArgumentParser(
         description=(
-            'Reads TROPOMI NO2 netCDF files provided by NESDIS'
+            'Reads TROPOMI NO2 netCDF files: official Copernicus product'
```
CoryMartin-NOAA, line 199: good catch :-)

src/compo/tropomi_no2_nc2ioda.py
```diff
@@ -60,110 +52,134 @@ def __init__(self, filenames):
...
-        for iodavar in ['nitrogen_dioxide_in_tropospheric_column', 'nitrogen_dioxide_in_total_column']:
-            self.varDict[iodavar]['valKey'] = iodavar, iconv.OvalName()
-            self.varDict[iodavar]['errKey'] = iodavar, iconv.OerrName()
-            self.varDict[iodavar]['qcKey'] = iodavar, iconv.OqcName()
-            self.varAttrs[iodavar, iconv.OvalName()]['coordinates'] = 'longitude latitude'
-            self.varAttrs[iodavar, iconv.OerrName()]['coordinates'] = 'longitude latitude'
-            self.varAttrs[iodavar, iconv.OqcName()]['coordinates'] = 'longitude latitude'
-            self.varAttrs[iodavar, iconv.OvalName()]['units'] = 'mol m-2'
-            self.varAttrs[iodavar, iconv.OerrName()]['units'] = 'mol m-2'
+        if self.columnType == 'total':
```
CoryMartin-NOAA, line 55: Since you're passing this through in obsVar, do you need this if statement here?

CoryMartin-NOAA (commented):
One small comment, otherwise looks good

src/compo/tropomi_no2_nc2ioda.py
jeromebarre: yes as what is passed is the dictionary defined in main().

jeromebarre: or do you mean total should be as default if no -c/--column option ?

CoryMartin-NOAA: instead of passing a dict, since you are setting it in an if, I would send it like so:
```
obsVar = {'inputvar': 'nitrogendioxide_total_column',
                    'iodavar': 'nitrogen_dioxide_in_total_column',
    }
```
so then you don't need the if, elif here, just use the keys

CoryMartin-NOAA: `iodavar = self.obsVar['inputvar']`

CoryMartin-NOAA: But I see how that complicates the loop below, so I guess it's fine

jeromebarre: I decided to keep that loop as I think we can still use the dictionary in case we need multiple obsvar in the ioda file.
But this might be irrelevant in the future. I'd just like to keep this feature for now, just in case, before we start generalizing this converter to other tropoomi and omi species or KNMI retrieval products in general, as I assume they are providing the same nc file template.

CoryMartin-NOAA (approved):

src/compo/tropomi_no2_nc2ioda.py
```diff
@@ -60,111 +52,131 @@ def __init__(self, filenames):
...
             # many variables are time, scanline, ground_pixel
             # but others are just time, scanline
             lats = ncd.groups['PRODUCT'].variables['latitude'][:].ravel()
             lons = ncd.groups['PRODUCT'].variables['longitude'][:].ravel()
             qa_value = ncd.groups['PRODUCT'].variables['qa_value'][:]  # 2D
             times = np.empty_like(qa_value, dtype=object)
             qa_value = qa_value.ravel()
+
+            # adding ability to pre filter the data using the qa value
+            # documentation recommends using > 0.75
```
srherbener, line 86: Is this the documentation from the data provider? If so it might be good to include, for the next developer, a URL reference to where the recommendation exists.

srherbener (approved):
Thanks for the update and for your patience with my slow response. I've just got the one comment that I am viewing as optional.

src/compo/tropomi_no2_nc2ioda.py
jeromebarre: > Thanks for the update and for your patience with my slow response. I've just got the one comment that I am viewing as optional.

No problems, thanks a lot for the review!

jeromebarre: I will put the url of the documentation on the next PR on this.

PR #1038, opened 2022-09-01
Add variable for GNSS bending angle data, save orginal GSI QC flags a…

gmao-jjin3:
Add variable for GNSS bending angle data, save orginal GSI QC flags and other corrections.

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -84,6 +84,16 @@ conv_bufrtypes = {
 all_LocKeyList = {
     'Station_ID': ('station_id', 'string'),
     'Time': ('datetime', 'string'),
+    'time': ('time', 'string'),
+    'ascending_flag': ('ascending_flag', 'string'),
+    'earth_radius_of_curvature': ('earth_radius_of_curvature', 'string'),
+    'reference_sat_id': ('reference_sat_id', 'string'),
+    'occulting_sat_id': ('occulting_sat_id', 'string'),
+    'record_number': ('record_number', 'string'),
+    'geoid_height_above_reference_ellipsoid': ('geoid_height_above_reference_ellipsoid', 'string'),
+    'gnss_sat_class': ('gnss_sat_class', 'string'),
+    'impact_height': ('impact_height', 'string'),
+    'impact_parameter': ('impact_parameter', 'string'),
     'Latitude': ('latitude', 'float'),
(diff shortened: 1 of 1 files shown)


src/gsi-ncdiag/gsi_ncdiag.py
```diff
@@ -1362,7 +1383,8 @@ def toIODAobs(self, OutDir, ObsBias, QCVars, TestRefs, clobber=True):
         gsiqcname = value, 'GsiEffectiveQC'
         errname = value, 'GsiFinalObsError'
         gsiqc = np.zeros_like(outdata[varDict[value]['valKey']])
-        gsiqc[outdata[errname] > 1e8] = 1
```
CoryMartin-NOAA, line 1365: This is a 'special' variable we added (GsiEffectiveQC) with the intention that it would be binary, there should already be 'QC_Flag' used somewhere I thought...

srherbener:
In an effort to get better load balancing of reviews across the group, I'm going to respectfully decline this review. Please find another reviewer for this, or you do have two reviewers without me which is sufficient.

I'm absolutely available for reviews, but I'm getting way too many requests and need to be careful about how much time I spend doing reviews. Thank you for your understanding.

src/gsi-ncdiag/gsi_ncdiag.py
gmao-jjin3: Thanks.  I noticed there was a variable "PreQC" but I couldn't understand how it was saved from "QC_flag" then, so I modified this one. Anyway, I removed this change in the updated PR.

gmao-jjin3:
> In an effort to get better load balancing of reviews across the group, I'm going to respectfully decline this review. Please find another reviewer for this, or you do have two reviewers without me which is sufficient.
> 
> I'm absolutely available for reviews, but I'm getting way too many requests and need to be careful about how much time I spend doing reviews. Thank you for your understanding.

No problem, Stephen.

climbfuji:
I just checked and http://cdash.jcsda.org:8080/viewProjects.php <http://cdash.jcsda.org:8080/viewProjects.php> is working for me. AWS says up and running, all status checks passed.

CoryMartin-NOAA (approved):
Approved based on @climbfuji 's comment

CoryMartin-NOAA (changes requested):
Will re-approve once @gmao-jjin3 fixes the coding norms

CoryMartin-NOAA (approved):

danholdaway (approved):

PR #1039, opened 2022-09-02
add capability to create MOPITT IODA files for DA window 

jeromebarre:
the MOPITT raw files are daily files we need to have additional options to slice the dataset when creating the IODA files for ingest and fit the DA windows (6h).

new option is `'-r', '--date_range'` and is optional. If not specified, the whole dataset in the raw file will be processed.

### DIFF
--- a/src/compo/mopitt_co_nc2ioda.py
+++ b/src/compo/mopitt_co_nc2ioda.py
@@ -243,7 +243,7 @@ def main():
     optional.add_argument(
         '-r', '--date_range',
         help="extract a date range to fit the data assimilation window"
-        "format -r YYYYMMDDHH YYYMMDDHH",
+        "format -r YYYYMMDDHH YYYYMMDDHH",
         type=str, metavar=('begindate', 'enddate'), nargs=2,
         default=('1970010100', '2170010100'))
 
--- a/src/compo/mopitt_co_nc2ioda.py
+++ b/src/compo/mopitt_co_nc2ioda.py
@@ -54,8 +54,9 @@ hPa2Pa = 1E2
 
 
 class mopitt(object):
-    def __init__(self, filenames):
+    def __init__(self, filenames, date_range):
         self.filenames = filenames
+        self.date_rage = date_range
         self.varDict = defaultdict(lambda: defaultdict(dict))
         self.outdata = defaultdict(lambda: DefaultOrderedDict(OrderedDict))
         self.varAttrs = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
@@ -153,6 +154,8 @@ class mopitt(object):
             # set flag: rule out all anomalous data
             flg = qa == 0
 
+            print(self.date_range)
+
             if first:
                 # add metadata variables
                 self.outdata[('datetime', 'MetaData')] = times[flg]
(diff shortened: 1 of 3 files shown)


src/compo/mopitt_co_nc2ioda.py
```diff
@@ -151,7 +152,15 @@ def _read(self):
...
 
             # set flag: rule out all anomalous data
-            flg = qa == 0
+            qaf = qa == 0
+
+            # date range to fit DA window
+            date_start = datetime.strptime(self.date_range[0], "%Y%m%d%H")
+            date_end = datetime.strptime(self.date_range[1], "%Y%m%d%H")
+            date_list = [datetime.strptime(date, "%Y-%m-%dT%H:%M:%SZ") for date in times]
+            tsf = [(date_i > date_start) & (date_i < date_end) for date_i in date_list]
```
fabiolrdiniz, line 161:
```suggestion
            tsf = [(date_i >= date_start) & (date_i < date_end) for date_i in date_list]
```

src/compo/mopitt_co_nc2ioda.py
```diff
@@ -230,10 +239,18 @@ def main():
         help="path of IODA output file",
         type=str, required=True)
 
+    optional = parser.add_argument_group(title='optional arguments')
+    optional.add_argument(
+        '-r', '--date_range',
+        help="extract a date range to fit the data assimilation windows"
```
fabiolrdiniz, line 245:
```suggestion
        help="extract a date range to fit the data assimilation window"
```

src/compo/mopitt_co_nc2ioda.py
```diff
@@ -230,10 +239,18 @@ def main():
...
+        "format -r YYYYMMDDHH YYYMMDDHH",
```
fabiolrdiniz, line 246:
```suggestion
        "format -r YYYYMMDDHH YYYYMMDDHH",
```

fabiolrdiniz (approved):
Thanks, @jeromebarre!

src/compo/mopitt_co_nc2ioda.py
```diff
@@ -230,10 +239,18 @@ def main():
         help="path of IODA output file",
         type=str, required=True)
 
+    optional = parser.add_argument_group(title='optional arguments')
+    optional.add_argument(
+        '-r', '--date_range',
```
CoryMartin-NOAA, line 244: nit picking, but this includes 'hour', so I would probably call it a 'time range'. But I won't block this from merging because of it.

jeromebarre: yeah sorry i really don't have time for this level of detail right now because of the release deadline that we have. i'll make a note and change this on the next PR on iodaconv for sure.

src/compo/mopitt_co_nc2ioda.py
CoryMartin-NOAA: @jeromebarre just need to fix this for if the date is exactly in the window, otherwise looks good

src/compo/mopitt_co_nc2ioda.py
jeromebarre: done now

CoryMartin-NOAA (approved):

PR #1043, opened 2022-09-07
add ompslpnc for geos adas along with cleanup.

karpob:
## Description

This adds another instrument which can come out of the GEOS-ADAS/GSI called ompslpnc, which is the gmao specific OMPS-LP combined uv/vis retrieval.  This also fixes a bug where layer instruments are output with a specified unit of mol mol-1 vs DU. The unit issue is just a matter of changing the unit attribute for layer sensors. Also a bit of cleanup by splitting the ozone sensors into two groups oz_lay_sensors and oz_lev_sensors to help cleanup some of the code/logic.
### Issue(s) addressed

None

adds ompslpnc to list of instruments handled by gsi_diags

## Dependencies
None

## Impact

Please list the other repositories (if any) that this PR will require changes in (example below)
None

### DIFF
--- a/src/gsi-ncdiag/gsi_ncdiag.py
+++ b/src/gsi-ncdiag/gsi_ncdiag.py
@@ -1497,7 +1497,7 @@ class Ozone(BaseGSI):
         ncout.createDimension("nlocs", nlocs)
         # other dims
         ncout.createDimension("nlevs", self.df.dimensions["mole_fraction_of_ozone_in_air_arr_dim"].size)
-        if (self.sensor  in oz_lay_sensors):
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):
Looks good to me, thanks

kwargan (approved):
For future consideration: the units that "oz_lev_sensors" actually use are ppmv, not mol/mol. There's a factor of 10^6 difference between the two. I understand that this is a legacy thing and this PR is not the place to fix it, but in the future, it would be good to either correct the units string to "ppmv" or have data written in mol/mol.

Otherwise this looks great

Issue #978, opened 2022-06-24
feature/query_cxx_types for Radiance


emilyhcliu:
These following variables should be integer (based on IODA convention).
They are exported from BUFR as float.

```
        float fieldOfViewNumber(Location) ;
                fieldOfViewNumber:_FillValue = 1.e+11f ;
                string fieldOfViewNumber:long_name = "Field of View Number" ;

        float sensorChannelNumber(Location, Channel) ;
                sensorChannelNumber:_FillValue = 1.e+11f ;
                string sensorChannelNumber:long_name = "Sensor Channel Number" ;


```

BUFR table:
```
| ORBN     |    0 |           0 |  24 | NUMERIC                  |-------------|
| SLNM     |    0 |           0 |   8 | NUMERIC                  |-------------|
| CHNM     |    0 |           0 |   6 | NUMERIC                  |-------------|
| FOVN     |    0 |           0 |   8 | NUMERIC                  |-------------|

```
Does "NUMERIC" indicate float type in BUFR?

Look like we need a mechanism to convert the generic BUFR data type the desired type that matches the IODA convention.

rmclaren:
Should these numbers be uint32 or int32 or int64 uint64 ? int64 may be the safest options for Numeric field...

rmclaren:
Should we add the unit tip Hz to the list of numeric types as well (uint64 would be appropriate for this field)

emilyhcliu:
> Should these numbers be uint32 or int32 or int64 uint64 ? int64 may be the safest options for Numeric field...

The data type defined in IODA convention document ([spreadsheet](https://docs.google.com/spreadsheets/d/1fHb2QIHOwgp4d6plyn9Dj9HZFZsDRxcPaUV6hdTb5BU/edit?usp=sharing)) for these variables is **uint32_t**

emilyhcliu:
I asked Jeff Ator a bufr data type question:

Here is my question:
```
There are also many data in the BUFR registered as "NUMERIC" (see example below)
Are these data stored as integers in the BUFR?

Example:
| SLNM     | 005041 | SCAN LINE NUMBER                                       
| CHNM     | 005042 | CHANNEL NUMBER                                         
| FOVN     | 005043 | FIELD OF VIEW NUMBER                                   
| ALFR     | 021166 | LAND FRACTION     

| MNEMONIC | SCAL | REFERENCE   | BIT | UNITS                                     
[...]

Issue #977, opened 2022-06-23
feature/query_cxx_types: Type inconsistency found in BUFR variables with the same type


emilyhcliu:
## Current behavior (describe the bug)
 
For ADPSFC PREPBUFR file ([bufr_ncep_prepbufr_adpsfc.yaml](https://github.com/JCSDA-internal/ioda-converters/blob/feature/query_cxx/test/testinput/bufr_ncep_prepbufr_adpsfc.yaml))

There are 7 quality markers: ZQM, PQM, PMQ, TQM, QQM, SSTQM, WQM, WQM  (for height, pressure, pressure reduced to mean sea level, temperature, moisture, SST, and winds, respectively)

The first fours (ZQM, PQM, PMQ, TQM) are converted to integer type.
However the rest of the markers (QQM, SSTQM, WQM) are converted to float. 

They should all be converted to integer.
The BUFR table indicates that the unit of these quality markers is CODE TABLE.  And all quality markers point to the same [code table](https://www.emc.ncep.noaa.gov/mmb/data_processing/prepbufr.doc/table_7.htm).

Query part:
```
          # QualityMarker
          heightQualityMarker:
            query: "*/Z___INFO/Z__EVENT/ZQM"
          pressureQualityMarker:
            query: "*/P___INFO/P__EVENT/PQM"
          pressureReducedToMeanSeaLevelQualityMarker:
            query: "*/PMSL_SEQ/PMQ"
          airTemperatureQualityMarker:
            query: "*/T___INFO/T__EVENT/TQM"
          specificHumidityQualityMarker:
            query: "*/Q___INFO/Q__EVENT/QQM"
          waterTemperatureQualityMarker:
            query: "*/SST_INFO/SSTEVENT/SSTQM"
          windEastwardQualityMarker:
            query: "*/W___INFO/W__EVENT/WQM"
          windNorthwardQualityMarker:
            query: "*/W___INFO/W__EVENT/WQM"

```
Export part:

```
group: QualityMarker {
  variables:
        uint airTemperature(Location, TemperatureEvent) ;
                airTemperature:_FillValue = 1215750144U ;
                string airTemperature:long_name = "Temperature Quality Marker" ;
                string airTemperature:coordinates = "longitude latitude" ;
        uint height(Location, HeightEvent) ;
                height:_FillValue = 1215750144U ;
                string height:long_name = "Height Quality Marker" ;
                string height:coordinates = "longitude latitude" ;
[...]

Issue #976, opened 2022-06-23
feature/query_cxx_types --- bug found when query variable with different precision


emilyhcliu:
## Current behavior (describe the bug)

When query a BUFR variable with mixed precision, the process will be terminated with errors.

Example:  
```
          latitude:
            query: "[*/CLAT, */CLATH]"
          longitude:
            query: "[*/CLON, */CLONH]"

```
While executing query for variable with the different precision, the process will be terminated with the following error message:
```
PRE-MAIN-ERROR Exception: Bad parameter: Different subsets units don't agree for field latitude. DEGREES vs DEGREE(N+,S-)N 
terminate called after throwing an instance of 'eckit::BadParameter'
  what():  Bad parameter: Different subsets units don't agree for field latitude. DEGREES vs DEGREE(N+,S-)N


```

This case can be reproduced using [bufr_ncep_adpsfc.yaml](https://github.com/JCSDA-internal/ioda-converters/blob/feature/query_cxx/test/testinput/bufr_ncep_adpsfc.yaml)

rmclaren:
Should be fixed, removed check for unit type...

emilyhcliu:
> 

@rmclaren Great! Thank.  I will update my build and test again.

rmclaren:
Please close this issue if it has been fixed.

Issue #965, opened 2022-06-04
Add variable type conversion in BUFR converter so the result variable type follows the IODA convention


emilyhcliu:
### In current BUFR converter, the output IODA variables inherit the data types queried from the BUFR.  These data types from BUFR are mostly float and string. There is a need to add the variable type conversion capability in BUFR converter so that the output data types follow the IODA convention.

The IODA data type can be found in the [Data Storage Types](https://docs.google.com/spreadsheets/d/1fHb2QIHOwgp4d6plyn9Dj9HZFZsDRxcPaUV6hdTb5BU/edit#gid=741804046) section in the [IODA Convention Table](https://docs.google.com/spreadsheets/d/1fHb2QIHOwgp4d6plyn9Dj9HZFZsDRxcPaUV6hdTb5BU/edit?usp=sharing).

For example: 

This is the output IODA variable (sensorChannelNumber) from BUFR.  The data type queried from BUFR is **float**. The IODA convention requires **uint32_t** ( a specific integer type)
````
        float sensorChannelNumber(Location, Channel) ;
                sensorChannelNumber:_FillValue = 1.e+11f ;
                string sensorChannelNumber:long_name = "Sensor Channel Number" ;
```

**Can we add a "type" keyword in the IODA part of the YAML to specify the data type to be converted to?**

```
        - name: "MetaData/sensorChannelNumber"
          source: variables/sensorChannelNumber
          longName: "Sensor Channel Number"
          type: uint32
```

Or 

the BUFR export part:
```
          sensorChannelNumber:
            query: "*/IASICHN/CHNM"
            type: uint32
```


## Requirements

## Acceptance Criteria (Definition of Done)

## Dependencies
This issue has a higher priority than [issue #394](https://github.com/JCSDA-internal/ioda-converters/issues/964)

emilyhcliu:
> So I think the right way to fix this would be to read the BUFR TABLE B data and grab the UNITS for the mnemonic. This way the user wouldn't need to indicate which fields to turn into int or float types (there are a lot of them). Unfortunately I would need to extend NCEPLIB-bufr a little to accomplish this... I can know the type automatically. What do you think?

@rmclaren what you proposed is even better since the types can be automatically determined internally.

srherbener:
[...]

PR #1044, opened 2022-09-12
Feature/query cxx types

rmclaren:
## Description

Adds support for a more flexible type system. BUFR field META data is inspected and the appropriate type is selected (could be float, int, uint, int64, uint64, or string). In addition the user has the ability to override the type in the bufr variables definition. 

### Issue(s) addressed

#965 
#977 
#976
#978

NETCDF file data fields will be represented in more appropriate types. The user will have the ability to override the types if necessary.

## Dependencies

#1006


CoryMartin-NOAA (commented):
Looks good to me, approval contingent on tests/builds passing though

rmclaren:
I removed support for unsigned types in order to ensure compatibility of output with the fortran clients.

emilyhcliu:
@rmclaren Are we still accept PRs for feature/query_cxx_types?

rmclaren:
@emilyhcliu In general I'd like to declare this branch as done so that we can code review and merge it as soon as possible as it contains a really important feature. Since you just want to update a YAML file and some meta-data (and there hasn't been a lot of code reviews) I think its OK to merge what you've got.

BenjaminRuston:
@rmclaren and @CoryMartin-NOAA ,,, is this the lead blocking issue now.

Get this in and more dominoes will fall?

src/bufr/BufrParser/Query/Constants.h
```diff
@@ -9,6 +9,6 @@
 namespace Ingester {
 namespace bufr {
     /// \brief The missing data value for all BUFR data.
-    const float MissingValue = 10e10;
+    const double MissingValue = 10.0e10;
```
gthompsnJCSDA, line 12: Not sure it matters, but would this make more sense as 10.0d10 since it is declared double precision?
I also admit some confusion because I thought BUFR had a single-precision missing value since a double missing would require more space to store than a single-precision.  Is this actually a NCEP bufr-lib thing not actually something inherent to BUFR storage itself?

src/bufr/BufrParser/Query/DataProvider.cpp
```diff
@@ -98,5 +101,60 @@ namespace bufr {
     {
         delete_table_data_f();
     }
+
+    TypeInfo DataProvider::getTypeInfo(FortranIdx idx) const
+    {
+        static const unsigned int UNIT_STR_LEN = 24;
+        static const unsigned int DESC_STR_LEN = 55;
```
gthompsnJCSDA, line 108: Is 55 a hard-wired limit?  Kinda arbitrary.  This may be entirely unimportant, but I often like to declare string lengths to remain multiples of 8 for memory reasons; so could this be 64?  No idea what downstream effects would occur.

src/bufr/README.md
```diff
@@ -67,6 +67,7 @@ Defines how to read data from the input BUFR file. Its sections are as follows:
               hoursFromUtc: 0  # Optional
           satellite_id:
             query: "*/SAID"
+            type: int64
```
gthompsnJCSDA, line 70: Is this done solely for testing purposes because the list of satellite identifiers is just an integer and I would find it wildly crazy if the number of values in this table exceeds the limits of single-precision integers (roughly a billion).

src/bufr/BufrParser/Query/Query.cpp
```diff
@@ -97,16 +96,15 @@ namespace bufr {
         maskCache_.insert({dataProvider_.getSubset(), masks});
     }
 
-    __details::Target Query::findTarget(const std::string &targetName,
-                                        const std::string &query) const {
+    std::shared_ptr<Target> Query::findTarget(const std::string &targetName,
```
nicholasesposito, line 99: Looks like accesses to structures now go through a pointer. Why the change to a pointer?

gthompsnJCSDA (approved):
I am approving the PR.  I had some simple inconsequential comments to make, but nothing that I think is ultra important.  I do not have sufficient C++ experience to bring a strong review to the code changes themselves.  This is a big PR and lots of changes. But, I believe the team at NCEP has done strong work to make this work.  Hats off to Ron for the tremendous amount of work here.  Forcing some data with scaling (or offsets) that require a type change from INT to FLOAT is important and this is a welcome addition.

src/bufr/DataObject.h
```diff
@@ -372,6 +399,131 @@ namespace Ingester
             typename std::enable_if<!std::is_arithmetic<T>::value, U>::type* = nullptr) const
         {
             throw std::runtime_error("The stored value was is not a number");
+            return 0.0f;
+        }
+
+        /// \brief Set the data associated with this data object (numeric DataObject).
```
nicholasesposito, line 405: What's the purpose of this new block of code?

nicholasesposito (approved):
The main things I saw in my review:

Lots of new use of pointers.
Addition of type information and forced overrides.
Added Target.h for gathering info/metadata when we execute a query.
Changes to DataObject to sets data for some of the objects.

Just had one question that I posted. Otherwise I think it looks good. Good job.

src/bufr/DataObject.h
```diff
@@ -372,6 +399,131 @@ namespace Ingester
...
+                        data_[i] = static_cast<T>(static_cast<double>(data_[i]) * val);
+                    }
+                }
+            }
+            else
+            {
+                std::ostringstream str;
+                str << "Multiplying integer field \"" << fieldName_ << "\" with a non-integer is ";
+                str << "illegal. Please convert it to a float or double.";
+                throw std::runtime_error(str.str());
```
srherbener, line 482: I realize that I do not fully understand the problems or context you are trying to solve with the data typing improvements in this PR. I apologize for having this limited understanding, but to me these new functions you added (setData, multiplyBy, offsetBy, etc.) seem overly complex since they seem to be trying to replace the compile time type matching/checking you get with templates with runtime type checking.

For example the multiplyBy routine here is set to use a factor that is of type double. Why not use a factor of type T instead. This removes the need to do the checking and type conversions (static_casting). Ie, the compiler will create a version with float, another with int, etc. according to what the rest of the code instantiates. In this scheme the non-numeric case (string) could be handled with a template specialization of DataObject that simply throws exceptions if you try to call multiplyBy with strings.

src/bufr/DataObject.h
BenjaminRuston: @rmclaren should we move forward and strip out if not needed

or even better this is something needed and useful, right?

srherbener (commented):
Thanks for all the hard work to provide this update. From the comments of others, this is a very useful update which is great.

PR #1046, opened 2022-09-13
Updated viirs_aod2ioda.py to include nchans for channel number

andytangborn:
Adds nchans = 4 to the viirs_aod2ioda.py converter.  

This is needed for the newer verisions of fv3jedi_var.x, which will fail without the channel number.


CoryMartin-NOAA (commented):
Coding norm failures:
```
./viirs_aod2ioda.py:47:38: E231 missing whitespace after ','
./viirs_aod2ioda.py:47:49: W291 trailing whitespace
./viirs_aod2ioda.py:162:28: E222 multiple spaces after operator
./viirs_aod2ioda.py:162:43: W291 trailing whitespace
```

CoryMartin-NOAA (approved):

bhuang95:
I tried to do a ctest. But it displayed the error below. Do I need to load other modules on Hera? Thanks. 
- "55:     import netCDF4 as nc
- 55: ModuleNotFoundError: No module named 'netCDF4'"

andytangborn:
Bo, this is a python module or python path issue. I'm assuming you have the
correct module for python. For python path, find your equivalent in your
directory path to:


 PYTHONPATH=$PYTHONPATH
:/scratch1/NCEPDEV/da/Andrew.Tangborn/JEDI/gdas_app_august5_2022/gdasapp/build/lib/python3.7/pyioda/

jeromebarre (changes requested):
@andytangborn can you please create the pull request that will modify the executable? So we have a series of PR that update this. I am not seeing anything in oops or fv3-jedi maybe I am missing something here? 
Just tested fv3-jedi tests on tier2 level and gfs aerosols var tests are currently passing fine. Should they break, then please link or create the PR in the other repos. So we can test.
Also wouldn't you need to also update the input obs data in fv3-jedi-data.

Thanks!

CoryMartin-NOAA:
@jeromebarre I'm a bit confused by your comment. Can you clarify? The existing 'develop' converter seems to be wrong, this fixes it so that files produced by it work properly in JEDI. For example, it should be a 2D variable with channel 4, not `aerosol_optical_depth_4`

jeromebarre:
> @jeromebarre I'm a bit confused by your comment. Can you clarify? The existing 'develop' converter seems to be wrong, this fixes it so that files produced by it work properly in JEDI. For example, it should be a 2D variable with channel 4, not `aerosol_optical_depth_4`

I guess I am confused too :)  So what the converter was producing is not what is currently in fv3-jedi-data in the testinput? Does it now with this current PR? It would have been good to have this linked to an issue were the problem is reported and explained as clearly as possible in the first place. We cannot keep having PR out of the blue like this. Thanks!

CoryMartin-NOAA:
Yes, sorry, the fv3-jedi tests seem to be using 'idealized' input obs files (probably produced by @mpagowski or @bhuang95) and they do not match what this converter is producing in `develop`.

Sorry about not creating the issue, that fell through the cracks, we've been working at a rapid pace the last week trying to get prototype cycling going and I needed a file that would work before a meeting yesterday. Do you want a 'posterior' issue created, or is explanation in this PR sufficient?

jeromebarre:
No that's fine for now. Not a big deal really. I would just like to have a better practice in general. Ideally I would like to have a system that doesn't allow you to create a mergable branch with opening an issue, that's the way it is at ECMWF for example with JIRA. We need to think of a way to have this a bit more fluid for everyone. Specific to this issue we need to make sure what is put in the converters is consistent with what is in the hofx or da tests. Is this the case? I just checked and the var names in fv3-jedi-data in the viirs testinput files are `aerosol_optical_depth` and not `aerosol_optical_depth_4` and have nloc x nchan as dimensions. Changes occured back in April see https://github.com/JCSDA-internal/fv3-jedi-data/pull/33. made by @srherbener it slipped through. Wonder how we can better keep track of this? I approve for now. Thanks for your patience and sorry to be that picky but it's for the best.

jeromebarre (approved):

PR #1055, opened 2022-09-19
Feature/query labeled dims

### LINKED ISSUE #1041: BUFR IODA: Add ability to add source field to dimension declarations.
## Description

It's desirable to be able to add the ability to have a source field on the dimension parameters. For example you might want to associate dimension information with Channels so that you can correlate data at a position with a channel number which is derived from another BUFR variable. As long as the source data numbers (like channels for example) just repeat the same way for each message this should be feasible without too much trouble. So basically we will need a process by which we validate the sources data and then associate it correctly with the dimension data.

## Requirements

dimension dictionary will be extended with a `source` parameter that points to a source field in the BUFR file. The netcdf output will then annotate the dimension fields with this data.

An example for the dimensions field will look like this:

```yaml
dimensions:
  - name: Channel 
    path: "*/BRITCSTC"
    source: variables/channelNumber
```

The `source` field is new.

rmclaren:
## Description

This pull requests adds the ability to assign a source field for an Ioda::dimension object. The values (or labaels) stored for the dimension will be derived from this source field (after some validation).

### Issue(s) addressed

- fixes #1041

The ioda::dimensions objects have an optional field called `source`. If you specify a source then the variable associated with the dimension will be filled with values from the source. The source is also validated to make sure it makes sense for the dimension. It must be 1:1 with the dimension path, and it must repeat for each sequence.

## Dependencies

Waiting on the following PRs:
- [x] waiting on [JCSDA-internals/ioda-converters/pull/1044](https://github.com/JCSDA-internal/ioda-converters/pull/1044)


gthompsnJCSDA:
@rmclaren   Would you please do an empty commit so the CI-testing is re-run?  There was a container problem when the CI-tests ran last time.  Thanks!

src/bufr/DataObject.h
```diff
@@ -96,6 +135,20 @@ namespace Ingester
                                       const std::vector<ioda::Dimensions_t>& chunks,
                                       int compressionLevel) const = 0;
 
+        /// \brief Makes an new dimension scale using this data objet as the source
```
CoryMartin-NOAA, line 138: These should be `a new dimension` x3 :-)

src/bufr/DataObject.h
```diff
@@ -172,6 +225,48 @@ namespace Ingester
...
+            for (size_t idx = 0; idx < data_.size(); idx += dimData->data.size())
+            {
+                if (!std::equal(data_.begin(),
+                                data_.begin() + dimData->data.size(),
+                                data_.begin() + idx,
+                                data_.begin() + idx + dimData->data.size()))
+                {
+                    std::stringstream errStr;
+                    errStr << "Dimension " << name << " has an invalid source field. ";
+                    errStr << "The values don't repeat in each sequence.";
```
CoryMartin-NOAA, line 251: Suggest not putting apostrophe `don't` in the comment, if someone opens the log in vi/emacs, it might highlight a lot of text thinking it is a string that is unclosed

CoryMartin-NOAA (commented):
few small comments

src/bufr/DataObject.h
```diff
@@ -172,6 +225,48 @@ namespace Ingester
             return var;
         };
 
+        /// \brief Makes an new dimension scale using this data objet as the source
```
gthompsnJCSDA, line 228: small spelling typo:  "object"

gthompsnJCSDA (approved):
Things look OK to me and I agree with a couple minor points that Cory mentioned as well.  Thanks for doing this Ron!!!

BenjaminRuston:
@rmclaren just merged PR1044 so can you resolve the conflicts and will try to get through a few more PRs

rmclaren:
@BenjaminRuston I believe it should be good OK now.

CoryMartin-NOAA (approved):

PR #1056, opened 2022-09-19
changes needed for new containers

mer-a-o:
## Description

*(Instructions: replace text in this and all sections with your own text)*

Provide a detailed description of what this PR does.

What problem does it fix? What new capability does it add?

[See the JEDI Documentation](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/practices/pullrequest.html) for further information.

### Issue(s) addressed

Link the issues to be closed with this PR
- fixes #<issue_number>

If this does not close an existing issue, then [be sure to add a Pipeline, Label, Estimate, Assignees, and Epic](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/practices/issues.html)

## Acceptance Criteria (Definition of Done)

What does it mean for this PR to be finished?

## Dependencies

If there are PRs that need to be merged before or along with this one, please add "waiting for another PR" label and list the dependencies in the other repositories (example below). Note that the branches in the other repositories should have matching names for the automated tests to pass.
Waiting on the following PRs:
- [ ] waiting on JCSDA/eckit/pull/<pr_number>
- [ ] waiting on JCSDA/atlas/pull/<pr_number>

## Impact

Please list the other repositories (if any) that this PR will require changes in (example below)

- [ ] saber
- [ ] ufo
- [ ] ...

PR #1057, opened 2022-09-20
bugfix: fix typo index in mopitt co

jeromebarre:
there was an index typo for multiple files ingest.
 the converter is failing concatenating fields form different files if not fixed


climbfuji:
Should we try to fix the general CI problems first before merging this PR? We won't be able to see if it works or not otherwise.

jeromebarre:
Seems that the code do not build on the containers? Not sure if this is a problem specific to that branch? It was fine on discover yesterday night when I debugged and tested.

climbfuji:
> Seems that the code do not build on the containers? Not sure if this is a problem specific to that branch? It was fine on discover yesterday night when I debugged and tested.

I assume it is because the containers were updated yesterday without the necessary changes to the CI system. This PR https://github.com/JCSDA-internal/ioda-converters/pull/1056 needs a few more changes and then it should work again.

mer-a-o (approved):

climbfuji (approved):

Issue #1047, opened 2022-09-13
Possible bug with the size of "size" when querying a bufr with multiple subtypes


nicholasesposito:
## Description
When converting some aircraft bufr files using my new aircraftAltitude variable from #997 with yamls that only need some subtypes' worth of data, the software, while it grabs the needed information, grabs extra data that it is not supposed to be looking for and makes it "missing" data. For example, using the bufr_d file with my "noamdar" yaml, it should look for NC004001, 2, 3, 6, 9, 10, and 11, for a total of 18367 data points. NC004103 is also in the bufr_d file and has 8455 data points, which the "size" function "sees" and adds to the total number of data points. The number of total data points is 26822 in the output nc file, which is the total of all the above. 

I've been looking through Ron's code and my branch. I think the problem is that the "size" of the latitude array is too large. It appears that it finds every latitude but does not quantify them, leaving them as "missing", but still including them in the "size"

The good news is that none of the output data is affected. For all of the "extra" data( including lat, lon, aircraftAltitude and whatever else), it defaults to missing with a dateTime variable of "-1". What do we want to do about this?

The iodaconv branch I've been using is:
feature/query_cxx_types_NE_aircraftAltitude
with the changes/additions being made in:
iodaconv/src/bufr/BufrParser/Exports/Variables

My yamls are in:
/scratch1/NCEPDEV/da/Nicholas.Esposito/JEDI/testnewbufr2ioda/
and the yamls that make the bad data are:
newNOAM.yaml    (amdar with NC specification)
newNCnoAM.yaml  (no amdar with NC specification)
newNOAM.yaml         (no amdar without NC speficiation)

## Requirements
This may be a bugfix? If it is, then the fix would not add data where dateTime is -1 and the rest of the data is missing.

## Acceptance Criteria (Definition of Done)
It is finished once 
[ ] There is an explanation
[ ] There is a fix

rmclaren:
Fix in feature/query_datetime_subset_fix

PR #1059, opened 2022-09-22
Feature/query datetime subset fix

rmclaren:
## Description

This pull request does several things:

- Filters the data down to only the data for the subsets that are included in the queries as in the example in the following file: test/testinput/bufr_specific_subsets_by_query.yaml.
- Adds a new section in exports called subsets. (see the following: test/testinput/bufr_specifying_subsets.yaml)
- Fixed some issues in the DatetimeVaraibles with regard to the interpretation of missing values.
- Includes some refactoring. The QuerySet now uses the QueryParser and queries are parsed into Query objects. The queryset can now tell you if a subset is one that needs to be processed or not.

### Issue(s) addressed

- https://github.com/JCSDA-internal/ioda-converters/issues/1047

- YAML file supports new exports field called subsets which can be used to limit the subsets that will be processed.
-  Only subsets that are specified in queries will be processed (* means everything).
- Fixes bugs with DatetimeVariable object.

## Dependencies

- https://github.com/JCSDA-internal/ioda-converters/pull/1055

### DIFF
--- a/src/bufr/DataObject.h
+++ b/src/bufr/DataObject.h
@@ -110,9 +110,9 @@ namespace Ingester
         /// \return Float data.
         virtual float getAsFloat(const Location& loc) const = 0;
 
-        /// \brief Get the data at the index as an float.
+        /// \brief Get the data at the index as an int.
         /// \return Float data.
-        virtual float getAsFloat(size_t idx) const = 0;
+        virtual int getAsInt(size_t idx) const = 0;
 
         /// \brief Get the data at the Location as an string.
         /// \return String data.
@@ -177,7 +177,7 @@ namespace Ingester
     {
      public:
         typedef T value_type;
-        constexpr T missingValue() const { return std::numeric_limits<T>::max(); }
+        static constexpr T missingValue() { return std::numeric_limits<T>::max(); }
 
         /// \brief Constructor.
         /// \param dimensions The dimensions of the data object.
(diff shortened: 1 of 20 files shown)


src/bufr/DataObject.h
```diff
@@ -110,9 +110,9 @@ namespace Ingester
         /// \return Float data.
         virtual float getAsFloat(const Location& loc) const = 0;
 
-        /// \brief Get the data at the index as an float.
+        /// \brief Get the data at the index as an int.
         /// \return Float data.
-        virtual float getAsFloat(size_t idx) const = 0;
+        virtual int getAsInt(size_t idx) const = 0;
```
nicholasesposito, line 115: Ron, I am using the getAsFloat function in my feature/query_cxx_types_NE_aircraftAltitude branch. I think we should keep both

rmclaren: Actually I really need to get rid of both as using this function is really dangerous if you are not careful... If you depend on any values in repeated sequences (or that could be in repeated sequences) you should not go near these methods. I'm gonna have to think about this a little I think.

nicholasesposito: OK, I'll rework my code to not use it.

nicholasesposito (approved):
Lots of good changes in this update, especially regarding the determination of array sizes. I will definitely use this to fix up my own aircraftAltitude. Thanks for doing this.

src/bufr/DataObject.h
rmclaren: Ok, I added it back. You need to validate the input data and make sure it all comes from the same place (same place I the tree).

rmclaren: This will need to be revisited in the future.

PR #1060, opened 2022-09-22
Add ability to convert Text radiosonde data (TEMP format) to IODA

gthompsnJCSDA:
## Description

This PR will permit ingestion of text alphanumeric legacy TEMP format radiosonde data still used by many countries into IODA.  Most countries now use higher resolution BUFR data (or produce both old and new formats).  The decoder code was originally provided by NCAR-RAL's Andy Gaydos with mods for IODA output by gthompsn.  This code can directly ingest LDM-received sounding data complete with WMO headers.

Running this code requires a station table (entered on the command-line using option: ```-t name_of_station_file``` in JSON format.  A file was created from the [source station file of the RAOB software](http://www.raob.com/assets/downloads/raob.stn.txt).  Only from the list of stations in the table file can a sounding be decoded from the incoming data.  This means that new sites need to be added as they become available.  The list of sites in the station table includes hundreds of sites that may not launch balloons regularly.

### Issue(s) addressed

A corresponding issue will be edited in (soon).

Working ctests and positive reviews

## Dependencies

None

## Impact

None anticipated


src/lib-python/meteo_sounding_utils.py
```diff
@@ -0,0 +1,159 @@
...
+# the various sounding parts (e.g. TTAA (mandatory P,Z,T), TTBB (significant
+# P,T) and PPBB (winds given height alone).
+#
+# Peter Neilley, NCAR/RAP 8/93
+#
+# --+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+
+import math
+
+# Constants
```
BenjaminRuston, line 13: looks good @gthompsnJCSDA ,,, in the future think about how to abstract these

sure they are already existing in other parts of the code.

test/CMakeLists.txt
```diff
@@ -321,6 +323,16 @@ foreach(FILENAME ${test_input} ${test_output})
            ${CMAKE_CURRENT_BINARY_DIR}/${FILENAME} )
 endforeach(FILENAME)
 
+# create the share directory and link files to test subdir
```
gthompsnJCSDA, line 326: @srherbener   Please see if you agree this is an acceptable way to implement a link to the share directory and that all files in that directory are linked.  We could limit the fileglob to use file extensions like ```*.json``` if you think that is a better idea.

huishao-r (approved):
looks good to me.

BenjaminRuston (approved):
this looks good, so let's move forward with this

PR #1064, opened 2022-09-23
 change from two composition related ctest to nonzero tolerances

BenjaminRuston:
## Description

currently the intel compiler is having issue with non-zero tolerances for a couple of composition related ctest

### Issue(s) addressed

Link the issues to be closed with this PR
- fixes #1063 

## Acceptance Criteria (Definition of Done)

These two ctests, and all the others as well, pass:
test_iodaconv_tropomi_no2_tropo
test_iodaconv_aeronet_aod

PR #1066, opened 2022-09-23
fix to using zero tolerances

### LINKED ISSUE #1063: remove zero tolerances for a couple composition related ctests
the intel compiler is failing a couple composition ctest due to a zero tolerance being set

PatNichols:
## Description

Some of the ctests are failing due to the tolerances being set to zero. This PR will hopefully fix this by assigning a small but non-zero tolerance from these tests.

### Issue(s) addressed
-fixes #1063  

  ctests that should fail will hopefully fail but those with small errors due to compiler differences will pass.

## Dependencies

None

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -337,7 +337,7 @@ endforeach(FILENAME)
 # iodaconv_comp.sh test.
 #
 # For a converter that is generating new data, use a non-zero tolerance.
-set(IODA_CONV_COMP_TOL_ZERO "0.0")
+set(IODA_CONV_COMP_TOL_ZERO "1.e-7")
 set(IODA_CONV_COMP_TOL "0.5e-4")
 
 # The ioda-engines CMake configuration has set the file target _ioda_python to


fabiolrdiniz (commented):
Thanks for adding me here, @PatNichols. I'm a little bit confused. Why don't use the `IODA_CONV_COMP_TOL` for the comparison of the tests that are failing? It's already defined in the code (see line 341).

test/CMakeLists.txt
```diff
@@ -337,7 +337,7 @@ endforeach(FILENAME)
 # iodaconv_comp.sh test.
 #
 # For a converter that is generating new data, use a non-zero tolerance.
-set(IODA_CONV_COMP_TOL_ZERO "0.0")
+set(IODA_CONV_COMP_TOL_ZERO "1.e-7")
```
BenjaminRuston, line 340: @PatNichols do not think this is the way to go... this definition should remain zero

my approach was to change those tests to use the other tolerance:

`IODA_CONV_COMP_TOL` please see my PR #1064

BenjaminRuston:
this is redundant functionally with #1064

PR #1079, opened 2022-10-13
amsr2 ice concentration converter

### LINKED ISSUE #1078: converter for AMSR-2 sea ice concentration
Create a converter for AMSR-2 sea-ice concentration available in NCEP data tanks.

ShastriPaturi:
## Description
This PR creates an IODA converter for sea-ice concem=ntration from l2p AMSR-2 available in NCEP data tanks.
### Issue(s) addressed

- fixes #1078 

passes the ctest.


src/marine/amsr2_icec2ioda.py
```diff
@@ -0,0 +1,134 @@
...
+    ObsVars, nlocs = iconv.ExtractObsData(icec.data, locationKeyList)
+
+    DimDict = {'nlocs': nlocs}
+    writer = iconv.IodaWriter(args.output, locationKeyList, DimDict)
+
+    VarAttrs = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
+    VarAttrs[('sea_area_fraction', 'ObsValue')]['units'] = '1'
+    VarAttrs[('sea_area_fraction', 'ObsError')]['units'] = '1'
+    VarAttrs[('sea_area_fraction', 'PreQC')]['units'] = 'unitless'
+    VarAttrs[('sea_area_fraction', 'ObsValue')]['_FillValue'] = 999
```
kbhargava, line 126: It probably doesn't matter here as there aren't any missing values. But it would be better to read the FillValue from the input file and using that here.

kbhargava (changes requested):

test/CMakeLists.txt
```diff
@@ -388,6 +390,18 @@ set(IODACONV_PYTHONPATH "$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
...
 #===============================================================================
 
+ecbuild_add_test( TARGET  test_${PROJECT_NAME}_amsr2_l2icec
+                  TYPE    SCRIPT
+                  ENVIRONMENT "PYTHONPATH=${IODACONV_PYTHONPATH}"
+                  COMMAND bash
+                  ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                          netcdf
+                          "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/amsr2_icec2ioda.py
+                          -i testinput/amsr2_icec_l2p.nc
```
kbhargava, line 400: @ShastriPaturi Do you need such big file for the test? If  not You can use nco to reduce the testinput and testoutput data.

ShastriPaturi: @kbhargava I have reduced the file size to 2.3mb using nco. I can make it much smaller.

liuxiao37k (approved):
I think we typically keep ~100 obs in testoutput files as a rule of thumb. <100KB for single band obs. Other than that, the PR looks good to me.

mjhossen (approved):

src/marine/amsr2_icec2ioda.py
ShastriPaturi: Read/add FillValue and units

kbhargava (approved):

srherbener (approved):
Thanks for reducing the test file sizes. Looks great now!

src/marine/amsr2_icec2ioda.py
gthompsnJCSDA: Line 126 should be deleted.  There is no such thing as ```unitless``` in our JEDI Data Convention - so it should simply be skipped.

ShastriPaturi: @gthompsnJCSDA line 126 with unitless has been omitted. The units are being read from "units" attribute in the nc file.

gthompsnJCSDA: OK, but please ensure that in the case for which a PreQC variable may have units assigned as ```unitless``` then the converter still should not use those units and instead skip this attribute for this variable.  And that should be the case for any variable in which the input file may contain unitless.  Thanks!

PR #1086, opened 2022-10-18
add optional soilDepth argument and convert mask to boolean switch

### LINKED ISSUE #1068: Add observation depth to SMAP soil moisture IODA converter
## Description

For UFS issue 1752 we are adding the ability to do soil level interpolation, so that the UFS can simulate the soil moisture over the same depths as is observed.  See https://github.com/JCSDA-internal/ufo/issues/1752

To do this, we need to add a vertical observation depth to the SMAP IODA files.

@YoulongXia-NOAA can you please add an observed depth variable to the SMAP soil moisture IODA converters? For sondes, there is a metadata field called height (in m). @fcvdb @BenjaminRuston @huishao-r : I assume we want to do the same? 

The depth of all observations should be 2.5 cm. (observed depth is approximately 0-5 cm layers, so giving the middle point will be most consistent with the UFO's point-based interpolation routines).

This is Task 2 of the UFO issue linked above.

## Requirements

No new dependencies. Success is the observation depth being added to the SMAP soil moisture IODA converter.

BenjaminRuston:
## Description

suggestion to use optional argument to add a assumed soil depth

### Issue(s) addressed

vicariously Linked to
- fixes #1068 

converter runs and produced IODA output files


BenjaminRuston:
@YoulongXia-NOAA 

does this work for you?  this can be used to allow soilDepth to be better described and optional if anyone wants to change it

YoulongXia-NOAA:
@BenjaminRuston Only needs to modify  CMakeLists.txt in test. For example, use -depth 0.025 as the input.

YoulongXia-NOAA:
@BenjaminRuston, based on your comments, I have updated my two converters, and the corresponding CMakeList in test. The ctests passed. I have done "git push" to my feature branch. See if this is what you want. @ClaraDraper-NOAA and @zcstanley, you can use -d 0.025 or default to execute the python code. In the future, if the you have the different depth, just change  CMakeList not python code.

ClaraDraper-NOAA:
> @BenjaminRuston, based on your comments, I have updated my two converters, and the corresponding CMakeList in test. The ctests passed. I have done "git push" to my feature branch. See if this is what you want. @ClaraDraper-NOAA and @zcstanley, you can use -d 0.025 or default to execute the python code. In the future, if the you have the different depth, just change CMakeList not python code.

Thanks @YoulongXia-NOAA This looks like a good solution to me.

BenjaminRuston:
@YoulongXia-NOAA think we can finally complete this task, the new naming convention merger is complete. Would you like to proceed and get this merged

YoulongXia-NOAA (approved):
Look good to me. I approve this PR.

BenjaminRuston:
@YoulongXia-NOAA this is not quite ready as the conflicts need to be resolved

can you confirm that the `refsec = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()`

this `tb_time_seconds` is indeed the `seconds since 1970-01-01T00:00:00Z` so a long integer

YoulongXia-NOAA:
@BenjaminRuston, look its user guide for 9km product and the description is below: tb_time_seconds
Daily global composite of the arithmetic average of the same parameters found in the fore- and aft-looking groups in the input
L1C_TB_E granule. The resulting parameter thus describes the average of UTC acquisition times of L1B_TB_E observations
whose boresights fall within a 9 km EASEGrid 2.0 cell. The result is then expressed in J2000 seconds [the number of seconds since 12:00:00.000 on January 1, 2000 Barycentric Dynamical Time (TDB)]. I will check 32 km realtime SMAP product. In this case, we need to add time seconds from 1970-01-01-00 to 2000-01-01 11:59:59.

YoulongXia-NOAA:
@BenjaminRuston, For 32 km real-time product, it is expressed as: tb_time_seconds
!  Arithmetic average of the same parameters found in the 
!  fore- and aft-looking groups in the input SPL1CTB granule. 
!  The resulting parameter thus describes the average of UTC 
!  acquisition times of SPL1BTB observations whose boresights 
!  fall within a 36 km EASE-Grid 2.0 cell. The result is then 
!  expressed in J2000 seconds (the number of seconds since 
!  11:58:55.816 on January 1, 2000 UT) in https://github.com/NASA-LIS/LISF/blob/master/lis/dataassim/obs/NASA_SMAPsm/read_NASASMAPsm.F90.

src/land/smap9km_ssm2ioda.py
```diff
@@ -140,6 +143,7 @@ def _read(self):
         self.outdata[('depthBelowSoilSurface', 'MetaData')] = deps
         self.varAttrs[('depthBelowSoilSurface', 'MetaData')]['units'] = 'm'
         self.outdata[('surfaceQualifier', 'MetaData')] = sflg
+        self.varAttrs[('surfaceQualifier', 'MetaData')]['units'] = 'unitless'
```
BenjaminRuston, line 146: @gthompsnJCSDA is the `unitless` deprecated now for the `surfaceQualifier`

BenjaminRuston:
@YoulongXia-NOAA great think I got it, think this converter was updated in the naming convention sprints but changed the epoch to be to the convention

is this a two step process where the `smap9km_ssm2ioda.py` produces output that is read in by the `smap_ssm2ioda.py`?  but since the `smap_ssm2ioda.py` is getting a time from the filename it still needs to be converted to the long integer going forward

src/land/smap_ssm2ioda.py
```diff
@@ -90,10 +90,8 @@ def _read(self):
         lons = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['longitude'][:].ravel()
         errs = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['soil_moisture_error'][:].ravel()
         qflg = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['retrieval_qual_flag'][:].ravel()
-
         deps = np.full_like(vals, self.assumedSoilDepth)
-        times = np.empty_like(vals, dtype=np.int64)
-
+        times = np.empty_like(vals, dtype=object)
```
BenjaminRuston, line 94: @YoulongXia-NOAA and @zcstanley didn't see any conversion of this string time (from the filename) to integer seconds so leaving as is for future update

YoulongXia-NOAA: @BenjaminRuston, see this part:
# get datetime from filename
        str_split = self.filename.split("_")
        str_datetime = str_split[7]
        my_datetime = datetime.strptime(str_datetime, "%Y%m%dT%H%M%S")
        time_offset = round((my_datetime - epoch).total_seconds())

src/land/smap_ssm2ioda.py
```diff
@@ -106,9 +105,12 @@ def _read(self):
             times = times[mask]
 
         # get datetime from filename
-        str_split = self.filename.split("_")
-        str_datetime = str_split[7]
-        my_datetime = datetime.strptime(str_datetime, "%Y%m%dT%H%M%S")
+        # file provides yyyy-mm-dd as an attribute
+        str_datetime = ncd.groups['Metadata'].groups['DatasetIdentification'].getncattr('creationDate')
+        my_datetime = datetime.strptime(str_datetime, "%Y-%m-%d")
```
BenjaminRuston, line 110: @YoulongXia-NOAA the file itself contains an attribute called `creationDate` this is just to the nearest day 

what is the update frequency of this product?

PatNichols (approved):

PR #1091, opened 2022-10-26
Using a best guesstimate of launch time, height, and wind components, estimate the lat/lon of a drifting radiosonde

### LINKED ISSUE #1092: Legacy-text radiosonde data has lat/lon of launch location only but we need drifting in air trajectory of balloon
## Description

The WMO TEMP format legacy text for radiosondes contains only a station identifier (WMO 5-digit number), the day of the month and the hour (effectively synoptic hour such as 00 or 12z).  The existing ioda-converter uses a lookup table for station number to lat/lon but we can use the height of balloon and prior winds to estimate the trajectory by assuming a 5 m/s ascent rate and an actual launch time that is 55 minutes before the referenced time (a good typical estimate).  Using these pieces of information, we can reconstruct the best estimate of sounding trajectory through the atmosphere.  Deviations from reality are obviously increasing with increasing altitude, but the estimates are pretty good through the troposphere by comparing estimated position to actual position at sites with corresponding high-resolution BUFR data.

## Requirements

The code needs to show balloon trajectories that make sense - given simple wind considerations.

New lat, lon positions make sense.

## Dependencies

There are none.

gthompsnJCSDA:
## Description

The legacy text radiosonde data is stamped only with a day of month and time in hours.  If we assume the launch time is 55 minutes before the stated time (usually 12z sonde launches happen about 11:05UTC), plus the height and prior winds at each level, we can estimate the latitude, longitude of the drifting balloon position.  While UFO has code to do this, we may wish to have it in the ioda-converter directly.  The two assumptions comprising this code addition is the 55 minute before the stated hour for launch time and an assumed ascent rate of 5 m/s (which is quoted in a textbook chapter by NRL's Patricia Pauley).  The ascent rate is generally correct when comparing BUFR soundings with text-formatted legacy versions of the same sounding.

### Issue(s) addressed

Closes #1092 

Two positive reviews and a sanity check via plotting of some sondes (which I will add as a graphic in the conversation section); also working ctest.

## Dependencies

There are none

## Impact

There should be no impact to other repos.


fcvdb (approved):
I built the branch and ran the following commands:
`python3 ../src/conventional/sonde_tac2ioda.py -i testinput/2021081612_RAOB_small.txt --netcdf  -o 2021081612_RAOB_small.nc --verbose --debug -t .././share/raob_stations_small.json`
Then:
`ncdiff 2021081612_RAOB_small.nc testoutput/2021081612_RAOB_small.nc -o diff.nc`
and verified that `diff.nc` was filled of 0.

src/conventional/sonde_tac2ioda.py
```diff
@@ -966,8 +970,12 @@ def change_vars(profile):
         if 'height' in levels[pressure] and levels[pressure]['height'] is not None:
             height = levels[pressure]['height']
             dz = height - heightKm1
-            # Typical radiosonde ascent rate is 5 m/s
-            this_datetime = previous_time + timedelta(seconds=dz*0.2)
+            # Legacy soundings produce null values at mandatory level below ground.
+            if (dz < 1.0):
```
ctgh, line 974: Can I check why the threshold of `1.0` is used here please?

src/conventional/sonde_tac2ioda.py
```diff
@@ -966,8 +970,12 @@ def change_vars(profile):
...
+                this_datetime = previous_time
+            else:
+                # Typical radiosonde ascent rate is 5 m/s
+                this_datetime = previous_time + timedelta(seconds=dz*0.2)
```
ctgh, line 978: Dividing by `5.0` here might make the intent clearer. Up to you.

ctgh (commented):
Thanks for adding this. I have a few comments.

1. The ufo transformation code uses the mean of vertically-adjacent wind components at each point in order to determine the horizontal motion. I am not sure if you'd want to implement that or not; I found that it does not have a large impact except sometimes at the highest altitudes.
2. Presumably you don't want to apply this transformation to BUFR sondes. Is there a option that ensures it is only applied to a subset? (Sorry if that's obvious - I am not familiar with ioda-converters.)
3. I have not looked into the mechanism behind this, but data obtained from ODB files seem to have a launch time set which can differ from hh:05. Therefore the ufo routine does not subtract 55 minutes. Would it be worth adding an option to do that?
4. The assumed ascent rate in ufo is 5.16 m/s, which is probably spuriously precise. It matches the value used in OPS.

src/conventional/sonde_tac2ioda.py
gthompsnJCSDA: Yes, it would be more clear and yet I often multiply by the inverse number for making code run as fast as possible since sometimes division is slower than multiplication and the comment above helps make it clear.

src/conventional/sonde_tac2ioda.py
```diff
@@ -994,6 +1002,36 @@ def change_vars(profile):
...
+
+        if (new_profile['eastward_wind'][idx-1] != float_missing_value and new_profile['northward_wind'][idx-1] != float_missing_value):
+            previous_idx = idx
+            # move north-south
+            d_north = new_profile['northward_wind'][idx-1] * delta_t[idx-1]
+            location = geod.direct(points=previous_loc[:2], azimuths=0., distances=d_north)[0]
+            new_profile['latitude'][idx] = location[1]
+            # move east-west
+            d_east = new_profile['eastward_wind'][idx-1] * delta_t[idx-1]
+            location = geod.direct(points=location[:2], azimuths=90., distances=d_east)[0]
```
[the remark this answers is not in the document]
gthompsnJCSDA: If you test that geodesic.direct function and give it a direct east (or north) azimuth over hundreds of km, then the result ends up with a slight offset towards the equator.  Must be internal precision or something.  So it was decided to go one direction than the other.  However, for the relatively short distances between position reports, this should be inconsequential.

src/conventional/sonde_tac2ioda.py
gthompsnJCSDA: Simply used a starting value of 1 to ensure the balloon is "lifting" away from the ground.  It's arbitrary.  In truth, we never seem to know exactly what elevation a real launch happens given the accuracy in most station tables.

gthompsnJCSDA:
@ctgh  Nice review, thanks.  The precision of 5.16 m/s ascent rate is kinda funny.  Clearly we could do that as well.  I simply went off the chapter text by Pat Pauley.  Regarding the launch time - it's impossible to know for legacy text soundings we receive.  It is not part of the WMO TEMP message whatsoever.  The timestamp for legacy text data we receive over NOAA-SBN is a WMO bulletin header of solely day-of-month and hour, which is intended to be considered the synoptic times like 12z or 00z.  If you have any better estimate of launch time offset from synoptic other than 55 minutes, then I am highly interested.  With BUFR soundings (co-located with legacy text), we could compute the mean.  There are probably times when this is woefully inadequate.  We have already done a minimal comparison of a single day for BUFR sondes at about 40 sites in which the errors introduced by using winds to get lat/lon at least appear reasonable and deviations that appear centered near zero.  The code in this ioda-converter **only** operates on the legacy text so this estimating of lat/lon is only for non-BUFR data.  Lastly, I'm not sure which winds you are averaging in the vertical but I can only assume it would be the one below and the one at the current report height since you would want the balloon to experience something "in the middle."  That's a good idea, but not sure how much operational difference it might have considering the guesswork on the ascent rate anyway.

ctgh (approved):
@gthompsnJCSDA Thanks for your replies - that all makes sense!

gthompsnJCSDA:
For the small input file with only 6 sounding sites, a plot of the 327 locations showing each balloons trajectory is shown here.

Screen Shot 2022-10-27 at 10 03 35 AM

gthompsnJCSDA:
And a sample for entire globe from 1200UTC 16-Aug-2021. Things look quite sensible to me.
sonde_plot

fabiolrdiniz (approved):
Thanks, @gthompsnJCSDA!

PR #1100, opened 2022-11-02
Improved compatibility with WMO BUFR files.

### LINKED ISSUE #1084: Query: WMO files that include multiple subsets are not read correctly.
There is an issue for WMO files that include more than one kind of subset. NCEP files store subset meta data information in the first message (tree data for all subsets). WMO files do not have this feature, therefor you can't just read the table once and reuse the information. You have to be a little smarter about reading and caching table information in each message subset... So currently WMO bufr files with 1 subset type work OK, but ones with multiples don't work. To fix this in a way that will not kill performance requires some refactoring...

rmclaren:
## Description

This pull request fixes several issues encountered when trying to read WMO BUFR files with the bufr2ioda.x tool. For example: WMO file subsets can come in different flavors (or variants) were the data structures are different (the subset string doesn't uniquely identify a specific type of data ugg...).

### Issue(s) addressed

- fixes #1084 

New test test_iodaconv_bufr_wmo_amdar_multi passes.


src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
```diff
@@ -87,17 +87,19 @@ namespace Ingester
         std::vector<int64_t> timeOffsets;
         timeOffsets.reserve(map.at(getExportKey(ConfKeys::Year))->size());
 
+        auto yearVar = map.at(getExportKey(ConfKeys::Year));
+
         // Validation
-        if (map.at(getExportKey(ConfKeys::Year))->getDims().size() != 1 ||
-            map.at(getExportKey(ConfKeys::Month))->getDims().size() != 1 ||
-            map.at(getExportKey(ConfKeys::Day))->getDims().size() != 1 ||
+        if (!yearVar->hasSamePath(map.at(getExportKey(ConfKeys::Year))) ||
```
nicholasesposito, line 93: Don't you want yearVar to haveSamePath as ConfKeys year in this case? The ! in the front would say that you don't. If not, why's that?

src/bufr/BufrParser/Query/File.cpp
```diff
@@ -13,96 +13,70 @@
...
     {
-        closbf_f(fileUnit_);
-        close_f(fileUnit_);
+        dataProvider_->close();
     }
 
-    void File::rewind() {
-        close();
-        open();
+    void File::rewind()
```
nicholasesposito, line 43: Why do we want to rewind here?

src/bufr/bufr2ioda.cpp
```diff
@@ -111,15 +111,18 @@ int main(int argc, char **argv)
...
+
+//    try
+//    {
+//        Ingester::registerParsers();
+//        Ingester::parse(yamlPath, numMsgs);
+//    }
+//    catch (const std::exception &e)
+//    {
+//        throw;
+//    }
```
nicholasesposito, line 125: Should this be deleted?

tools/bufr/QueryPrinter/NcepQueryPrinter.cpp
```diff
@@ -0,0 +1,94 @@
...
+    {
+    }
+
+    std::vector<QueryData> NcepQueryPrinter::getQueries(const SubsetVariant& variant)
+    {
+        if (dataProvider_->isFileOpen())
+        {
+            std::ostringstream errStr;
+            errStr << "Tried to call QueryPrinter::getQueries, but the file is already open!";
+            throw eckit::BadParameter(errStr.str());
```
nicholasesposito, line 34: Why would it be a problem if the file is already open?

nicholasesposito (commented):
A few questions/comments on my initial look through.

src/bufr/BufrParser/Exports/Variables/DatetimeVariable.cpp
rmclaren: This was a little silly as it would always have the same path. This iff statement looks for an inconsistency in file paths an creates an exception if there is one.

src/bufr/BufrParser/Query/File.cpp
rmclaren: This just gives the user of the File class the ability to rewind the file (start over). It could be that we don't need a seperate DataProvider class.

src/bufr/bufr2ioda.cpp
rmclaren: oops, needed to coment this back

tools/bufr/QueryPrinter/NcepQueryPrinter.cpp
rmclaren: You can not open a file twice. The operating system would not permit it. We would have an exception... this just provides a nicer error message.

nicholasesposito: Understood.

nicholasesposito (approved):

rmclaren:
@gthompsnJCSDA sent me the following in an email last week and I'm not sure how to respond:

_"2) can you please leave the file names alone that existed before you changed things?  I have a feature branch waiting for your code updates and I don't wanna go through all new file names based on what you changed for this one.  My naming convention for yaml files, BUFR files, and testoutput files existed for a logical reason (IMO): that is, the data type always comes first, not whether the data are bufr format ncep flavor or WMO flavor.  That way I can more readily merge your PR together with my feature branch to move onwards from there without worry about my CMakeLists.txt file having to rename everything."_

 I think he was mostly referring to naming conventions of files in the test directory (not the source files). It would be nice to have some kind of naming convention for these files that we try to adhere to universally instead of several different ones. I liked Emilys convention as it naturally groups types of things (when sorted alphabetically). Not sure if we need to deal with this now (while things are changing anyways as apposed to later). So I'm just throwing this out there as a question. Is there any consensus as to what direction we want to go?

BenjaminRuston:
@rmclaren is this still in play,, are you all in on the sprint branch of ioda and ioda-converters and this no longer is an issue or should we proceed with this?

rmclaren:
@BenjaminRuston Fixing this involved some major code changes and doesn't really involve the name changes from the code sprint. I might be confused about what you mean exactly, but yes this pull request is still very much "in play".

srherbener (approved):
Thanks for addressing my comments.

BenjaminRuston:
thanks @rmclaren this is what I thought, you need it and it shouldn't conflict with the changes in the sprint branches.  We'll need to merge this into the sprint branch as well... will keep you in the loop

rmclaren:
@BenjaminRuston Thanks Ben!

gthompsnJCSDA:
@rmclaren   Can we discuss this file naming business by telecon please?  I had a naming convention specific to a few things such as .bufr at the end of filename does not need a prefix of bufr as well.  Then, I also used something like <format>2ioda in the name to indicate what is happening in the yaml names.

And by your name changes, you broke the python AMDAR converter because it was using the same input bufr file that was changed name but not to the argument in the python converter in CMakeLists.txt

rmclaren:
Could someone share the CDASH output for these builds? I'm not able to access them...

gthompsnJCSDA:
@rmclaren 
`test_iodaconv_amdar` is the only failing test - that's my python converter attempting to use the file you renamed so it doesn't exist any longer.  Your code changes can easily be a sole PR and we can deal with file names at a later date.  There's clearly personal preference here acting on how we name files but we should postpone that until later and get your PR code merges so we can finish the sprint branch asap please.

Here is the failure message for the ctest (from cdash)...
```
Traceback (most recent call last):
  File "/build_container/bin/amdar_bufr2ioda.py", line 532, in <module>
    parser.error('Input (-i option) file: ', file_name, ' does not exist')
TypeError: ArgumentParser.error() takes 2 positional arguments but 4 were given
```

rmclaren:
@gthompsnJCSDA Would be nice to have some kind of convention for naming these files, but at the same time I'm not sure we should be spending time on this right now... I changed the name of the amdar wmo BUFR file back to the way it was. Some day we should get together with all the stake holders (you, emily, steve ... etc) and come up with a convention to follow.

BenjaminRuston (approved):
@rmclaren thanks for working on this.

going to go ahead and move forward with this one, note the reversion of the amdar name can review up to and after merge of sprint branch

PR #1112, opened 2022-11-09
Use timegm instead of mktime.

srherbener:
## Description

This PR replaces calls to mktime (which assumes local time) with timegm (which assumes UTC). This alleviates the need to force the TZ environment variable to UTC to get the time_t value that is relative to UTC (instead of local time).

### Issue(s) addressed

None

iodaconv ctests pass

## Dependencies

None

## Impact

None


gthompsnJCSDA (approved):
Thanks Steve!  I don't actually know why this was so hard to find in C++ when I have used this for 25 years in Perl. UGH!

PR #1142, opened 2023-01-12
make L2 trace gas converters work and be consistent with what the upgrader

jeromebarre:
This makes those two converters actually working and be in phase with what the upgrader is doing.
We have tested various repos and skylab experiments and everything else seems good to go now.
This is the last piece to have everything in place before the merge.

Of course more changes will be done to actually go with the convention as soon as we actually can (i.e. once the IODA Fortran API merged and tested) on the ancilliary variables (averaging kernel and pressure vertices).

For the ObsVal names I think it is better to keep this out as otherwise we would need to re-upgrade all testinput/output/yaml data in ufo/ufo-data/fv3-jedi/fv3-jedi-data and r2d2 for slylab...

Thanks for your understanding on this!

### DIFF
--- a/src/compo/mopitt_co_nc2ioda.py
+++ b/src/compo/mopitt_co_nc2ioda.py
@@ -30,22 +30,20 @@ locationKeyList = [
     ("dateTime", "long"),
 ]
 
-varname_co = 'carbonmonoxideColumn'
-
-# This is apparently not used
 obsvars = {
     'carbonmonoxide_total_column': 'carbon_monoxide_in_total_column',
 }
 
 AttrData = {
     'converter': os.path.basename(__file__),
+    'nvars': np.int32(len(obsvars)),
 }
 
 DimDict = {
 }
 
 VarDims = {
-    varname_co: ['Location'],
+    'carbon_monoxide_in_total_column': ['Location'],
 }
 
 # constants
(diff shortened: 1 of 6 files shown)


BenjaminRuston (approved):
nice to have regression testing to show passing for that last sanity check

at this point I see no problem as both version of the code are save for posterity if needed

src/compo/mopitt_co_nc2ioda.py
```diff
@@ -30,22 +30,20 @@
...
-varname_co = 'carbonmonoxideColumn'
-
-# This is apparently not used
 obsvars = {
     'carbonmonoxide_total_column': 'carbon_monoxide_in_total_column',
 }
 
 AttrData = {
     'converter': os.path.basename(__file__),
+    'nvars': np.int32(len(obsvars)),
```
gthompsnJCSDA, line 39: Technically, this line should have been eliminated. It is deprecated.

src/compo/tropomi_no2_co_nc2ioda.py
```diff
@@ -25,27 +25,19 @@
...
+    ("latitude", "float"),
+    ("longitude", "float"),
+    ("dateTime", "string"),
 ]
-meta_keys = [m_item[0] for m_item in locationKeyList]
 
 AttrData = {
-    'converter': os.path.basename(__file__)
+    'converter': os.path.basename(__file__),
+    'nvars': np.int32(1),
```
gthompsnJCSDA, line 35: Same as last comment. This line should be eliminated, `nvars` is deprecated.

src/compo/mopitt_co_nc2ioda.py
```diff
@@ -183,42 +179,44 @@ def _read(self):
...
             first = False
 
         DimDict['Location'] = len(self.outdata[('dateTime', 'MetaData')])
+        AttrData['Location'] = np.int32(DimDict['Location'])
 
         for k in range(nlevs):
             varname = 'averaging_kernel_level_'+str(k+1)
             vkey = (varname, 'RtrvlAncData')
             self.varAttrs[vkey]['coordinates'] = 'longitude latitude'
+            self.varAttrs[vkey]['units'] = ''
```
gthompsnJCSDA, line 219: unitless and blank units should be omitted entirely.

PR #1147, opened 2023-01-19
Radiosonde legacy/text (TAC) bugfix for knots vs m/s based on day-of-month coding per WMO manual on codes

gthompsnJCSDA:
## Description

A bug existed for converting legacy/text (TAC) radiosonde reports with respect to how winds are sometimes in knots versus meters per second.  WMO manual on codes (chain of pages 85, 203, 211, 254) says that day of month is incremented by 50 when winds are reported in knots versus m/s.  So a new global variable `scale_wspd` was added to signify a units conversion is needed.

### Issue(s) addressed

Working ctests and approval.  I checked many sites in USA vs. Europe to see that BUFR sondes and text sondes give the same results for wind speeds.

## Dependencies

None

## Impact

None expected.

### DIFF
--- a/src/conventional/sonde_tac2ioda.py
+++ b/src/conventional/sonde_tac2ioda.py
@@ -727,6 +727,7 @@ def getDayHour(str):
     Parse day of month and hour from the given string
     :param str: The string to parse
     :return: (day, hour)
+    When the day is coded as day+50, windspeed is in knots, otherwise m/s.
     """
     global scale_wspd
     try:
--- a/src/conventional/sonde_tac2ioda.py
+++ b/src/conventional/sonde_tac2ioda.py
@@ -60,6 +60,10 @@ LEVEL_CODES = {
 # List of sounding data sections
 TYPES = ['TTAA', 'TTBB', 'PPBB', 'TTCC', 'TTDD', 'PPDD']
 
+# Wind speed in radiosondes could be either knots or meters per second, which
+# is determined within code if the day of month is actually day of month + 50.
+scale_wspd = 0.51444   # convert from knots to meters per second
+
 STATIONS = {}
 
 # The outgoing IODA MetaData variables, their data type, and units.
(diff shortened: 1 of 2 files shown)


src/conventional/sonde_tac2ioda.py
```diff
@@ -724,10 +728,12 @@ def getDayHour(str):
     :param str: The string to parse
     :return: (day, hour)
     """
+    global scale_wspd
```
BenjaminRuston, line 731: confused why this specification of `global` is needed, as I read the code `scale_wspd` is already a global in this routine

src/conventional/sonde_tac2ioda.py
```diff
@@ -724,10 +728,12 @@ def getDayHour(str):
...
     try:
         day = int(str[:2]) - 50
```
BenjaminRuston, line 733: may be good to leave a comment that WMO codes for TAC TEMP messages notes that day of month incremented by 50 indicates winds in knots

gthompsnJCSDA: I documented in the comments of the PR, but documenting every little thing in the code directly would be a nuisance. So decide if this is a show-stopper or just approve as the list of things to get to match the WMO manual on codes would be purely crazy.

src/conventional/sonde_tac2ioda.py
```diff
@@ -723,11 +728,16 @@ def getDayHour(str):
...
+    When the day is coded as day+50, windspeed is in knots, otherwise m/s.
     """
+    global scale_wspd
     try:
         day = int(str[:2]) - 50
         if day < 0:
             day += 50
+            scale_wspd = 1.0
+        else:
+            scale_wspd = knots2mps
```
huishao-r, line 740: Just trying to understand this: so if the day is negative, the unit is mps; vice versa. Is it correct?

huishao-r: Ok. I understand now. They use "50" as a trick to distinguish the use of the different units. Thanks!

huishao-r (approved):
The code seems clear to me.

PatNichols (approved):

PR #1148, opened 2023-01-19
Bugfix:  launch hour can be inconsistent across the globe

gthompsnJCSDA:
## Description

Legacy/text (TAC) formatted radiosonde data has 2 digits for hour in the TTAA/TTBB/PPBB/etc. sections.  Some countries appear to populate it with the hour of the launch (which is typically 23z or 11z) whereas other countries push the hour to the intended "synoptic" time.  Since the actual correct launch time in hours and minutes is not provided in the raw text, we have to make a standard assumption that launch time is roughly 55 minutes before synoptic time.  For countries giving 11z for a 12z launch, that meant it was pushed back to 10:05UTC, which is worse.  By comparing dozens of sites with both BUFR and TAC formatted data, it is better if we properly assign synoptic time to the launch, then do the subtraction of the typical 55 minutes before.  This PR achieves this change.  Example use case:  Poland radiosondes were reporting launch hour as "11" whereas England, USA, and many other countries encode at "12" - even though the WMO manual of codes states it should be the hour of launch/release (but who in the weather world properly adheres to standards?).  Using Poland's stated hour as "11" was then giving the `releaseTime` as 10:05 which contradicted the correct release time of BUFR-formatted soundings that said something like "11:17" or "11:08", etc.  This change gives a better approximation in the majority of cases as some sites will still be assigned the wrong `releaseTime` no matter what solution is picked.

### Issue(s) addressed

A separate one was not created to estimated points were created based on work time.

Passing ctests and reviews.  The results were already compared between coincident BUFR and TAC radiosondes to decide this was problematic in the beginning.

## Dependencies

None

## Impact

None expected.


src/conventional/sonde_tac2ioda.py
```diff
@@ -961,6 +965,11 @@ def change_vars(profile):
 
     # SPECIAL:  Most soundings launched 50-55 minutes prior to stated synoptic time, so a 12Z
     # launch is usually initiated close to 11:05Z.
+    target_hr = target_date[11:13]
+    if target_hr == '12' and profile['hour'] == 11:
```
CoryMartin-NOAA, line 969: are you not worried about the (albeit fractionally sized) cases where balloons are launched at 06z or 18z?

CoryMartin-NOAA: What about converting target_hr to a datetime object and using timedelta, if it is -1 hour, then set profile['hour'], I think this would result in no need for an `elif`

gthompsnJCSDA: Agreed, I am worried about 6/18z a little. I pondered it, just not sure how reliable it will end up.  Going 100% to BUFR would totally help, but that is a wish-cast.

Meanwhile, I also thought about the datetime object in place of what I did here.  I will give it a go with some 00z data because I fear the TTAA YYGGi WMO_number stuff will have YY (day-of-month) set by synoptic time and GG which should be release time (23) misaligned.  I was surprised to read the WMO manual on codes says to put launch hour in the GG section and USA and other countries are setting that to synoptic hour, not actual launch hour.  Arghhh.  I'll check a bunch of worldwide 00z soundings and see what their pattern is.  Not following standards globally stinks.

PatNichols (approved):

src/conventional/sonde_tac2ioda.py
BenjaminRuston: we can just flat out reject the outliers for now, or specify a data quality flag

gthompsnJCSDA: @CoryMartin-NOAA  It turns out that I must do a stricter check on date/time because some reports centered at 00z contain the new day and 00 for hour, while others have the prior day and 23z (even 22z in some) which would then assign a completely wrong date.  I have the fix being tested now.

gthompsnJCSDA: Fixed.

PatNichols (approved):

BenjaminRuston:
was able to use the 2022-02-16T00:00:00Z file in latest develop

radiosonde airTemperature omb 2022021600

BenjaminRuston (approved):
was able to run and use in a test run, likely more to come but this is a great capability to have

PR #1149, opened 2023-01-20
Radiosondes: BUFR incoming data (from LDM), skip duplicates

gthompsnJCSDA:
## Description

BUFR radiosonde files (picked up with LDM in real-time) often contain duplicates - sometimes as many as three duplicates of the first message, more often only one dupe.  This PR causes the converter to capture which stations are found and skip past any subsequent soundings from the same site.  For a real-time example file, the output file size was decreased by 2/3rds when doing the skipping.

### Issue(s) addressed

Positive reviews and working ctests.

## Dependencies

None

## Impact

None

### DIFF
--- a/src/conventional/sonde_bufr2ioda.py
+++ b/src/conventional/sonde_bufr2ioda.py
@@ -42,8 +42,9 @@ locationKeyList = [
     ("wmoBlockNumber", "integer", "", "toss"),
     ("wmoStationNumber", "integer", "", "toss"),
     ("stationIdentification", "string", "", "keep"),
+    ("stationLongName", "string", "", "toss"),
     ("instrumentIdentifier", "integer", "", "keep"),
-    # ("instrumentRadiationCorrectionInfo", "integer", "", "keep"),
+    ("instrumentCorrectionInfo", "integer", "", "keep"),
     # ("instrumentHumidityCorrectionInfo", "integer", "", "keep"),
     # ("temperatureSensorType", "integer", "", "keep"),
     # ("humiditySensorType", "integer", "", "keep"),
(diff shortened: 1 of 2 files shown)


src/conventional/sonde_bufr2ioda.py
```diff
@@ -664,20 +669,39 @@ def read_bufr_message(f, count, start_pos, data):
...
                 meta_data['stationIdentification'][n] = "{:02d}".format(block) + "{:03d}".format(number)
+            elif meta_data['stationLongName'][n] != string_missing_value:
+                meta_data['stationIdentification'][n] = meta_data['stationLongName'][n]
+                if n == 0:
+                    logging.info(f"  this site lacks WMO number, using stationLongName: {meta_data['stationIdentification'][n]}")
             if n == 0:
                 count[3] += 1
                 logging.info(f"Processing sonde for station: {meta_data['stationIdentification'][n]}")
 
+        if meta_data['stationIdentification'][0] in stations_found:
```
srherbener, line 686: How do you tell the difference between a duplicate entry (same station, same location, same time), and a case where the sounding from a given station was split up into multiple subsets or multiple messages (same station, different location or time)?

BenjaminRuston: and by location you mean vertical @srherbener ,

this is something which is a loaded issues, yes there can be multiple report from a single station.  They may still separate the mandatory levels from the high resolution full level data.  At this point think the goal is to produce a single profile per station as best as we can for now.  Greg has been talking with NRL I believe on strategies and needs, as this converter should be readily used for their system

BenjaminRuston:
reduction of 2/3 !!!

src/conventional/sonde_bufr2ioda.py
gthompsnJCSDA: I have looked at dozens of profiles so far (in SkewT plots) and I have yet to see a visual difference of any kind before I was removing dupes.  So the dupes are most likely pure duplicates as far as I can see.  The only problem with this scheme is if we try to decode large blocks of time (more than 6 hours basically), because the combination of station ID and launch time is more difficult to find then just keeping track of found station via list and using python's "in" method.  It happens rather rarely that soundings are launched from the same site more frequently than 6 hours and the solution for us is simply use the converter once per hour or once per 2-3 hours to avoid the issue.

srherbener: @BenjaminRuston, thanks for the clarification.

Location in this comment is (lat, lon, vertical coordinate) to cover the case where drift is being reported.

I was thinking of duplication of a single observation, whereas I think you are saying in this context the duplication is at the report (complete sounding ?) level. I take it code in the other places of this converter loads up the stations_found entry (line 776 below) once a complete report is established (so that you know you are on a duplicate report when you see a match in stations_found).

My concern came from thinking of duplication of a single observation and wondering why the location and time aren't also being checked.

Am I on the same page now? Thanks!

gthompsnJCSDA: Steve, I think there is some unnecessary confusion here.  The duplicate elimination I'm doing is profiles, not single lat/lon/time checks, which I do not believe are needed - I've never seen a problem on that.

So if an entire Denver sounding is sent over the data circuits multiple times (and it happens so much that 2/3rds of the output file size is eliminated), then seconds, thirds, fourth distributions of the same full sounding is what I am eliminating.

srherbener: Got it. Thanks @BenjaminRuston and @gthompsnJCSDA for the clarifications. I think it all looks good now.

srherbener (approved):

PatNichols (approved):

PR #1151, opened 2023-01-25
Bugfix:  found times when converters crashed when testing with dozens of input files

gthompsnJCSDA:
## Description

When running an entire month of files through the radiosonde converters (BUFR and TAC input), they would sometimes crash.  So this PR is bug fixing.  Also, since the saturation water vapor pressure code is unreliable at super low temperatures (less than about -78C), the specificHumidity variable is limited to when dewpointTemperature is greater than -77C as a precaution.

### Issue(s) addressed

Approvals. The existing test data did not contain the kind of data that caused failure so no pre-existing ctest found the error.

## Dependencies

None

## Impact

None expected.

### DIFF
--- a/src/conventional/sonde_bufr2ioda.py
+++ b/src/conventional/sonde_bufr2ioda.py
@@ -692,16 +692,6 @@ def read_bufr_message(f, count, start_pos, data):
             meta_data['geopotentialHeight'][0] = float_missing_value
             meta_data['pressure'][0] = float_missing_value
 
-        # Very rarely the first level is way up high then followed by at surface. Toss first values.
-        if len(meta_data['pressure']) > 2:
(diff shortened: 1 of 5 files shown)


PatNichols (approved):

src/conventional/sonde_bufr2ioda.py
```diff
@@ -761,7 +751,7 @@ def read_bufr_message(f, count, start_pos, data):
         tvirt = np.full(target_number, float_missing_value)
         for n, temp in enumerate(airt):
             pres = meta_data['pressure'][n]
-            if (temp != float_missing_value and spfh[n] and pres < 108000 and pres > 10000):
```
BenjaminRuston, line 764: yup good catch @gthompsnJCSDA , it is your own converter

BenjaminRuston (approved):
thanks Greg this is a great foundation to build from

PR #1087, opened 2022-10-20
Common Calling Structure

ericlingerfelt:
## Description

This pull request allows all IODA converters written in python to be installable by pip and callable by module and package names. It also

- Adds the required `setup.py` file with all known python dependencies. 
- Adds the correct `__init__.py` files to the code.
- Adds the following code to `src/lib_python/orddicts.py` to manage the following deprecation so that python 3.10 can be used. See this [link](https://stackoverflow.com/questions/53978542/how-to-use-collections-abc-from-both-python-3-8-and-python-2-7).

```
try:
    from collections.abc import Callable
except ImportError:
    from collections import Callable
```
 
- Removes the following code from all files and includes the correct module imports to replace it. These classes can now be called with `lib_python` import statements. 

```
IODA_CONV_PATH = Path(__file__).parent/"@SCRIPT_LIB_PATH@"
if not IODA_CONV_PATH.is_dir():
    IODA_CONV_PATH = Path(__file__).parent/'..'/'lib-python'
sys.path.append(str(IODA_CONV_PATH.resolve()))
```

- Removes all unused imports. 
- Moves the script `generic_ob2ioda.py` to the top level in source. 
- Changes any files or directories in `src` to use `_` instead of `-` which is not allowed in python for modules or file names. 

Important notes: 

- This passes all ctests in `ioda-bundle` except `581 - test_iodaconv_airnow`. This is due to the converter's name being mismatched due to the change from `_` to `-`. The test file `test/testoutput/airnow_2020081306.nc` needs to have the top level attribute for converter updated to `airnow2ioda_nc.py` from `airnow2ioda-nc.py`.

### Issue(s) addressed

Link the issues to be closed with this PR
- fixes #<issue_number>

If this does not close an existing issue, then [be sure to add a Pipeline, Label, Estimate, Assignees, and Epic](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/practices/issues.html)

## Acceptance Criteria (Definition of Done)

What does it mean for this PR to be finished?

## Dependencies

If there are PRs that need to be merged before or along with this one, please add "waiting for another PR" label and list the dependencies in the other repositories (example below). Note that the branches in the other repositories should have matching names for the automated tests to pass.
Waiting on the following PRs:
- [ ] waiting on JCSDA/eckit/pull/<pr_number>
- [ ] waiting on JCSDA/atlas/pull/<pr_number>

## Impact

[...]

Issue #1158, opened 2023-02-02
[Bug] Python path is wrong and ctests are failing. 


PatNichols:
## Current behavior (describe the bug)
When running ctests, there are many failures. Ioda converter python files contain imports from lib_python which does not exist.
Typical error look like:
567/820 Testing: test_iodaconv_gds2_sst_l2p
567/820 Test: test_iodaconv_gds2_sst_l2p
Command: "/usr/bin/bash" "/home/ubuntu/ib_test/bld/bin/iodaconv_comp.sh" "netcdf" "/usr/bin/python3.8 /home/ubuntu/ib_te
                          -i testinput/gds2_sst_l2p.nc
                          -o testrun/gds2_sst_l2p.nc
                          -d 2018041512
                          -t 0.5" "gds2_sst_l2p.nc" "0.0"
Directory: /home/ubuntu/ib_test/bld/iodaconv/test
"test_iodaconv_gds2_sst_l2p" start time: Feb 02 17:40 UTC
Output:
----------------------------------------------------------
/usr/lib/python3/dist-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.26.12) or chardet (3.0.4)
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
Traceback (most recent call last):
  File "/home/ubuntu/ib_test/bld/bin/gds2_sst2ioda.py", line 19, in <module>
    import lib_python.ioda_conv_engines as iconv
ModuleNotFoundError: No module named 'lib_python'
<end of output>
Test time =   0.30 sec
----------------------------------------------------------
Test Failed.
"test_iodaconv_gds2_sst_l2p" end time: Feb 02 17:40 UTC
"test_iodaconv_gds2_sst_l2p" time elapsed: 00:00:00

## To Reproduce
Run the ctests in ioda bundle.

## Expected behavior

Ctest should pass.

PatNichols:
Basically this:IODA_CONV_PATH = Path(__file__).parent/"@SCRIPT_LIB_PATH@"
if not IODA_CONV_PATH.is_dir():
    IODA_CONV_PATH = Path(__file__).parent/'..'/'lib-python'
sys.path.append(str(IODA_CONV_PATH.resolve()))

import meteo_utils
import ioda_conv_engines as iconv
from collections import defaultdict, OrderedDict
from orddicts import 

Was replaced with:

import lib_python.ioda_conv_engines as iconv
[...]

PR #1160, opened 2023-02-02
Revert "Common Calling Structure"

PatNichols:
Reverts JCSDA-internal/ioda-converters#1087


PatNichols:
@BenjaminRuston @fcvdb This should fix our problems for now.

BenjaminRuston (approved):
would have preferred to have moved forward but seems like we need to discuss library names first

srherbener (approved):

ericlingerfelt (commented):
I do not believe that we should revert https://github.com/JCSDA-internal/ioda-converters/pull/1087. I think it would be best to create a new bug fix PR that responds to the issues that our users are experiencing.

srherbener:
> I do not believe that we should revert #1087. I think it would be best to create a new bug fix PR that responds to the issues that our users are experiencing.

I think this is a very good point. However, the issue tied to this PR #1158 is suggesting that the develop branch is seriously broken and we need to restore develop to health quickly. But, restoring all of the work in the original feature branch might be difficult.

Would it work to merge this PR into develop now and then go back to the original PR and cherry pick the commits into a new PR. It would be painful if all of that work in the original PR would have to be redone, which is why I'm guessing @ericlingerfelt is suggesting to not merge this PR. 

@climbfuji can you help recommend how we should proceed? Thanks!

srherbener:
Or can we simply merge this PR and then revert it which should re-create the state we are in now on a new feature branch. Then fix the packaging/PYTHONPATH issues on the new feature branch before issuing a PR.

climbfuji:
I agree with @ericlingerfelt that reverting the PR is not a good idea. Decide on a name for the library quickly (i.e. use `lib_iodaconv` instead of `lib_python`), then fix the handful of files (probably less than five), update the documentation and share the information (that's what didn't happen apparently when the original PR was merged) with the developers and users.

climbfuji (dismissed):
Recommend to fix in develop instead of reverting

PatNichols:
@climbfuji @ericlingerfelt I might be in favor of a bugfix branch provided it can be applied quickly ( like in a hour or so). The ioda-develop is failing multiple ctests. This in turn means all the bundles are broken.

PatNichols:
@climbfuji The number of files to fix depends on how you fix it. There are probably dozens if it's just a source code fix. A CMakefile fix might be quicker.

PatNichols:
@climbfuji  Look up to see how many files are changed to get an idea of how many to fix in naive way.

climbfuji:
But wouldn't the easiest way to "for now fix the ctests" be to simply change the ctest setup to append the missing directory to `PYTHONPATH`? This gives you time to think about a better name and update the documentation for users and developers. And it only touches a handful of files, not 100+.

climbfuji:
@PatNichols This change in the iodaconv repository fixes the failing ctest for me (tried on my mac):
```
heinzell@JCSDA-L-18146:~/work/ioda-converters/ioda-bundle/iodaconv [brew-x86_64]> git diff
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 860a21f8..c1faa990 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -407,7 +407,12 @@ set(IODA_CONV_COMP_TOL "0.5e-4")
 #        ENVIRONMENT "PYTHONPATH=$<TARGET_FILE_DIR:_ioda_python>/../"
 #        ...
 #        )
-set(IODACONV_PYTHONPATH "$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
+message (INFO "/Users/heinzell/work/ioda-converters/ioda-bundle/iodaconv/src")
+message (INFO "CMAKE_CURRENT_SOURCE_DIR is ${CMAKE_CURRENT_SOURCE_DIR}")
+message (INFO "<TARGET_FILE_DIR:_ioda_python>/../ is $<TARGET_FILE_DIR:_ioda_python>/../")
+set(IODACONV_PYTHONPATH "${CMAKE_CURRENT_SOURCE_DIR}/../src:$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
+#set(IODACONV_PYTHONPATH "$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
+#set(IODACONV_PYTHONPATH "${CMAKE_CURRENT_SOURCE_DIR}/../src:$ENV{PYTHONPATH}")

 #===============================================================================
 # Marine converters
heinzell@JCSDA-L-18146:~/work/ioda-converters/ioda-bundle/iodaconv [brew-x86_64]> git diff
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 860a21f8..47452eaa 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -407,7 +407,10 @@ set(IODA_CONV_COMP_TOL "0.5e-4")
 #        ENVIRONMENT "PYTHONPATH=$<TARGET_FILE_DIR:_ioda_python>/../"
 #        ...
 #        )
-set(IODACONV_PYTHONPATH "$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
+#
+# DH 20230202 - need to add a bit more documentation here why to add ${CMAKE_CURRENT_SOURCE_DIR}/../src
+#
+set(IODACONV_PYTHONPATH "${CMAKE_CURRENT_SOURCE_DIR}/../src:$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")

 #===============================================================================
 # Marine converters
```
Can you test this, and if it works issue a PR for ioda-converters? 

I'll unblock this PR but I think it's no longer needed

climbfuji (commented):
Unblocking this PR

climbfuji (commented):
Unblocking this PR, hope this works.

PatNichols:
Now that I have time to breathe i will give it a shot. We still need to make the " no lib_python" folks happy bit ...

climbfuji:
> Now that I have time to breathe i will give it a shot. We still need to make the " no lib_python" folks happy bit ...

Sure, that can be done in a second step. There are probably also documentation/documentation updates needed for users who point to the Python scripts in ioda-converters (but I haven't checked how much is there already and how it needs to change).

PR #1162, opened 2023-02-04
Add ATMS remap option.

junmeiban:
## Description

Add ATMS remap option: remapping ATMS channel 1 and 2 with 5.2degree beamwidth to AMSU 3.3degree beamwidth.

The file "atms_BGremap_coeffs_ch1ch2.nc" includes coefficients which are offline calculated.   
Files modified/added:
A       src/hdf5/apply_BG/apply_BG.py
M       src/hdf5/atms_netcdf_hdf5_2ioda.py
M       test/CMakeLists.txt
A       test/testinput/SNDR.J1.ATMS.20180415T0000.nc4
A       test/testinput/atms_BGremap_coeffs_ch1ch2.nc
A       test/testoutput/2018041500_atms_BGRemap_sdr.nc4

Reference:
J. Zhou, H. Yang and R. Iacovazzi, "Improving ATMS Remapping Accuracy Using Adaptive Window and Noise-Tuning Method in Backus–Gilbert Inversion," in IEEE Transactions on Geoscience and Remote Sensing, vol. 60, pp. 1-12, 2022, Art no. 5304412, doi: 10.1109/TGRS.2022.3182630.

Original source code written by @JunUMD can be found at: https://github.com/JunUMD/BG_AdaptiveWindow_v1

### Issue(s) addressed
https://github.com/JCSDA-internal/ioda-converters/issues/1166

## Dependencies
N/A

This figure shows the BT (channel 1) distribution before remap:
noremap_2018041500

This figure shows the BT (channel 1) differences between after remapping and before remapping:
diff_global

How to use this function:
Assume the following files (in +-3hr time window) in your dir "2018041500":
```
SNDR.J1.ATMS.20180414T2100.m06.g211.L1B.std.v03_15.G.210521052327.nc4
SNDR.J1.ATMS.20180414T2106.m06.g212.L1B.std.v03_15.G.210521052327.nc4
...
SNDR.J1.ATMS.20180415T0248.m06.g029.L1B.std.v03_15.G.210521052026.nc4
SNDR.J1.ATMS.20180415T0254.m06.g030.L1B.std.v03_15.G.210521052026.nc4
```
python $scriptDir/atms_netcdf_hdf5_2ioda.py -m BG -d 2018041500 -i $inputDir/2018041500/* -o  atms_n20_obs_2018041500.h5


BenjaminRuston:
@junmeiban great, and thanks for getting this rolling.  We will need to figure out where these coefficients should live and additionally if it's possible to move away from pickle files

hailingz:
I did not review it scientifically but I have fixed all the code norms. Now all tests passed  and I think this PR is ready to for reviewers. @liujake @BenjaminRuston

hailingz:
reference file (in testouput) needs to be updated.
please see test/CMakeLists.txt and look for "TARGET  test_iodaconv_atms_nc4"

src/hdf5/apply_BG/apply_BG.py
```diff
@@ -0,0 +1,181 @@
...
+from pathlib import Path
+import os, glob, time
+import pickle, sys
+import h5py
+import numpy as np
+import statistics
+from scipy.interpolate import griddata
+
+
+class apply_BG_class:
```
srherbener, line 19: Is the `apply_BG` functionality something that is specific to ATMS data, or is `apply_BG` something that could be used for many instrument type. I'm wondering if `apply_BG` belongs here (specific to ATMS) or if it possibly belongs in UFO somewhere instead (can apply to many instruments).

BenjaminRuston:
it looks like Xarray support pickle to netCDF could we potentially convert these pickle files using this:

https://docs.xarray.dev/en/stable/user-guide/io.html#reading-and-writing-files

src/hdf5/apply_BG/apply_BG.py
junmeiban: Sorry for the late reply @srherbener. apply_BG can also be used for other instruments.

PatNichols (dismissed):
We cannot use pickle files. It's a huge security hole. Please convert these to netcdf or some other format.

hailingz:
Hi @srherbener 
@junmeiban added a ctest "test_iodaconv_atms_BGRemap_nc4" that uses a input files of 6.74M (testinput/SNDR.J1.ATMS.20180415T0000.nc4)
Please advise if this size is OK or not.
Otherwise this PR looks good to me.

hailingz (approved):
Thank you for having this option Junmei!

src/hdf5/atms_netcdf_hdf5_2ioda.py
```diff
@@ -393,6 +400,83 @@ def set_obspace_attributes(VarAttrs):
     return VarAttrs
 
 
+def remapBG(input_files):
```
BenjaminRuston, line 403: could you move this remapBG also to `apply_BG.py`

src/hdf5/atms_netcdf_hdf5_2ioda.py
```diff
@@ -393,6 +400,83 @@ def set_obspace_attributes(VarAttrs):
...
+
+    print('Remap ATMS: Backus–Gilbert Inversion method')
+
+    obs_data = init_obs_loc()
+
+    src_ch = [1, 2]
+    tgt_ch = [1, 2]
+    src_bm = 5.2
+    tgt_bm = 3.3
+    nfov = 96
```
BenjaminRuston, line 413: maybe optional argument saying sensor is ATMS, as nfov 96 is specific to this instrument... also if the src_bm and tgt_bm is in the coefficient file it could be checked here as well.

src/hdf5/apply_BG/apply_BG.py
srherbener: Sorry for my late reply too. I lost track of this PR.

I agree that it is good to place apply_BG and related functions in a separate python file/module so that this code can be shared. Does it seem that `src/hdf5` is the place for this, or would it make sense to move to `src/lib-python`?

Thanks for taking care of this!

srherbener (dismissed):
The data files for new ctest are quite large in that the resulting output ioda file contains 12960 locations. For ctests (unit level testing) we are going for code coverage, as opposed to scientific validation, which means that we want very small tests that exercise the code and run fast.

We have been typically thinning test data down to 100s of locations which allow for many tests to run in under 1 second. The log for the clang CI test show the new test running in 21 seconds.

Please see if there is a way to trim the test data down so that the test can run much faster (say 2 seconds or faster). Thanks!

junmeiban:
> The data files for new ctest are quite large in that the resulting output ioda file contains 12960 locations. For ctests (unit level testing) we are going for code coverage, as opposed to scientific validation, which means that we want very small tests that exercise the code and run fast.
> 
> We have been typically thinning test data down to 100s of locations which allow for many tests to run in under 1 second. The log for the clang CI test show the new test running in 21 seconds.
> 
> Please see if there is a way to trim the test data down so that the test can run much faster (say 2 seconds or faster). Thanks!

@srherbener , thank you so much for your suggestion. Do you have a tool/script to reduce the file? The scan line is 135, I will reduce it.  But I need keep 96 FOVs.

srherbener:
@junmeiban take a look at: https://stackoverflow.com/questions/46373359/how-can-i-reduce-the-dimension-size-in-a-netcdf-file

You should be able to use ncks to extract a block of locations, or to extract every nth location, etc. Ncks is part of the nco tools which should be a module in the JEDI spack-stack.

src/hdf5/atms_netcdf_hdf5_2ioda.py
```diff
@@ -387,6 +393,83 @@ def set_obspace_attributes(VarAttrs):
...
+    satzen = []
+    satazi = []
+    solzen = []
+    solazi = []
+    tim = []
+    taAllCh = []
+    viewang = []
+    [viewang, taAllCh, lat, lon, satzen, satazi, solzen, solazi, tim] = apply_BG_obj.apply()
+
+    g = h5py.File(input_files[0], 'r')
```
BenjaminRuston, line 437: think this line can be removed?

junmeiban: > think this line can be removed?

In line 444, "g" is needed for "sensorChannelNumber".

PatNichols (approved):
Thank you for making the changes. Looks good.

srherbener (approved):
@junmeiban thanks for trimming down the test file. The test is running in 2 - 3 seconds now which is great!

src/hdf5/atms_netcdf_hdf5_2ioda.py
junmeiban: > maybe optional argument saying sensor is ATMS, as nfov 96 is specific to this instrument... also if the src_bm and tgt_bm is in the coefficient file it could be checked here as well.

Hi @BenjaminRuston , do you mean add, such as, "optional.add_argument(
        '-s', '--sensor',"....) for ATMS?

PR #1163, opened 2023-02-05
Bugfix: Many BUFR radiosonde data lack explicit lat/lon data but it can be estimated from wind data

### LINKED ISSUE #1157: Bug: `src/conventional/bufr2ioda.py` needs further improvements especially for missing lat/lon data

Even with BUFR data for radiosondes, its creators might still not include any lat/lon data for the launch location. Insanity.  So we really need to use a lookup table of station locations as a desperate measure to keep data from being useless.  Furthermore, some locations have lat/lon at launch location but a singular value for the balloon ascension which is also sub-optimal.  Since the legacy/text/TAC converter already has a way to best-guesstimate the trajectory, we can also help fix this problem to a moderate level of accuracy. 

## Expected behavior

Go to much greater lengths to improve the capture of lat/lon/time data during the balloon ascent when these data are not explicitly provided.  I have code in progress now to fix in branch `bugfix/sonde_fix_missing_latlon`

gthompsnJCSDA:
## Description

Legacy/text/TAC formatted radiosonde data contains no explicit lat/lon so `src/conventional/sonde_tac2ioda.py` code was written to estimate the balloon ascent/trajectory using reported wind data.  It was a big surprise to find so many BUFR encoded radiosondes also lack explicit lat/lon data and provide solely a WMO station ID.  Therefore, the new code does the following:

1. Uses a static look-up table of radiosonde sites to retrieve latitude, longitude, and station elevation
2. Provide balloon ascent timestamps assuming 5 m/s ascent rate and lat/lon from reported wind data
3. Set the new `meta_data['positionEstimated']` variable to 1 for true and zero when not estimating position.

### Issue(s) addressed

- closes #1157 

Positive reviews and passing ctests

## Dependencies

Waiting on the following PRs:
- [ ] waiting on JCSDA-internal/ioda/pull/899

## Impact

None expected


PatNichols (approved):
looks good

fcvdb (approved):
I built the branch and all the tests passed. In particular, I ran the following two test:
ctest -VV -R test_iodaconv_bufr_read_wmo_radiosonde
ctest -VV -R test_iodaconv_sondetac

Issue #791, opened 2022-02-10
Create Subset 080 - NOAA-15/18/19 YAML for AVHRR IRLW


AshleyStanfield-NOAA:
## Description
>Generate and validate NOAA-15/18/19 AVHRR  IRLW YAML.

## Requirements
>Are we using the correct nomenclature?
>How does it compare before and after YAML implementation? Does it match up?
>Does it pass validation tests?

AshleyStanfield-NOAA:
Thus far, the YAML file appears to be working and matching up with equal counts (23,232).

### Heat Maps showing report locations:
> ![Yaml80_Heat_Map1_80](https://user-images.githubusercontent.com/92459593/154376507-16187fb2-4a14-46a2-961e-a1f462e82e77.png) ![Yaml80_Heat_Map2_80](https://user-images.githubusercontent.com/92459593/154376571-bcdb952d-cbd0-47f5-b9ca-2754abc3485b.png)

### SZA Comparisons
> ![Yaml80_sza_Map_80_yaml](https://user-images.githubusercontent.com/92459593/154376687-7c08736c-e6ad-4436-8e1a-bedcea1ee793.png) ![Yaml80_sza_Map_80_v2](https://user-images.githubusercontent.com/92459593/154376692-c3501c6c-8c90-4ef9-92ed-13334dacd49e.png)

### Wind Speed Comparisons:
> ![Yaml80_wndspd_Map_80_yaml](https://user-images.githubusercontent.com/92459593/154376905-4ca400b3-a1a4-4656-a507-443b36c95dbd.png) ![Yaml80_wndspd_Map_80_v2](https://user-images.githubusercontent.com/92459593/154376914-073ef82d-911e-4e1b-8a42-5a0299e9e0d8.png)

### Pressure Comparisons:
> `
![Yaml80_pres_Map_80_yaml](https://user-images.githubusercontent.com/92459593/154377142-02225b20-2d08-47e9-85d5-348f02593874.png) ![Yaml80_pres_Map_80_v2](https://user-images.githubusercontent.com/92459593/154377150-05406ccc-e634-4df9-9a41-18418e8c3be8.png)

BenjaminRuston:
@AshleyStanfield-NOAA are you still working through this?  If complete please close, and if you would like to create a new issue for further development or problems which arise should be fine.

PR #1165, opened 2023-02-10
update GNSSRO fortran converter following the new name convention

hailingz:
## Description

This PR is to update the GNSSRO fortran converter following the new name convention. 
Other changes include (1) remove unused variable "time" and related codes; and (2) add compress ability. The file size with the "compress" is reduced by about 4-5 times.
 
Notes:
a. A couple of variables were not renamed during the name convention sprint as listed below. Could be a future work.
process_center
     originally data processing center
occulting_sat_is
   satellite instrument
b. Variable "satelliteConstellationRO" could be renamed in the future as this variable is the GNSS satellite constellation rather than RO. It is confusing now.

Quick fix, no issue created.

### Issue(s) addressed
N/A

(1) pass test.
(2) processed observation files are tested with fv3jedi 3denvar and compared with the file upgraded by ioda-upgrade-v2-to-v3.x (done)

What does it mean for this PR to be finished?
This fortran converter generates IODA V3 files with the new name convention. ioda-upgrade-v2-to-v3.x is not needed any more.

## Dependencies
N/A

## Impact
N/A


src/gnssro/gnssro_bufr2ioda.f90
```diff
@@ -1,17 +1,18 @@
...
+!  Last upate: Feb 9 2023
 
 !!!---------  to run   -----------------------------------------------------------------
 !  ./gnssro_bufr2ioda2 yyyymmddhh $bufrfile_input $netcdffile_output
 !
 ! Return codes:
 !  0 - Success.
 !  1 - Unrecoverable system or logical error.
-!  2 - All provided observations are invalid.  Cannot create NetCDF IODA2 file.
+!  2 - All provided observations are invalid.  Cannot create NetCDF IODA3 file.
```
BenjaminRuston, line 15: @hailingz `IODA3` is not an official name should we just remove this and say `IODA`?

hailingz: you got it. Done!

src/gnssro/gnssro_bufr2ioda.f90
```diff
@@ -320,129 +310,189 @@ program gnssro_bufr2ioda2
     stop 2
 endif
 
+chunksizes=(/ndata,ndata/)
```
srherbener, line 313: It can be a bit tricky finding the optimal chunk size. There are a number of things to consider, such as the impact on runtime and compression. There exists a chunk cache and one thing you don't want to happen is using a chunk size that is larger than the chunk cache - which will kill performance. This won't happen in the small ctest case, but setting the chunk size to the number of locations could cause the chunk size to exceed the chunk cache size with more realistic datasets to be run through DA flows. 

The netcdf/hdf libraries utilize a default chunking strategy if none is specified. I think for now it would be good to use the built in default and not set the chunking sizes.

However the default library chunking strategy cannot take advantage of access patterns we might be aware of that lead to better chunking strategies. I think an investigation into an optimal chunking strategy for JEDI obs data is warranted (there is an issue in ZenHub for this: #791) but that is work that isn't in the plans yet.

My recommendation is to remove the settings of chunk sizes now and investigate this at a later time. Thanks for considering this.

hailingz: Thank you for the suggestion.  chunksize is removed now.

srherbener (approved):
Thanks for addressing my comments!

huishao-r (approved):
Agree with Hailing (in a separate discussion), this "satelliteConstellationRO" is not accurate and might need to revise in the future.

BenjaminRuston (approved):
great thanks for the updates appreciate it

PR #1168, opened 2023-02-17
Fix AOD converters

jeromebarre:
## Description

- Make sure the converter are producing the correct obs format to be handled correctly by ioda and oops. 
- Removing non-pythonic for loops, which speeds up the converters, relevant for viirs @CoryMartin-NOAA.

### Issue(s) addressed

- fixes partially https://github.com/JCSDA-internal/jedi-bundle/issues/29

### DIFF
--- a/src/compo/modis_aod2ioda.py
+++ b/src/compo/modis_aod2ioda.py
@@ -130,9 +130,7 @@ class AOD(object):
             sen_zen = sen_zen[pos_index]
             unc_land = unc_land[pos_index] * 1E-3  # see scale factor
             modis_time = modis_time[pos_index]
-            obs_time = np.full(len(modis_time), long_missing_value, dtype=np.int64)
-            for n, t in enumerate(modis_time):
-                obs_time[n] = round(t)
+            obs_time = np.full(len(modis_time), np.around(modis_time), dtype=np.int64)
 
             # uncertainty estimates:
             # From MODIS file (over ocean) and Levy, 2010 (over land)
--- a/src/compo/modis_aod2ioda.py
+++ b/src/compo/modis_aod2ioda.py
@@ -43,7 +43,8 @@ AttrData = {
 DimDict = {}
 
 # A dictionary of variable names and their dimensions.
-VarDims = {'aerosolOpticalDepth': ['Location']}
+VarDims = {'aerosolOpticalDepth': ['Location', 'Channel']}
+channels = [4]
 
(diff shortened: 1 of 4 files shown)


mer-a-o (approved):
Changes look good. I didn't test them though. Thanks for updating the converters.

PatNichols:
@jeromebarre @BenjaminRuston  I believe that the loop in the original code was to protect against missing value problems for some inputs that are not in the ctests. Ben might remember better than I. Are we sure that will not occur if we take it out?

PatNichols:
src/compo/modis_aod2ioda.py is the file I am talking about.

srherbener (approved):
Thanks for fixing these!

jeromebarre:
> @jeromebarre @BenjaminRuston I believe that the loop in the original code was to protect against missing value problems for some inputs that are not in the ctests. Ben might remember better than I. Are we sure that will not occur if we take it out?

How can a loop like this:
`            for t in range(len(obs_time)):
                obs_time[t] = base_datetime`
help against missing values?

src/compo/modis_aod2ioda.py
```diff
@@ -129,9 +130,7 @@ def _read(self):
             sen_zen = sen_zen[pos_index]
             unc_land = unc_land[pos_index] * 1E-3  # see scale factor
             modis_time = modis_time[pos_index]
-            obs_time = np.full(len(modis_time), long_missing_value, dtype=np.int64)
-            for n, t in enumerate(modis_time):
-                obs_time[n] = round(t)
+            obs_time = np.full(len(modis_time), np.around(modis_time), dtype=np.int64)
```
BenjaminRuston, line 133: @jeromebarre this look correct this is a replacement for the two preceding lines

I think the assumption here is the `modis_time` is just a single scalar?

BenjaminRuston: oh no I see now... it's where `aod` is defined

BenjaminRuston: regardless line 133 should provide the same behavior as before the issue is if a valid `pos_index` where `aod` is greater than 0 may have a missing time which will not get set.  happy to punt on this for now?

BenjaminRuston (approved):
this should get you back to where you were, should potentially look into a trap for missing:

`modis_time = hdf.select('Scan_Start_Time')[:].ravel()`

PR #1189, opened 2023-03-09
iso_fortran_env modernization

PatNichols:
Description
The iso_fortran_env intrinsic module standardized real and integer types. There are places in the jedi code that try to define these independent of the fortran compiler. This PR is an attempt to use iso_fortran_env standards without modifying much code. All types in. the kinds.f90 are defined in these terms which allows us to stave off technical debt. This is done so that the kinds defined are independent of which fortran compiler is used.

Requirements
None

Acceptance Criteria (Definition of Done)
All ctests and code runs normally.

Dependencies
None.

### Issue(s) addressed

- fixes


src/ncar-bufr2nc-fortran/kinds.f90
```diff
@@ -36,38 +38,43 @@ module kinds
...
 ! Integer types
-  integer, parameter, public  :: i_byte  = selected_int_kind(1)      ! byte  integer
-  integer, parameter, public  :: i_short = selected_int_kind(4)      ! short integer
-  integer, parameter, public  :: i_long  = selected_int_kind(8)      ! long  integer
-  integer, parameter, private :: llong_t = selected_int_kind(16)     ! llong integer
-  integer, parameter, public  :: i_llong = max( llong_t, i_long )
+  integer, parameter, public  :: i_byte  = int8       ! one byte integer
+  integer, parameter, public  :: i_short = int16      ! short integer
+  integer, parameter, public  :: i_long  = int64      ! long  integer = 64 bit on LP64 systems
+  integer, parameter, public  :: i_init = int32       ! 32 bit integer
```
fcvdb, line 52: Shouldn't be `i_int`?

fmahebert:
I really like the idea of using `iso_fortran_env` to standardize these types; I would like even more to see a discussion of adopting it across the JEDI system so the way we use `kind` remains consistent (as much as possible) throughout our repos. What do you think?

src/gnssro/gnssro_bufr2ioda.f90
```diff
@@ -13,638 +13,637 @@
...
+   use, intrinsic :: iso_fortran_env
+   use netcdf
+   implicit none
 
-integer, parameter :: i_kind  = selected_int_kind(8)    !4
-integer, parameter :: i_64    = selected_int_kind(10)   !8
-integer, parameter :: r_kind  = selected_real_kind(15)  !8
+!integer, parameter :: i_kind  = selected_int_kind(8)    !4
+!integer, parameter :: int64    = selected_int_kind(10)   !8
+!integer, parameter :: real64  = selected_real_kind(15)  !8
```
fmahebert, line 23: are these commented-out lines needed?

src/ncar-bufr2nc-fortran/kinds.f90
```diff
@@ -3,11 +3,12 @@ module kinds
...
 ! module:   kinds
 !   prgmmr: treadon          org: np23                date: 2004-08-15
+!   prgmer: pnichols                                  data: 2023-03-09
 !
 ! abstract:  Module to hold specification kinds for variable declaration.
-!            This module is based on (copied from) Paul vanDelst's 
+!            This module is based on (copied from) Paul vanDelst's
 !            type_kinds module found in the community radiative transfer
-!            model
+!            model. Modifed to take
```
fmahebert, line 11: - Some of this sentence is missing
- Could most/all of kinds.f90 be deleted, if the iso types are being used instead?

src/ncar-bufr2nc-fortran/kinds.f90
PatNichols: @fcvdb Yes it has been changed now. Thank you for the catch.

src/gnssro/gnssro_bufr2ioda.f90
PatNichols: @fmahebert That was stripped out of the new version of this file. I left it in the old one until I can ascertain why it was failing. Turned out there were line contrinuations, "&", in the middle of strings. Thanks for the catch.

PatNichols:
> I really like the idea of using `iso_fortran_env` to standardize these types; I would like even more to see a discussion of adopting it across the JEDI system so the way we use `kind` remains consistent (as much as possible) throughout our repos. What do you think?

I would love that! My strategy right now is to edit any XXXkind.f90  file with iso_fortran_env definitions instead of selected_XX_kinds and such. Files that define kinds will need to be changed inside the file itself. Talking others into it might be a feat but we can at least get the low hanging fruit.

src/bufr/BufrParser/Query/ResultSet.cpp
```diff
@@ -100,28 +100,6 @@ namespace bufr {
             if (groupByField != "")
             {
                 groupByFieldIdx = dataFrames_[0].fieldIndexForNodeNamed(groupByField);
-
```
CoryMartin-NOAA, line 103: why are all of these C++ lines changing when the PR is for `iso_fortran_env`?

FreddyEngelkingMei:
It seems like this is a technical conversation regarding a specific pull request for a codebase written in Fortran. The pull request aims to standardize the real and integer types in the codebase by using the iso_fortran_env intrinsic module.

CoryMartin-NOAA (approved):

BenjaminRuston (approved):
thanks Pat this is a good step in the right direction we'll need to get back to this in the future

Issue #1180, opened 2023-03-01
BUFR Query: Refactoring Part 1


rmclaren:
The initial version of the bufr ioda ingestion (or conversion) code was written in pure Fortran and then translated into C++. The C++ code retains many of the aspects the original Fortran code, which was written in a functional style with very limited use of more advanced programming techniques. The result is code that does most of its work with a lot of nested for loops and if statements. This works but it does make the code hard to understand, and it makes extending it difficult (this style of code tends to be inflexible). The purpose of this issue is to refactor the initial stages of the code in order to make more elaborate queries possible and to refactor the target identification code (the code that finds the unique identity for a BUFR node given a query string) to enable future enhancements and fixes.

So basically the following major components will be updated (the supporting components will change as well):

**QueryParser** - This code is responsible for parsing a query string into the information the target finding code needs. Currently this code is relatively ridged and expects the query string to have a very specific format (for example indices are always at the very end, etc...). The refactored version uses a Tokenizer to find parse the query string into a list of tokens that can be evaluated later. This enables query syntax that is a bit more elaborate and accurate. So now (as an example) you can put an index anywhere in the query string. Its also easier to add new query elements (like "{}" for filters) in the places where this makes sense. So for example we can have strings like `*/QCPRMS[3]{2-3}/GNAP[1]` (purely a notional example where you have multiple repeated sequence ). This would have been impossible before.

**SubsetTable** - Currently its being used to discover all the queries that are possible given a BUFR file by looping through all the messages and "walking" the BUFR meta data tree. Its currently only used by the query_printer executable. The refactored version walks the tree, but encodes the result into a tree of BufrNode objects. Once this tree is created we can ask it questions which it can answer via some recursive algorithms. Overall this simplifies the code and makes the SusbetTable a lot more useful to other objects.

**QueryRunner** - Query runner is responsible for for target identification in each new subset, and then collecting the data associated with the target in the current subset. It is used to accumulate the data inside a a BUFR message subset. This refactoring updates that the target finding alogorithm. Instead of walking the BUFR meta data tree directly it now uses the SubsetTable object to find what it needs to uniquely identify the target. Note: the data collection code (code that actually gathers the data in the BUFR message) will be refactored later.

rmclaren:
Currently refactored code lives in `feature/query_filters`.

Issue #967, opened 2022-06-06
Indicate an index in the query string to reduce multi-dimensional arrays to 1-D


nicholasesposito:
## Description
It seems when the bufr2ioda code sees a sequence with brackets, it will make a dimension of it, whether or not there are multiple replications. For my results, there's lots of length of 1 dimensions. For example, for my aircraft bufr, the dimensions look like this:

netcdf aircraft_prepbufr_select_thinned {
dimensions:
        CloudSequence = 1 ;
        HeightEvent = 1 ;
        HumidityEvent = 1 ;
        Location = 13664 ;
        PresentWeatherSequence = 1 ;
        PressureEvent = 1 ;
        TemperatureEvent = 1 ;
        WindEvent = 1 ;

We were wondering, is there a way to flatten the data into a 1-D array when the non-Location dimension length =1. For example, since WindEvent =1 in the case above, instead of having a variable's dimensions be (Location, WindEvent), it could be flattened into (Location).

Having a way of indicating an index (or set of indices) in the query string which would let you do this. Right now there is no way to do this. Repeated things (or things where there is a potential for a repeat will always add an extra dimension

## Requirements

Any fix would reduce the number of dimensions and "flatten" the data to a 1-D for any multi-dimensional array where the length of one of the dimensions is 1. 

## Acceptance Criteria (Definition of Done)
Files and yamls affected would need to be edited and tested with the updates.

## Dependencies
None.

rmclaren:
Implementing a way to specify a "column" filter as part of a query string. So for example you can do something like this `*/BRIT{1, 3-5}/TMBR`. '{1, 3-5}' is a filter and lets you specify the "columns" (really repeats of the sequence BRIT) you want to keep in the output data. If you do something like this `*/BRIT{2}/TMBR` it means we can drop the extra dimension associated with the BRIT sequence since we are only keeping 1 repeats worth of information.

PR #1192, opened 2023-03-18
Query: Refactoring Part 1 with query filtering.

rmclaren:
## Description

This pull request primarily deals with 2 things:

1. Refactoring the code from the Query Parsing function (`QueryParser::parse`) up to (and including) the code that identifies the targets in the BUFR data (`QueryRunner::findTargets`). Some significant changes are listed bellow:

- `Tokenizer` class was added. It is responsible for taking a raw string the user provided and parsing it into a list of `Token` objects.
- `QueryParser` class is rewritten to use the tokenizer. It parses the `Token` objects into a list of `QueryComponent` objects and stores the result into a `Query` object.
- `SubsetTable` class was rewritten. It now re-interprets the BUFR meta data into into a recursive tree where each node represents a data element. This tree can then be interrogated to get the information we need to do our job.
- 'QueryRunner::findTargets' method was rewritten. It now uses `SubsetTable` to do its work (previously had implemented a similar, but slightly different process). `Targets` now stores `TargetComponents` which map to the different components in the query.
- Because of the new code above there is a fix/improvement in some query strings. Some BUFR tables list repeated mnemonics multiple times (ex: satwind data often lists the QCPRMS mnemonic multiple times...). Before the index for the repeat was listed at the end of the query (ex: `*/QCPRMS/GNAP[4]`). This was kinda incorrect and misleading giving the impression that GNAP was the thing that is repeated in the BUFR table. It also meant that it couldn't handle situations where there were multiple places with repeating mnemonics in a query path. The legacy code just couldn't deal with situations like this... The updated syntax now places the index were it belongs (ex: `*/QCPRMS[4]/GNAP`). 

_Note: Refactoring Part 2 will cover code from `QueryParser::collectData` onward._

2. Adds functionality which allows for the inclusion of a filter in the query string for cases when you want to select a subset of data in a particular repeat in the data (functionality dependent on refactoring). Here are some examples of what the new query strings can look like:

- `*/BRIT{2}/TMBR` Uses only the second repeat of BRIT and drops the extra dimension associated with the BRIT repeat (you get 1d not 2d dimensional data).
- `*/BRIT{1-3, 5}/TMBR` Keeps repeats 1,2,3 and 5 from BRIT. The data dimensionality does not change because you are filtering for more than 1 repeat.
- `*/ROSEQ1{3}/ROSEQ3{2-4}/PCCF` Pick the third instance of ROSEQ1 and drop that dimension, 2 thru 4 of ROSEQ3 retaining that dimension. 
- `*/QCPRMS[4]{2}/GNAP` The mnemonic for QCPRMS is repeated 4 times. Pick the BUFR data that corresponds to the 
4th instance of that mnemonic and only keep the 2nd repeat dropping that dimension.

_Note: The code that does the actual filtering can be found in `QueryParser::collectData`. Yes, it will get caught up in Refactoring Part 2._

### Issue(s) addressed

#1180 
#967 

Unit tests pass. Queries like `*/ABC{1-2}/XYZ` work.


rmclaren:
@CoryMartin-NOAA Would it be possible to grant Edward Hartnett access to this JCSDA internal? It might be time time make him aware of what we are doing...

CoryMartin-NOAA:
@rmclaren I don't have the power to do that, this is a public repo though (despite it being in internal). Does he need write access or will an "@" for an FYI suffice? (if the former, we have to go through Daryl)

rmclaren:
@CoryMartin-NOAA I don't think he needs write access, I just wanted him to take a look... I was not aware that this repo was publicly visible. Thanks:)

rmclaren:
Note to the reviewers: Looking at the diffs for this PR is not much use as in most cases things have been completely replaced. It would probably be easier just to look at the new code directly without the diffs.

src/bufr/BufrParser/Query/ResultSet.cpp
```diff
@@ -296,7 +283,7 @@ namespace bufr {
 
         // Convert dims per data frame to dims for all the collected data.
         dims[0] = totalRows;
-        if (dataFrames_.size() > 1)
+        if (dataFrames_.size() > 0)
```
nicholasesposito, line 286: Why the change to > 0 instead of > 1?

src/bufr/BufrParser/Query/SubsetTable.h
```diff
@@ -11,25 +11,271 @@
...
+        /// \return True if this node is a leaf node.
+        bool isLeaf() const
+        {
+            return type == Typ::Number ||
+                   type == Typ::Character;
+        }
+
+        /// \brief Does this type of node contain children?
+        /// \return True if this node is a container node.
+        bool isContainer() const
```
nicholasesposito, line 79: What is an example of when this would happen and/or be needed?

src/bufr/BufrParser/Query/Tokenizer.h
```diff
@@ -0,0 +1,386 @@
...
+        virtual ~Token() = default;
+
+        /// \brief Get the string representation of the token
+        std::string str() const  { return str_; }
+
+        /// \brief Get the debug string for the token.
+        virtual std::string debugStr() const = 0;
+
+        /// \brief Get the subtokens for this token.
+        virtual std::vector<std::shared_ptr<QueryToken>> queryTokens() const
```
nicholasesposito, line 43: So would a subtoken be when we're looking at something like `*/BRIT{2}/TMBR` where TMBR  is the subtoken?

nicholasesposito (commented):
Reviewing and I have some questions. I'll likely have more as I continue to dig. Looks good so far though.

BenjaminRuston:
this is a lot of files changed at once, any way to break it down further or the remaining parts?

doc/bufr_uml/BigPicture.puml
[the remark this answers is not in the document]
rmclaren: These are plantuml files. PlantUml is a library which lets you specify UML images via text files. They can be rendered using the plantuml executable. They are basically like source files. I'll leave it up to the group wether we want to check in the resulting images as well...

src/bufr/BufrParser/Query/ResultSet.cpp
rmclaren: I was accidentally not slicing if there is only 1 dataframe in the file.

rmclaren: By slicing I mean removing unwanted dimensional data from the dimensions array.

src/bufr/BufrParser/Query/SubsetTable.h
rmclaren: This function tells you if a node is a type of node that can contain children. It can be useful in a lot of ways, for example while recursively searching the BUFR subset tree you can use it to figure out if you need to dive deeper or not.

src/bufr/BufrParser/Query/SubsetTable.h
```diff
@@ -11,25 +11,271 @@
...
+            return getPathNodes(components);
+        }
+
+        /// \brief Recursively get the path to this node as a vector of BufrNodes.
+        /// \param components The path to this node as a vector of BufrNodes.
+        /// \return The path to this node as a vector of BufrNodes.
+        std::vector<std::shared_ptr<BufrNode>>
+            getPathNodes(std::vector<std::shared_ptr<BufrNode>>& components)
+        {
+            if (!parent.expired())
```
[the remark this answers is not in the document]
rmclaren: Parent is a weak_ptr. This call checks to make sure the weak_ptr is set to a valid pointer. In practice they will be expired if you are at the root of the tree.

doc/bufr_uml/BigPicture.puml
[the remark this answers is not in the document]
emilyhcliu: @rmclaren Same question as @nicholasesposito It would be great if we could visualize the plantuml files.  
The diagrams may help us understand the new code.

emilyhcliu:
**Big Picture**

VPD1RnCn48Nlyoj6FQM1L9MuLT6gWhOI1rGG5VV9UfAQUcrhx4WKjFyTkydMim7raCE-xsNcUSQ-JfaaxmS7ZzOmyX3pCmwS7uD9a9wiZoGqe8aMFrnU0eJDJ-upVY7h0Q1tb1BUKARx5ttzqC4VG8HHiq_1v-9dcPWFcT7nDb_XlUp8sz-Sq1GdMfy3zfHv5yHo0e

**Big Picture Sequence**

TP7FJiCm3CRlVWeVuL2f_3cXgXeSa3XmGDiBXDOIqTAqnAvWbBqx6XdEieYZVz_lYsHdnMA2p8C3tbe_cM06x2UBDtKDt1b7U5LVt-A7xUKDe7E66JVZbhpzeW0B8EB2OiHs-yWPvHXVlCyKTfUmptgmd847mzHisbF-SYWMzlFiv9_4qp1wswKWyYVVqsTXhscYO6GCHNBlT-lvbKcA

**Data Collection**

fLLDRzim3BtxL_1Kk9i3T1-d8JFQhGZG0IlQhjkrK2mc5IfBhaWtoORzzqcM4zcjTzbEYTvxV9H8IZub5exRMcVqe4mZdAX1DWhUdPnaL0cDyERaxNlOACdtMTQ8wa6i4QvRT3kOpnTMItHb2Tcl3A3IWWWMIgDV103WDMwnQXdpoXhY6FSD6OuU-ty5aFg9Tmm6jtmC7-06gTNicHZl

**Query Class Diagram**

dLHHRziW67mslmAbKfLEdLPjRtkOccIAD6dQkZRx0GG-9oW4N0nT-k_tWHt0TH9DUsdm-Uw-uy2zRnsppkyLsKe1j3R6hSs1j3kf6sRPdef6qaztTmNo1CyUD8SPTyREZHLWLrl9TnhQbdwyJ5mOPIozggM5jKLDmUvOZEysrdWjUgQIcwrx1QNCduBxW9abbmtJxZyKRoDVeeq24nPo

**Get Result Set** 

dL9DJ-Cm4BtxLuoqb_QmgsgvvO3w0K98L2eg5uwkFKcifdQOZ87-UvmqQcmQUk2McVSnJo-UDoo9_Q4IfT48kNEySv-YsHjRIv87qBM1czaiscpnpQDLk53iQEb88xsKHkqjDWtylqvSkSeH_CaDuOw29cAVM5Bj2tBUwfvPcQBa8rQL-uYu6qbibAcbvLyehaIEQUE1XMnpmbeQA-g1

**Tokenizer**

bLDDQzj043rxVyL08CJWG6XlFPJ43OPUsf2KtiVQaRtrQaUTdOtjVz-LFoAvKWGvIJozZxSFxNrK54sLDsjd2KfcNVBEn8qBDGfMO6i7d-_kEiWB_KqK2deeb6NEOab-hLsn2HGZV1edVcFF0bUb4rfArdJOHnOMcvLm2lR4z6wrrZrvpziEzmb5NU5g3FfnHQ_o2ArVs63J4-Qfb4oE

emilyhcliu:
@rmclaren Are you addressing [Issue #967](https://github.com/JCSDA-internal/ioda-converters/issues/967) in this PR?
If I understood correctly, this PR does not address Issue #967.

BenjaminRuston:
@CoryMartin-NOAA to add Edward Hartnett to internal you need to have Daryl make a change request

PatNichols:
@rmclaren Don't worry about the failing tests. It's on our end.

srherbener:
> @rmclaren sorry about this but I won't be able to review this week because of my priority being the upcoming skylab release. I'm targeting to get the review done by the middle of next week. If that is too much of a delay, please feel free to assign another reviewer.

I was able to start looking at this but didn't make much progress yet. I'll continue next week. Sorry about the delay.

rmclaren:
@srherbener No problem! Please don't feel like you need to rush.

src/bufr/BufrParser/Query/Tokenizer.h
rmclaren: So this returns a list of QueryToken objects. So for example if you've got a MultiQuery like `[*/BRITCSTC/TMBR, */BRIT/TMBR]` then the query tokens are `*/BRITCSTC/TMBR` and `*/BRIT/TMBR`.

rmclaren:
I updated some of the plantUML files so that they are more up to data.

**Create a new QuerySet**
CreateQuerySet

**Collecting Data**
DataCollection

**Class Diagram**
QueryClassDiagram

rmclaren:
@BenjaminRuston Sorry for all the code changes. My ultimate goal for this was to get the filtering working. In order to get there it made sense to refactor a lot of code that I've been meaning to improve. Much like pulling a thread in a knitted sweater an improvement to one class or module will demand changes in the next one in the chain...

It would be a bit of work to try separate things out into smaller pull requests at this point, as I'd have to write compatibility layers (new to old code) at each step of the way. I'll try to break the next round of changes into smaller chunks, however.

Like I said previously, you can pretty much take the refactored code and treat it like new code... If you want to make it easier on yourself you could just focus one class at a time in the order they appear within the sequence diagrams. So for example you could start with "Create a new QuerySet" (above) and work through those classes one at a time... And then you could go through the Data Collection sequence and work through those classes.

PatNichols:
@rmclaren We finally got the CI testing problem fixed and your PR passesd. We used it for a test actually :)

rmclaren:
@PatNichols Cool! Thanks @PatNichols.

emilyhcliu:
I checked out the branch and ran a test with a few satellite radiances data (GMI, ATMS, ...etc). 
I do not see any problem.

src/bufr/BufrParser/Query/QueryRunner.cpp
```diff
@@ -52,251 +57,110 @@ namespace bufr {
...
-                         subQuery < queries.cend();
-                         ++subQuery)
-                    {
-                        if (subQuery != queries.cbegin()) oops::Log::warning() << ", ";
-                        oops::Log::warning() << subQuery->queryStr;
-                    }
-                    oops::Log::warning() << "]";
-                }
 
+                // Print message to inform the user of the missing targetz
```
CoryMartin-NOAA, line 101:
```suggestion
                // Print message to inform the user of the missing targets
```

src/bufr/BufrParser/Query/Target.h
```diff
@@ -10,23 +10,122 @@
...
+
+
+        /// \brief Check if this component adds a dimension to the data.
+        bool addsDimension() const
+        {
+            return (type == Type::Subset || type == Type::Repeat) &&
+                   (queryComponent->filter.empty() || queryComponent->filter.size() > 1);
+        }
+
+        /// \brief Sets the TargetComponent type based on the type of the BUFR query node TYP.
```
CoryMartin-NOAA, line 45: is `TYP` type? if so, why not just write out `type` in the comments?

src/bufr/BufrParser/Query/Tokenizer.cpp
```diff
@@ -0,0 +1,39 @@
+/*
+ * (C) Copyright 2022 NOAA/NWS/NCEP/EMC
```
CoryMartin-NOAA, line 2: 2023

src/bufr/BufrParser/Query/Tokenizer.h
```diff
@@ -0,0 +1,386 @@
+/*
+ * (C) Copyright 2022 NOAA/NWS/NCEP/EMC
```
CoryMartin-NOAA, line 2: 2023

test/testinput/bufr_ncep_adpsfc.yaml
```diff
@@ -12,7 +12,7 @@ observations:
 
       exports:
         variables:
-          # MetaData
+#          # MetaData
```
CoryMartin-NOAA, line 15: why the double comment?

test/testinput/bufr_query_filtering.yaml
```diff
@@ -0,0 +1,73 @@
+# (C) Copyright 2020 NOAA/NWS/NCEP/EMC
```
CoryMartin-NOAA, line 1: 2023

CoryMartin-NOAA (commented):
Code looks good, have a few minor squabbles on some comments and one question, but otherwise I think I am ready to approve once these superficial fixes are made

doc/bufr_uml/BigPicture.puml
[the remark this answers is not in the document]
rmclaren: @emilyhcliu Checked in the PNG files that go along with the plantUML files.

srherbener (changes requested):
Thanks for the refactoring and adding the ctests! I think in general the code looks good but I'm seeing test failures on my Mac. Here's a listing from the `test_iodaconv_bufr_ncep_atms` test:

```
26242 665/850 Testing: test_iodaconv_bufr_ncep_atms
26243 665/850 Test: test_iodaconv_bufr_ncep_atms
26244 Command: "/usr/local/bin/bash" "/Users/steveherbener/projects/CONVERTERS/ioda-bundle/build/bin/iodaconv_comp.sh" "netcdf" "/Users/steveherbener/projects/CONVERTERS/ioda-bundle/build/bin/bufr2ioda.x testinput/bufr_ncep_atms.yaml" "gdas.t00z.atms_n20.tm00.nc" "0.0"
26245 Directory: /Users/steveherbener/projects/CONVERTERS/ioda-bundle/build/iodaconv/test
26246 "test_iodaconv_bufr_ncep_atms" start time: Apr 14 10:58 MDT
26247 Output:
26248 ----------------------------------------------------------
26249 PRE-MAIN-INFO BufrParser: Parsing file ./testinput/gdas.t00z.atms.tm00.bufr_d
26250 libc++abi: terminating due to uncaught exception of type std::__1::regex_error: One of *?+{ was not preceded by a valid regular expression.
26251 /Users/steveherbener/projects/CONVERTERS/ioda-bundle/build/bin/iodaconv_comp.sh: line 20:  2144 Abort trap: 6           $cmd
26252 <end of output>
26253 Test time =   0.34 sec
26254 ----------------------------------------------------------
26255 Test Failed.
26256 "test_iodaconv_bufr_ncep_atms" end time: Apr 14 10:58 MDT
26257 "test_iodaconv_bufr_ncep_atms" time elapsed: 00:00:00
26258 ----------------------------------------------------------
```

I'm seeing these warnings from the apple clang compilers (not sure if they are relevant):
```
/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/ResultSet.cpp:266:17: warning: logical not is only applied to the left hand side of this comparison [-Wlogical-not-parentheses]
            if (!targetField.data.size() == 0) {
                ^                        ~~
/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/ResultSet.cpp:266:17: note: add parentheses after the '!' to evaluate the comparison first
            if (!targetField.data.size() == 0) {
                ^
                 (                           )
/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/ResultSet.cpp:266:17: note: add parentheses around left hand side expression to silence this warning
            if (!targetField.data.size() == 0) {
                ^
                (                       )



/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/SubsetTable.h:22:5: warning: class 'PathComponent' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
    class PathComponent;
    ^
/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/QueryParser.h:72:12: note: previous use is here
    struct PathComponent : public QueryComponent
           ^
/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/SubsetTable.h:22:5: note: did you mean struct here?
    class PathComponent;
    ^~~~~
    struct



/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/../QuerySet.h:20:5: warning: class 'Query' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
    class Query;
    ^
/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/QueryParser.h:119:12: note: previous use is here
    struct Query
           ^
/Users/steveherbener/projects/CONVERTERS/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/../QuerySet.h:20:5: note: did you mean struct here?
    class Query;
    ^~~~~
    struct
```
I'm running on an M1 Mac using the Rosetta intel (x86_64) emulator. I have Ventura 13.3 installed with command line tools version 14.1 rc2 which has Apple Clang version 14.0.0. I'm also using GNU Fortran 12.2.0. I'm using version 1.3.0 of spack-stack.

srherbener:
Everyone I am on leave for the next two weeks (return on Mon, May 1st). I am fine approving once it is working on the Mac. Please feel free to dismiss my "request changes" once the tests pass on the Mac.

emilyhcliu (approved):

rmclaren:
Is there some trick to building oops with the llvm clang compiler, I have all kinds of problems... For example I keep getting the following when building the oops tests `ld: file not found: @rpath/libquadmath.0.dylib for architecture x86_64` even when I've confirmed, via otools, that the `liboops.dylib` file rpath has a valid path for libquadmath.

When I comment out the tests in the ./src/CMakeLists.txt file I then get a little further but end up with this:

```
[ 48%] Linking CXX shared library ../../lib/libqg.dylib
ld: warning: ld: warning: could not create compact unwind for ___qg_fields_mod_MOD_qg_fields_read_file: stack subq instruction is too different from dwarf stack size
could not create compact unwind for ___qg_error_covariance_mod_MOD_qg_error_covariance_setup: stack subq instruction is too different from dwarf stack size
ld: warning: could not create compact unwind for ___qg_fields_mod_MOD_qg_fields_write_file: stack subq instruction is too different from dwarf stack size
ld: warning: could not create compact unwind for ___qg_obsdb_mod_MOD_qg_obsdb_setup: stack subq instruction is too different from dwarf stack size
```

Its using the gcc version 12 version of gfortran. Weird thing all this stuff works correctly when I just use GCC/G++ tools. So I'm a bit stuck...

climbfuji:
> Is there some trick to building oops with the llvm clang compiler, I have all kinds of problems... For example I keep getting the following when building the oops tests `ld: file not found: @rpath/libquadmath.0.dylib for architecture x86_64` even when I've confirmed, via otools, that the `liboops.dylib` file rpath has a valid path for libquadmath.
> 
> When I comment out the tests in the ./src/CMakeLists.txt file I then get a little further but end up with this:
> 
> ```
> [ 48%] Linking CXX shared library ../../lib/libqg.dylib
> ld: warning: ld: warning: could not create compact unwind for ___qg_fields_mod_MOD_qg_fields_read_file: stack subq instruction is too different from dwarf stack size
> could not create compact unwind for ___qg_error_covariance_mod_MOD_qg_error_covariance_setup: stack subq instruction is too different from dwarf stack size
> ld: warning: could not create compact unwind for ___qg_fields_mod_MOD_qg_fields_write_file: stack subq instruction is too different from dwarf stack size
> ld: warning: could not create compact unwind for ___qg_obsdb_mod_MOD_qg_obsdb_setup: stack subq instruction is too different from dwarf stack size
> ```
> 
> Its using the gcc version 12 version of gfortran. Weird thing all this stuff works correctly when I just use GCC/G++ tools instead of LLVM. So I'm a bit stuck...

Not sure if this helps, but have a look at https://github.com/JCSDA-internal/jedi-bundle/blob/87e5c94134e02a46a3096a6fcafb8707aff2cbe0/CMakeLists.txt#L24 and https://github.com/JCSDA-internal/jedi-bundle/blob/87e5c94134e02a46a3096a6fcafb8707aff2cbe0/CMakeLists.txt#L28 please.

rmclaren:
@climbfuji Did have a look, I'm not using JEDI-bundle (don't need most of the stuff in there) but I did try some of those flags jsn't to see if it makes a difference (doesn't unfortunately). I think the problem is a more fundamental. It has the correct path to the resource in the rpath but fails to recognize it as a valid DYLIB. The library is part of the GCC toolchain (libquadmath.dylib) and I'm sure was compiled with gcc. The problem is that (llvm) clang doesn't like it for some reason... I've compiled the entire toolchain with LLVM on a Ubuntu virtual machine and was able to compile everything with clang there. The problem is that it does not have the problem that Steve witnessed (all the tests pass). So I don't know which regular expression is causing the issue in the Tokenizer...

BenjaminRuston (approved):
`develop` is passing all ctests for me on my Mac M1 with clang-1400.0.29.202 and  latest stack spack-stack-1.3.0

BenjaminRuston (commented):
shooooot,, strike that had accidentally tested `develop` and not `feature/query_filters` am now seeing the same errors as @srherbener

emilyhcliu:
> this is passing all ctests for me on my Mac M1 with clang-1400.0.29.202 and latest stack spack-stack-1.3.0

@rmclaren   Are you seeing the same failure on your Mac?

CoryMartin-NOAA:
@emilyhcliu Ron has an Intel Mac I think, the problems shown here are for Apple Silicon Macs, so the issue may be ARM vs x86_64 architecture related

src/bufr/BufrParser/Query/Tokenizer.h
```diff
@@ -0,0 +1,386 @@
...
+        constexpr static const char* Pattern = "\\{\\d+(\\-\\d+)?(\\,\\d+(\\-\\d+)?)*\\}";
+        constexpr static const char* DebugStr = "<filter>";
+
+        std::vector<size_t> indices() { return indices_; }
+
+        /// \brief Tokenize the str_ into tokens.
+        void tokenize() final
+        {
+            static const auto filterRegex =
+                std::regex("\\{(\\d+(\\-\\d+)?)(,\\d+(\\-\\d+)?)*\\}");
```
BenjaminRuston, line 147: this line does not have  `\\` in the second set of parenthesis different from the Pattern constexpr line above
```suggestion
                std::regex("\\{(\\d+(\\-\\d+)?)(\\,\\d+(\\-\\d+)?)*\\}");
```

src/bufr/BufrParser/Query/Tokenizer.h
```diff
@@ -0,0 +1,386 @@
...
+     private:
+        std::vector<size_t> indices_;
+    };
+
+    /// \brief Token for a query (i.e. '*/BRIT{2-4}/PCCF[2]'). This token contains tokens for all
+    ///        the elements of the query string.
+    class QueryToken : public TokenBase<QueryToken> {
+     public:
+        constexpr static const char* Pattern =
+            "([A-Z0-9_\\*\\/]+(\\[\\d+\\])?+(\\{[0-9\\-\\,]+\\})?)+";
```
BenjaminRuston, line 215: this one too, I have a regex check that doesn't like the forward slash in from of the right brack

```suggestion
            "([A-Z0-9_\\*\\]+(\\[\\d+\\])?+(\\{[0-9\\-\\,]+\\})?)+";
```

when I change this it does change the error but still failing:

97: libc++abi: terminating with uncaught exception of type std::__1::regex_error: The parser did not consume the entire regular expression.

src/bufr/BufrParser/Query/Tokenizer.h
rmclaren: It's actually more correct to remove the `\\` in front of the comma as commas don't need to be escaped (are not treated as special characters in regex).

src/bufr/BufrParser/Query/Tokenizer.h
rmclaren: I'd be careful with that regex checker... What tool is it? What variant of regex parser is it checking against? There are unfortunately a lot of regex parsers with different quirks( for example the python regex parser might have different quirks than the PHP regex parser, etc). That forward slash there is pretty damn important :). No way any of this is going to work without that. Also you just escaped the `]` character...

I'll try to sanatize my regex strings a little (make them compatible with more regex parsers). I use tools like https://regex101.com to play around with these. If you try this don't forget to remove the extra `\` characters so for example the string above would become `([A-Z0-9_\*\/]+((\[\d+\])?)+(\{[0-9\-\,]+\})?)+`. You have to escape the escape character in the c++ strings ;).

@BenjaminRuston Are you able to reproduce @srherbener exactly (same error message)?

rmclaren: If anyone can reproduce @srherbener's regex problem please try doing a `git pull` and give my latest version a try. I went through and a made my regex strings play better with more types of regex interpreters. I think I might have found one thing that might have caused his error.

BenjaminRuston (approved):
awesome this now passes all ctest on the M1 mac.

thanks @rmclaren looked at your sanitization and very happy this did the trick!

src/bufr/BufrParser/Query/Tokenizer.h
rmclaren: @BenjaminRuston Thanks Ben! I spent way to much time trying to recreate the issue on my mahcine... ugg. Thanks for your help :)

Issue #1194, opened 2023-03-24
BUFR Query: Python Interface for executing queries.


rmclaren:
Its time to build a python interface for the BUFR query code so that you can define and execute queries from a python app. The results should come out as numpy arrays.

PR #1195, opened 2023-03-24
BUFR Query Python Bindings

rmclaren:
## Description

Adds python interface so that you can define and execute queries from python code. Here is some sample python code that uses this new interface:

```python
import bufr

# Make the QuerySet for all the data we want
q = bufr.QuerySet()
q.add('latitude', '*/CLON')
q.add('longitude', '*/CLAT')
q.add('radiance', '*/BRIT/TMBR')

# Open the BUFR file and execute the QuerySet
with bufr.File( './testinput/gdas.t00z.1bhrs4.tm00.bufr_d') as f:
    r = f.execute(q)

# Use the ResultSet returned to get correctly dimensioned numpy arrays of the data
lat = r.get('latitude')
lon = r.get('longitude')
rad = r.get('radiance')
```

### Issue(s) addressed
#1194 

You can execute queries from inside a python program. No need for YAML files or writing C++ code...

## Dependencies

Waiting on the following PRs:
- [x] waiting on #1192


CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set( PYIODACONV_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/p
 ## Dependencies
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
+find_package( pybind11 REQUIRED)
```
CoryMartin-NOAA, line 35: how much effort is it to keep this optional and only build this if it is found? Not sure about everyone's thoughts on adding a dependency. This isn't 'required' (but optional) for IODA I don't think...

rmclaren: Ok, I made pybind11 optional. I followed the existing patterns. The BUFR python component will only be built if the bufr component is built and pybind11 is found.

src/bufr/BufrParser/BufrDescription.cpp
```diff
@@ -32,15 +31,6 @@ namespace Ingester
     {
         setFilepath(conf.getString(ConfKeys::Filename));
 
-        if (conf.has(ConfKeys::IsWmoFormat) && conf.getBool(ConfKeys::IsWmoFormat))
```
CoryMartin-NOAA, line 35: Just for my curiousity, why are the WMO format changes included in this PR? Is it required for this API?

test/testinput/bufr_query_filtering.yaml
```diff
@@ -0,0 +1,73 @@
+# (C) Copyright 2020 NOAA/NWS/NCEP/EMC
```
CoryMartin-NOAA, line 1: 2023

emilyhcliu:
@rmclaren I ran the ctest for all bufr converter related tests.
There are total 37 test.  All passed but one.  The test_iodaconv_bufr_query_python_test failed.
The error message is the following:
```
674:   File "/scratch1/NCEPDEV/da/Emily.Liu/JEDI-iodaconv/ioda-bundle-query_filters/build/iodaconv/test/testinput/bufr_query_python_test.py", line 7, in <module>
674:     import bufr
674: ModuleNotFoundError: No module named 'bufr'
1/1 Test #674: test_iodaconv_bufr_query_python_test ...***Failed    0.04 sec
```

rmclaren:
@emilyhcliu  Sorry about that, I changed the way its setting the PYTHONPATH variable a little. Wondering if you could update and try it again?

emilyhcliu:
> @emilyhcliu Sorry about that, I changed the way its setting the PYTHONPATH variable a little. Wondering if you could update and try it again?

@rmclaren Ok. I see, the library path was wrong.  I updated the branch and ran again.  The test found the bufr but failed to find the numpy.  Here is the error mesasge:
```
674:  PYTHONPATH=/scratch1/NCEPDEV/da/Emily.Liu/JEDI-iodaconv/ioda-bundle-query_filters/build/lib/pyiodaconv
674:  OMP_NUM_THREADS=1
674: Test timeout computed to be: 1500
674: Traceback (most recent call last):
674:   File "/scratch1/NCEPDEV/da/Emily.Liu/JEDI-iodaconv/ioda-bundle-query_filters/build/iodaconv/test/testinput/bufr_query_python_test.py", line 8, in <module>
674:     import numpy as np
674: ModuleNotFoundError: No module named 'numpy'
1/1 Test #674: test_iodaconv_bufr_query_python_test ...***Failed    0.06 sec

I checked my module list.  Python version 3.9.12 is loaded.  

```

rmclaren:
@emilyhcliu So I revised the way that the CMakeLists.txt file in the test directory builds up its python path, with the hopes that it might resolve your issue with the test script... Could you try?

rmclaren:
I modified the code to return proper numpy.ma.array objects (masked array) to better deal with the "missing values".

test/testinput/bufr_query_python_to_ioda_test.py
emilyhcliu, line 1: @rmclaren Thanks for adding this Python to IODA test!!  
In the test file, more than 50% of the code is for preparing output to IODA.  
Based on what I saw in this test and the [Issue](https://github.com/JCSDA-internal/ioda/issues/945) @rmclaren  posted, there is indeed necessary to enhance the Python interface in IODA Engines.

emilyhcliu (approved):

src/bufr/BufrParser/Query/DataProvider/DataProvider.cpp
```diff
@@ -58,6 +63,24 @@ namespace bufr {
         }
 
         deleteData();
+
+        if (!foundBufrMsg)
+        {
+            std::ostringstream errStr;
+            errStr << "No BUFR messages were found! ";
+            errStr << "Please make sure that " << filePath_ << " exists and is a valid BUFR file.";
+            throw eckit::BadValue(errStr.str());
```
CoryMartin-NOAA, line 72: Do these need wrapped in an IFDEF like `#ifdef BUILD_IODA_BINDING`? Or is eckit always assumed to be here when this file is compiled?

src/bufr/CMakeLists.txt
```diff
@@ -3,118 +3,180 @@
...
+		IodaEncoder/IodaDescription.h
+	)
+	#>>Ron
+	list (APPEND _atmslib_srcs
+					BufrParser/Exports/Variables/Transforms/atms/kinds.F90
+					BufrParser/Exports/Variables/Transforms/atms/atms_spatial_average_mod.f90
+					BufrParser/Exports/Variables/Transforms/atms/atms_spatial_average_interface.h
+					BufrParser/Exports/Variables/Transforms/atms/atms_spatial_average_interface.f90
+	)
+	ecbuild_add_library( TARGET   atms_lib
```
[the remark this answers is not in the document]
CoryMartin-NOAA: Formatting of this CMakeLists.txt still looks odd to me, can you check it in a different editor?

test/CMakeLists.txt
```diff
@@ -429,7 +432,15 @@ set(IODA_CONV_COMP_TOL "0.5e-4")
 #        )
 #  Dom's fix to get right path
 #set(IODACONV_PYTHONPATH "$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
-set(IODACONV_PYTHONPATH "${CMAKE_CURRENT_SOURCE_DIR}/../src:$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
+
+list(APPEND python_path
+  "${CMAKE_CURRENT_SOURCE_DIR}/../src"
+  "${CMAKE_BINARY_DIR}/lib/pyiodaconv"
```
CoryMartin-NOAA, line 438: not sure this is the right directory now that `lib_python` install stuff is in ioda-converters...

CoryMartin-NOAA (commented):
A few minor comments/questions, otherwise looks good

src/bufr/BufrParser/Query/DataProvider/DataProvider.cpp
CoryMartin-NOAA: @rmclaren just double checking, does this need an ifdef or will eckit always be available?

CoryMartin-NOAA (approved):

BenjaminRuston (commented):
playing my best @srherbener role , there's an compile error when working on my Mac:

```[ 96%] Building CXX object iodaconv/src/bufr/CMakeFiles/bufr.dir/DataObject.cpp.o
In file included from /Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/DataObject.cpp:8:
In file included from /Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/DataObject.h:37:
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/QueryParser.h:41:16: error: 'auto' return without trailing return type; deduced return types are a C++14 extension
        static auto parse(std::vector<std::shared_ptr<Token>> tokens)
               ^
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/QueryParser.h:61:20: error: no viable conversion from returned value of type 'std::shared_ptr<Ingester::bufr::SubsetComponent>' to function return type 'int'
            return component;
                   ^~~~~~~~~
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/QueryParser.h:74:16: error: 'auto' return without trailing return type; deduced return types are a C++14 extension
        static auto parse(const std::vector<std::shared_ptr<Token>> tokens)
               ^
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/QueryParser.h:106:20: error: no viable conversion from returned value of type 'std::shared_ptr<Ingester::bufr::PathComponent>' to function return type 'int'
            return component;
                   ^~~~~~~~~
4 errors generated.
make[2]: *** [iodaconv/src/bufr/CMakeFiles/bufr.dir/DataObject.cpp.o] Error 1
make[1]: *** [iodaconv/src/bufr/CMakeFiles/bufr.dir/all] Error 2
```

BenjaminRuston:
thanks @rmclaren ,, getting a different error now:

```[ 96%] Building CXX object iodaconv/src/bufr/CMakeFiles/bufr.dir/BufrParser/Query/DataProvider/DataProvider.cpp.o
In file included from /Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/DataProvider.cpp:8:
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/DataProvider.h:163:57: error: no matching constructor for initialization of 'Ingester::bufr::SubsetVariant'
        SubsetVariant getSubsetVariant() const { return {subset_, variantId(), hasVariants()}; }
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/SubsetVariant.h:16:12: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 3 were provided
    struct SubsetVariant
           ^
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/SubsetVariant.h:16:12: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 3 were provided
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/SubsetVariant.h:16:12: note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 3 were provided
1 error generated.
```

BenjaminRuston:
@rmclaren have been able to pass all ctest using develop on my Mac, assume Mac use LLVM but happy to tag up on this to see if we can't get over the hump on this one

srherbener:
I'm seeing the same thing on my Mac as reported by @BenjaminRuston. 

```
/Users/steveherbener/projects/DEVELOP/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/DataProvider.h:163:57: error: no matching constructor for initialization of 'Ingester::bufr::SubsetVariant'
        SubsetVariant getSubsetVariant() const { return {subset_, variantId(), hasVariants()}; }
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/steveherbener/projects/DEVELOP/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/SubsetVariant.h:16:12: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 3 were provided
    struct SubsetVariant
           ^
/Users/steveherbener/projects/DEVELOP/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/SubsetVariant.h:16:12: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 3 were provided
/Users/steveherbener/projects/DEVELOP/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/SubsetVariant.h:16:12: note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 3 were provided
1 error generated.
```

Apple clang seems to not like the initializer list syntax in DataProvider.h line 163, yet the develop version of DataProvider.h is the same and it works. Not sure what has changed?

PatNichols:
It's an error in the c++ standard being used plus how an aggregation initiializer should be used.. This type of initialization is a c++11 standards thing which Xcode does not honor unless you pass -std=c++11. We need to add that to the CMakeLists.txt at some point.  The fix is proably something like:

SubsetVariant getSubsetVariant() const { SubsetVariant sv =  {subset_, variantId(), hasVariants()}; return sv;}

The standards are in:
https://en.cppreference.com/w/cpp/language/aggregate_initialization

rmclaren:
@srherbener Thanks for the insight Steve. I spent most of my day rebuilding a build environment in a virtual machine using MacPorts instead of Homebrew and was finally able to recreate the problem on my side... I'll never understand how Homebrew got to be so popular... MacPorts is a lot more capable... anyways.

I added a proper constructor to that Struct and just use it which fixes Benjamins problem.

srherbener:
> @rmclaren @BenjaminRuston @srherbener
> 
> It's an error in the c++ standard being used plus how an aggregation initiializer should be used.. This type of initialization is a c++11 standards thing which Xcode does not honor unless you pass -std=c++11. We need to add that to the CMakeLists.txt at some point. The fix is proably something like:
> 
> SubsetVariant getSubsetVariant() const { SubsetVariant sv = {subset_, variantId(), hasVariants()}; return sv;}
> 
> The standards are in: https://en.cppreference.com/w/cpp/language/aggregate_initialization

Thanks @PatNichols for the information. Just want to note that we should already be using `-std=c++14` in our builds.

PatNichols:
> > @rmclaren @BenjaminRuston @srherbener
> > It's an error in the c++ standard being used plus how an aggregation initiializer should be used.. This type of initialization is a c++11 standards thing which Xcode does not honor unless you pass -std=c++11. We need to add that to the CMakeLists.txt at some point. The fix is proably something like:
> > SubsetVariant getSubsetVariant() const { SubsetVariant sv = {subset_, variantId(), hasVariants()}; return sv;}
> > The standards are in: https://en.cppreference.com/w/cpp/language/aggregate_initialization
> 
> Thanks @PatNichols for the information. Just want to note that we should already be using `-std=c++14` in our builds.

Two ways to do that. 1) modify the cmake or modify the cmake compiler flags. I don't know if it's an issue but some partners still use cray compilers. This is perhaps the problem.

PatNichols:
> @srherbener Thanks for the insight Steve. I spent most of my day rebuilding a build environment in a virtual machine using MacPorts instead of Homebrew and was finally able to recreate the problem on my side... I'll never understand how Homebrew got to be so popular... MacPorts is a lot more capable... anyways.
> 
> I added a proper constructor to that Struct and just use it which fixes Benjamins problem.

@rmclaren Thank you for fixing this. I feel your pain.

BenjaminRuston:
sorry for the bad news but still not compiling for me:

`[ 96%] Building CXX object iodaconv/src/bufr/CMakeFiles/bufr.dir/BufrParser/Query/File.cpp.o
In file included from /Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/File.cpp:8:
In file included from /Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/File.h:10:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/string:520:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__functional_base:15:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__functional/invoke.h:14:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__functional/weak_result_type.h:16:
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/type_traits:1838:38: error: implicit instantiation of undefined template 'std::hash<Ingester::bufr::Typ>'
    : public integral_constant<bool, __is_empty(_Tp)> {};
                                     ^
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/unordered_map:454:18: note: in instantiation of template class 'std::is_empty<std::hash<Ingester::bufr::Typ>>' requested here
          bool = is_empty<_Hash>::value && !__libcpp_is_final<_Hash>::value>
                 ^
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/unordered_map:937:13: note: in instantiation of default argument for '__unordered_map_hasher<Ingester::bufr::Typ, std::__hash_value_type<Ingester::bufr::Typ, Ingester::bufr::TargetComponent::Type>, std::hash<Ingester::bufr::Typ>, std::equal_to<Ingester::bufr::Typ>>' required here
    typedef __unordered_map_hasher<key_type, __value_type, hasher, key_equal> __hasher;
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/benr/work/JCSDA/JEDI/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/Target.h:49:50: note: in instantiation of template class 'std::unordered_map<Ingester::bufr::Typ, Ingester::bufr::TargetComponent::Type>' requested here
            static std::unordered_map<Typ, Type> typMap = {
                                                 ^
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__memory/shared_ptr.h:1710:50: note: template is declared here
template <class _Tp> struct _LIBCPP_TEMPLATE_VIS hash;
                                                 ^
1 error generated.
make[2]: *** [iodaconv/src/bufr/CMakeFiles/bufr.dir/BufrParser/Query/File.cpp.o] Error 1
`

rmclaren:
So I'm able to compile with the MacOSX13.3 SDK on my virtual machine (should be the latest available). What version of XCode do you have installed?

PatNichols:
> So I'm able to compile with the MacOSX13.3 SDK on my virtual machine (should be the latest available). What version of XCode do you have installed?

@rmclaren We are using Xcode 14.0.0. @srherbener and @BenjaminRuston are using the version for the M1 chip which might be different from the intel one (it's a different SDK).

srherbener:
I'm getting the same compiler error that @BenjaminRuston reported. I think I found the culprit. For the bufr code clang command has a `-std=gnu++11` spec appearing near the end that overrides the `-std=gnu++14` spec. We should be using C++14 everywhere (JEDI advertises C++14 compliance) and I suspect that the Apple command line tools use C++11 by default (which causes that override to appear).

I think the appropriate fix is to explicitly specify `CXX_STANDARD 14` in the ioda converters cmake configuration. That is what the ioda repo does, and I suspect the other repos as well. I'll give that a try and push the changes to github if it works.

rmclaren:
The SDK's are typically associated with the operating system (13.3 being the latest version of Ventura). I wouldn't expect the differences between the ARM64 and Intel versions of the SDK to be on the level of totally different C++ header files (mainly different hardware abstraction layer [HAL], ARM64 binaries, some extra libraries to support M1 specific features).

The thing is that Apple goofs quiet often. For example I ran into an issue very recently where the version of `ld` (linker) that shipped with XCode 14.0 would not properly link object files that were compiled with GNU compilers (see: https://stackoverflow.com/questions/73734125/xcode-14-0-gcc-12-linker-issue). Apple has fixed this issue, but there may have been others associated with that release.

I don't want to introduce a lot of code changes if there is an issue in that toolchain.

rmclaren:
Xcode 14.3 toolchain with the 13.3 version of the SDK worked for me.

srherbener:
I've imitated the C++14 settings from ioda in the bufr code cmake config and they work. The settings in ioda are a bit more complicated that I was expecting, but they do work. I'll make a PR based on the feature/query_python branch so my changes can be reviewed by all of you.

rmclaren:
@srherbener Thanks Steve!

Quick Question: Are we officially moving to the C++ 14 standard? I've been writing all my code to be complaint with the C++ 11 standard...

srherbener:
PR is submitted.

We officially moved to C++14 a ways back, I think over a year ago, so you should be okay using C++14 features.

cmake/compiler_flags_Clang_CXX.cmake
```diff
@@ -7,25 +7,25 @@
 # FLAGS COMMON TO ALL BUILD TYPES
 ####################################################################
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++14 -g -Wall -Wno-c++11-extensions")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -g -Wall -Wno-c++14-extensions")
```
BenjaminRuston, line 10: @climbfuji and @DJDavies2 was there a reason this was `gnu` a question asks

does this have an impact on anyone

srherbener (approved):
Builds successfully and all tests pass on my Mac! Thanks for everyone's patience and efforts to get this working.

BenjaminRuston (approved):
same @srherbener would like to hear from @climbfuji and someone from UKMO just to be sure as we changed the top level CMake

PR #1199, opened 2023-03-31
demo scripts for machines with rda access

### LINKED ISSUE #1198: ingest scripts callable by ewok
## Description
design some simple scripts which could be called by `ewok` to ingest observations to be processed and stored in `r2d2`

BenjaminRuston:
## Description

initial commit to demonstrate idea towards script which can utilize RDA as resource for observations

### Issue(s) addressed

- fixes #1198


BenjaminRuston:
this is the basic blueprint for how to get observations for period more than 48hours old using the RDA

this gives us access to numerous in-situ and satellite observations.  these too can likely be refine and maybe even more to `ewok` as they should be called by a meta scheduler to stage the data and run conversion

tools/ingest/rda/convert_rda_gdas.sh
```diff
@@ -0,0 +1,118 @@
...
+        # path to utility does not work
+        ${data_repos}/${data_type}/link_RDA.sh ${dtg}
+    fi
+    if [[ ! -s ${adir}/${afile} ]]; then
+        echo "  ... WARNING missing file: ${adir}/${afile}"
+        return 1
+    fi
+    make_temp_output_path
+    ### cmd="${ioda_bundle}/build/bin/bufr2nc_fortran.x -i ${adir} -o ${output_temp_dir}  ${afile}"
+    cmd="mpirun -n 1 ${ncar2ioda_bundle}/obs2ioda-v2.x -i ${adir} -o ${output_temp_dir}  ${afile}"
```
BenjaminRuston, line 71: actually need to v2 to v3 added too !!!

tools/ingest/rda/convert_rda_gdas.sh
```diff
@@ -0,0 +1,118 @@
...
+
+assim_freq=${assim_freq:-6}
+window_length=$( printf "%d" ${assim_freq} )
+
+workdir=${UWORKDIR:-/glade/p/univ/ucos0003/${USER}}
+data_type=${data_type:-amsua}
+data_gdas_type=${data_gdas_type:-1bamua}
+data_repos=${DATA_REPOS:-${workdir}/data_repos}
+ioda_bundle=${IODA_BUNDLE:-${WORKDIR}/work/JCSDA/JEDI/ioda-bundle}
+ncar2ioda_bundle=${NCAR2IODA_BUNDLE:-${WORKDIR}/work/JCSDA/JEDI/ncar2ioda/obs2ioda/obs2ioda-v2/src}
```
BenjaminRuston, line 30: note --  relying on an external bundle there is an old version in this repository

tools/ingest/rda/convert_rda_gdas.sh
```diff
@@ -0,0 +1,118 @@
...
+    for afile in ${wild_string}; do
+        platform=${afile#${data_type}_}
+        platform=${platform%_obs*}
+        echo "   .... ahhh ... platform ... ${platform}"
+        output_file="${output_dir}/${atime}_${data_type}_${platform}.nc4"
+        cmd="mpirun -np 1 ${ioda_bundle}/build/bin/ioda-upgrade.x ${afile} ${output_file}"
+        echo "${cmd}"
+        ${cmd}
+        # path to utility does not work
+        python3 ${UWORKDIR}/data_repos/fix_nchans.py --input ${output_file} --sensor ${data_type}
```
BenjaminRuston, line 52: this also only needs to be called for radiances `amsua`, `mhs`, `iasi`, `cris` but does not need to be called for `prepBUFR` for instance

BenjaminRuston:
as you can guess there are more of these however need to establish the real time and archival feeds for each ideally

for `ldm` we have the real time feed

then also for some sensors are prescription based and for various servers but essentially ones which require credential verification

BenjaminRuston:
@ashley314  if you're not using this going to remove the draft PR

ashley314:
@BenjaminRuston I think this will be helpful in the future. Is it possible to keep as a draft PR for now? I do not know when or if we are going to prioritize running on Cheyenne or Casper.

BenjaminRuston:
@ashley314 may go ahead and close this unless you think they are still useful

ashley314:
@BenjaminRuston  I will close for now and make a note that it might be useful when we get to airs and wind data! thanks!

PR #1206, opened 2023-04-10
adding GNSSRO adjustLocalSpectralWidth

hailingz:
## Description
This PR adds the local spectral width (LSW) of bending angle observations (UCAR processed data only).  

A few notes:
1. LSW is added as adjustLocalSpectralWidth metadata
   Specifically, we use a "top-down" method to trace LSW. Starting from the top of each profile, when a non-zero LSW is detected at a certain height, all the "adjustLocalSpectralWidth" below this height in the entire profile are assigned with the fractional LSW (%). That says "adjustLocalSpectralWidth" in the profile always decrease monotonically from the bottom. This will facilitate the QC procedure.

2. LSW are added only for UCAR processed data, i.e., process_center=60

3. earlier test data does not contain LSW, so was replaced with new test data. 
    The new test data contains 3 profiles. These selected profiles cover various scenarios and are expected to be used for UFO tests in ext step.
        prof1. low latitude, no Super refraction (SR), Descending
        prof2. low latitude, with Super refraction (SR), Descending
        prof3. high latitude, no Super refraction (SR), Ascending

4. "test_iodaconv_gnssro_conv" was removed along with the old test input "gnssro_kompsat5_20180415_00Z" and test output "gnssro_obs_2018041500.nc4" because (a) the test was a duplication of "test_iodaconv_gnssro_bufr_conv" as part of the NCAR fortran converter, and (2) the NCAR fortran converter is not maintained now.

ctest -R test_iodaconv_gnssro_bufr_conv
 
## Dependencies
N/A

## Impact
N/A

### DIFF
--- a/src/gnssro/gnssro_bufr2ioda.f90
+++ b/src/gnssro/gnssro_bufr2ioda.f90
@@ -315,7 +315,7 @@ program gnssro_bufr2ioda2
                gnssro_data%azim(ndata) = azim
                CALL bendingangle_err_gsi(rlat, impact - roc, obsErr, ogce)
                gnssro_data%bndoe_gsi(ndata) = obsErr
-               if (ref > r_missing) then
+               if (ref < r_missing) then
                   CALL refractivity_err_gsi(rlat, height, GlobalModel, obsErr)
                   gnssro_data%refoe_gsi(ndata) = obsErr
                else
--- a/src/gnssro/gnssro_bufr2ioda.f90
+++ b/src/gnssro/gnssro_bufr2ioda.f90
@@ -224,30 +224,31 @@ program gnssro_bufr2ioda2
          nrec = nrec + 1
          ndata0 = ndata
 
-     if (ogce .eq. 60) then
+         preqc = 0
+         if (ogce .eq. 60) then
 
-        do k = 1, levs
-           if (data1b(5,k) < 1.e+9_real64 ) then
-              if ( data1b(5,k) - roc > 25000 ) then
-                 startp = 1
-                 endp   = levs
-                 stride = 1 ! top down
-              else
-                 startp = levs
-                 endp   = 1
-                 stride = -1 !bottom up
+            do k = 1, levs
+               if (data1b(5,k) < 1.e+9_real64 ) then
+                  if ( data1b(5,k) - roc > 25000 ) then
+                     startp = 1
+                     endp   = levs
+                     stride = 1 ! top down
+                  else
+                     startp = levs
+                     endp   = 1
+                     stride = -1 !bottom up
+                  end if
+                  exit
               end if
-              exit
-          end if
-       end do
+            end do
 
-       do k = startp, endp, stride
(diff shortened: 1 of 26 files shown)


src/gnssro/gnssro_bufr2ioda.f90
```diff
@@ -223,6 +224,32 @@ program gnssro_bufr2ioda2
...
+            end do
+
+            do k = startp, endp, stride
+               if ( data1b(20,k) < 1.e+9_real64 .and. data1b(18,k) <1.e+9_real64 ) then
+                  preqc(k:endp:stride) = max(int(data1b(20,k)/data1b(18,k)*100), preqc(k))
+               end if
+            end do
+
+         end if
+
```
huishao-r, line 252: @hailingz so preqc is the variable containing normalized LSW value, Then we'll use the bound filter in a yaml to use this preQC, is it correct?

huishao-r: Thanks! Another question: Would it be necessary to use the variable name"preQC" in this code?  Should we just use normalizedLsw (or something similar) for a more descriptive variable name. Because really here this variable is just the LSW value, not the QC yet.

huishao-r (approved):
Thanks for the update. I would accept it now. Later, we can test the data set in our runs.

hailingz:
> Thanks for the update. I would accept it now. Later, we can test the data set in our runs.

I found the original bufr data used for the converter does not contain LSW. have to replace the bufr data. am working on it now.

huishao-r (approved):

fcvdb (approved):
The changes look good!

src/gnssro/gnssro_bufr2ioda.f90
```diff
@@ -431,6 +457,15 @@ program gnssro_bufr2ioda2
    call check(nf90_put_att(grpid_obserror, varid_bndoe, "valid_range", real((/0.0, 0.008/))))
    call check(nf90_def_var_fill(grpid_obserror, varid_bndoe, 0, real(r_missing)))
 
+   call check(nf90_def_var(grpid_metadata, "adjustLocalSpectralWidth", NF90_INT, nlocs_dimid, varid_adjlsw))
+   call check(nf90_def_var_deflate(grpid_metadata, varid_adjlsw,   &
+                                    & shuffle=1, deflate=1, deflate_level=deflate_level))
+   call check(nf90_put_att(grpid_metadata, varid_adjlsw, "units", "%"))
+   call check(nf90_put_att(grpid_metadata, varid_adjlsw, "longname",  &
+                                    & "adjust local spectral width in percentage of bending angle value" ))
+   call check(nf90_put_att(grpid_metadata, varid_adjlsw, "valid_range", "0 - 100" ))
```
fabiolrdiniz, line 466: This should be something like:
```
   call check(nf90_put_att(grpid_metadata, varid_adjlsw, "valid_range", real((/0.0, 100.0/)) ))
```

PR #1210, opened 2023-04-14
Bugfix/install raob

### LINKED ISSUE #1209: [Bug] Data Files not installed
## Description 
-------------------------

      The  ioda-converters package does not install the RAOB data files when one installs iodaconv package.  This requires any code that uses these to have keep a  build directory around. 

## Reproduce
-----------------------

     Set a cmake install directory and do make installl. Note the data files are not instalaled.

## Expected behavior
-----------------------------

      One should be able to delete the build directory after installing the package and there  will be no loss in functionality.

PatNichols:
## Description

The PR allows the raob data files to be installed when doing make install. The source directory can then be deleted without loss. of functionality.

### Issue(s) addressed

- fixes #1209 

One should be able to run iodaconv from an installation. 

## Dependencies

None


fcvdb (approved):
Thanks Pat!

Issue #1181, opened 2023-03-01
Bufr Query: Refactoring Part 2


rmclaren:
Refactor the following components:

NCEPLIB-bufr - New functions will be added to expose fixed repeat information, and information about the creation time for each BUFR message.

QueryRunner - QueryRunner is responsible for identifying targets and collecting data from the currently loaded BUFR message subset. Currently the DataCollector needs to walk the BUFR data in order to manually discover the number of repeats, especially for fixed repeats. The refactored version will use the new data available in NCEPLIb-bufr in order to read the number of repeats directly and so avoid the need for a lot of looping (hopefully speeding things up a bit) and greatly simplifying this code. Accommodations will need to be made for non-subset data (Meta Data) like the message creation time. Overall the code will become simpler, more efficient and an effort will be made to make it more object oriented.

ResultSet - This object is responsible for creating the result data sets at the end. A part of this is understanding how to apply the grou by fields properly. An effort will be made to make this code more object oriented, and to fix some issues with group by fields when they are used in conjunction with none repeated "Binary" sequences. I believe there may be a performance improvement we can do here as well.

BufrDescription - The YAML parser and associated code will be updated to be able to reference BUFR message meta data (specifically the creation time parameter we plan to expose). We will also need to invent a way to reference groups of input files from different directories (?) Optional/Alternatively: May implement a python interface to give the user the freedom to do whatever they want without the need to use a YAML file. This flexibility would make future maintenance easier as we wouldn't need to try to accommadate every conceivable use case in the YAML file format.

PR #1221, opened 2023-04-24
Refactored QueryRunner::collectData

rmclaren:
## Description

This pull request refactors the data collection process in QueryRunner::collectData. The user benefits from a small performance improvement in the data collection process, but the main benefit is much easier to maintain code.

The C interface on NCEPLIB-bufr was extended to expose a data structure that makes it possible to determine the number of repeats in a fixed repeated sequence. This means we don't need to walk the data structure any more to discover these values. Instead we can get the counts directly.

A new class called NodeLookupTable was introduced. So for each node we care about, given the targets we are looking for (these include the target nodes (ex: TMBR) plus the intermediate nodes (like BRIT) for those targets. Its initialized with a data provider and the targets, and it collects the relevant data from the currently open message subset. 

** NOTE: This pull request requires a new NCEPLIB-bufr tag to work **bufr_v12.0.0**.

### Issue(s) addressed

Is a part of #1181 

Unit tests run successfully.

## Dependencies

Has the following updated external dependency.
NCEPLIB-bufr  [bufr_v12.0.0]. https://github.com/NOAA-EMC/NCEPLIBS-bufr/projects/4#card-87633084


rmclaren:
No idea why the commit list is so long... Doesn't make sense.

BenjaminRuston:
@climbfuji adding you (or someone from INFRA) as well as INFRA label as there is a dependency on the NCEP BUFRlib which would need to be in the stack

climbfuji:
> @climbfuji adding you (or someone from INFRA) as well as INFRA label as there is a dependency on the NCEP BUFRlib which would need to be in the stack

Thanks for the heads up @BenjaminRuston ! Once bufrlib 12.0.0 is available, please have someone create an issue in spack-stack to update the version and we'll take it from there.

PatNichols:
@rmclaren @BenjaminRuston @climbfuji  After checking the recent releases for NCEP-bufrlib I found that 11.7.x was the latest release. Is 12.0 a development branch? I am somewhat confused on how we are to get this version.

rmclaren:
No idea why the tests are failing as I don't have access to the logs...

srherbener:
@rmclaren please try clicking on the CDASH links above. These are the lines in the check report above that have green check marks and titles like `AWS CodeBuild gnu - link to the corresponding CDASH page for viewing...`.

The CDASH links should take you to a web page (of which you should have access) that summarizes the test failures. For the GNU case, I see this message:
```
/jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/NcepDataProvider.cpp: In member function 'virtual void Ingester::bufr::NcepDataProvider::updateTableData(const string&)':
/jcsda/ioda-bundle/iodaconv/src/bufr/BufrParser/Query/DataProvider/NcepDataProvider.cpp:69:13: error: 'get_irf_f' was not declared in this scope; did you mean 'get_inv_f'?
   69 |             get_irf_f(&intPtr, &size);
      |             ^~~~~~~~~
      |             get_inv_f
```
If you can't access the CDASH web page, please let us know and we will get that resolved.

rmclaren:
That makes sense actually, its because they are using the older version of NCEPLIB-bufr (not v12 yet). 

I time out trying to access the CDASH pages unfortunately...

srherbener:
@rmclaren, thanks for the explanation.

I'll ask about CDASH. It should be public. Do you think it might be an issue with a firewall on your side?

rmclaren:
CDASH seems to be working now. Pretty sure its not a firewall on my side...

rmclaren:
@climbfuji This PR already moves to to the **bufr_4** version of the NCEPLIB-bufr 12 library. I don't know if there is a need for an issue for this...

emilyhcliu (approved):
Tested and approved!

PatNichols:
@rmclaren looks like we finally are able to test this!

src/ncar-bufr2nc-fortran/CMakeLists.txt
```diff
@@ -21,4 +21,4 @@ ecbuild_add_executable( TARGET  bufr2nc_fortran.x
                         SOURCES ${bufr2nc_source}
                         LIBS    bufr::bufr_4
                                 NetCDF::NetCDF_Fortran
-				iodaconv_utils)
+				                        iodaconv_utils)
```
PatNichols, line 24: @rmclaren could you delete the space here.

rmclaren: @PatNichols Done

srherbener (approved):
Looks good! I built and tested on my Mac M1 (arm64) and I'm seeing  tests fail, but they match test failures that I see using the develop branch. I'll submit an issue for the develop test failures on the Mac M1 (these test failures seem to occur only on the Mac M1).

If it's possible, please break up PRs into smaller sets of changes. It can be difficult and time consuming to review a large set of changes. I realize that this isn't possible sometimes, so no worries now and please keep in mind for future PRs. Thanks!

rmclaren:
@srherbener Thanks for looking at this.

PR #1233, opened 2023-05-03
Specify C++14 standard for C++ compiles

srherbener:
## Description

This PR contains cmake configuration changes that explicitly specify using the C++14 standard for C++ compiler calls. This is being done to fix issues with Mac builds where, in some cases, the C++11 standard was being used.

### Issue(s) addressed

None

The feature/query_python branch compiles successfully on the Mac.

## Dependencies

None

## Impact

None

### DIFF
--- a/src/bufr/CMakeLists.txt
+++ b/src/bufr/CMakeLists.txt
@@ -87,9 +87,6 @@ if ( iodaconv_bufr_ENABLED )
                        INSTALL_HEADERS LISTED
     )
 
-  set_property(TARGET atms_lib PROPERTY CXX_STANDARD 14)
-  target_compile_features(atms_lib PUBLIC cxx_std_14)
-
   list(APPEND _ingester_deps
               Eigen3::Eigen
               eckit
@@ -106,9 +103,6 @@ if ( iodaconv_bufr_ENABLED )
                        LINKER_LANGUAGE CXX
     )
 
-  set_property(TARGET ingester PROPERTY CXX_STANDARD 14)
-  target_compile_features(ingester PUBLIC cxx_std_14)
-
   target_link_libraries(ingester PUBLIC ${_ingester_deps})
 
   target_include_directories(ingester PUBLIC
@@ -123,8 +117,6 @@ if ( iodaconv_bufr_ENABLED )
   ecbuild_add_executable( TARGET  bufr2ioda.x
                           SOURCES bufr2ioda.cpp
                           LIBS    ingester )
-  set_property(TARGET bufr2ioda.x PROPERTY CXX_STANDARD 14)
-  target_compile_features(bufr2ioda.x PUBLIC cxx_std_14)
 
   ecbuild_add_test( TARGET  ${PROJECT_NAME}_bufr_coding_norms
                     TYPE    SCRIPT
@@ -183,8 +175,6 @@ if ( iodaconv_bufr_python_ENABLED )
                            ARCHIVE_OUTPUT_DIRECTORY "${PYIODACONV_BUILD_LIBDIR}"
                            LIBRARY_OUTPUT_DIRECTORY "${PYIODACONV_BUILD_LIBDIR}"
     )
-  set_property(TARGET bufr PROPERTY CXX_STANDARD 14)
-  target_compile_features(bufr PUBLIC cxx_std_14)
 
   install (TARGETS bufr DESTINATION ${PYIODACONV_INSTALL_LIBDIR})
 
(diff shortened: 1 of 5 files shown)


PatNichols (commented):
Can we just do 
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
int the top level CMakeLists.txt and be done with? I do remember that this fix depends on the version of cmake but I think we are way above that level.

PatNichols:
If we can set once for all c++ code we will not have to add it for each new c++ library added. The other way would be to add compiler flags in the cmake directory like we do for fortran.

rmclaren:
I agree with @PatNichols

BenjaminRuston:
@srherbener will have Pat make the change to your branch and test

srherbener:
> Can we just do set(CMAKE_CXX_STANDARD 14) set(CMAKE_CXX_STANDARD_REQUIRED ON) set(CMAKE_CXX_EXTENSIONS OFF) int the top level CMakeLists.txt and be done with? I do remember that this fix depends on the version of cmake but I think we are way above that level.

I agree that this is a better solution if it works. I was just emulating what Ryan did for the ioda python API. I'm not sure why Ryan chose to do it that way. Perhaps it's because the CMAKE_CXX_* commands just set default values, and Ryan's way is explicitly setting the CXX_STANDARD property. However, if CMAKE_CXX_* works, then I think that would be better.

PatNichols:
@srherbener I checked in my changes to the branch. The cmake compiler flags directory were wrong actually so there are a few additional minor fixes. Anytime someone compiles a c++ code it will do so wth --std=c++14 flag if it's available.

srherbener:
My mac build works with the latest changes. Do we need to remove the original changes I made?: [0162180](https://github.com/JCSDA-internal/ioda-converters/pull/1233/commits/01621807a77cc73ba7689f94bd1fe8a850df4d1a)

srherbener:
Also do we want gnu++14 or c++14? gnu++14 is used all over the place in JEDI and I'm guessing someone put that in intentionally.

rmclaren:
If those changes are no longer needed we should probably remove them... Not sure about gnu++ vs c++.

BenjaminRuston (approved):
woo-hoo this is working all ctests passing for me @rmclaren should we merge into the `feature/query_python` branch?

src/bufr/CMakeLists.txt
```diff
@@ -87,6 +87,9 @@ if ( iodaconv_bufr_ENABLED )
                        INSTALL_HEADERS LISTED
     )
 
+  set_property(TARGET atms_lib PROPERTY CXX_STANDARD 14)
```
BenjaminRuston, line 90: @PatNichols  this `14` can not be inherited from the higher level CMake?

rmclaren:
@BenjaminRuston I think @srherbener was going to the delete the changes out of src/bufr/CMakeLists.txt as they are not needed anymore. As soon as that happens I'll be happy to merge the changes.

@srherbener Would you like me to make the changes to save you some time?

srherbener:
> @BenjaminRuston I think @srherbener was going to the delete the changes out of src/bufr/CMakeLists.txt as they are not needed anymore. As soon as that happens I'll be happy to merge the changes.
> 
> @srherbener Would you like me to make the changes to save you some time?

Sorry about the confusion. I thought @PatNichols was going to do that, but I'm happy to do it now. I'll try it out and let you know what happens.

BenjaminRuston:
@srherbener sure go ahead ,, and yes saw your comments about the clean up and agree we should do that ... after merging this into Ron's branch may be best to add a couple more folks on this to look at it, thinking UKMO and Dom just to make sure all OK

PatNichols (approved):
It's good start for our repos

src/bufr/CMakeLists.txt
PatNichols: @BenjaminRuston No, this is the highest CMakeLists.txt in the repo. We could put these in the bundles but then it's not guaranteed it will present in general.

srherbener:
I just tried removing all of the original changes I made ([0162180](https://github.com/JCSDA-internal/ioda-converters/pull/1233/commits/01621807a77cc73ba7689f94bd1fe8a850df4d1a)) and the compile worked. It appears that the default (C++14) setting with the `CMAKE_CXX_STANDARD_*` commands is enough to propagate the C++14 setting throughout the repo.

Is it okay to push my changes, or is there an issue removing the explicit settings on the library targets?

rmclaren:
@srherbener Go ahead... Not all of them were removed. Sorry for all the confusion....

PR #1234, opened 2023-05-04
Copy surface_geopotential_height to surface_geometric_height.

gmao-jjin3:
Save surface heights twice as surface_geopotential_height and surface_geometric_height.


CoryMartin-NOAA (approved):
Works for me, I am unclear on which one is actually used in GSI... From my recollection of my previous investigation, I felt like they were interchangeable...

BenjaminRuston (approved):
thanks Jian, looks fine here @danholdaway move forward correct

Issue #1128, opened 2022-12-07
New dateTime construct does not work with group_by_variable in a YAML


PraveenKumar-NOAA:
## Current behavior (describe the bug)
I am trying to update the ADPUPA PrepBufr YAML after the new _dateTime_ construction method that has been added recently (#1090) to the ioda-converters. 

**YAML works with _bufr2ioda.x_ for the following two cases:** 

- Using a _group_by_variable_ field, but NO _dateTime_ 

<img width="380" alt="Screenshot 2022-12-07 at 9 53 36 AM" src="https://user-images.githubusercontent.com/63739712/206212074-ff9991fe-b65c-460a-b5e9-255d8b5215a7.png">

I used CAT mnemonic as a group_by_variable from the {PRSLEVEL}: 

<img width="557" alt="Screenshot 2022-12-07 at 10 15 30 AM" src="https://user-images.githubusercontent.com/63739712/206217587-dd34c35b-2f6e-480a-bed7-f4484bbfc0d5.png">

- NO _group_by_variable_ field, but _dateTime_ 

<img width="382" alt="Screenshot 2022-12-07 at 9 52 30 AM" src="https://user-images.githubusercontent.com/63739712/206211629-73f9f569-e6f6-4f7d-8cc0-059e2fb9e1ad.png">

**When I use both, _group_by_variable_ and _dateTime_, the YAML doesn't like it.**

## To Reproduce

- Computer used: Hera/Orion
- Compilers/modules used: 

<img width="796" alt="Screenshot 2022-12-07 at 10 03 54 AM" src="https://user-images.githubusercontent.com/63739712/206214421-2db62a45-cde0-4b49-8a92-7548bf6cb3df.png">

- Steps to reproduce the behavior:
1. YAML: /scratch1/NCEPDEV/da/Praveen.Kumar/JEDI/IODA_CONVR_ADPUPA_PB_2020110112/bufr_ncep_prepbufr_adpupa.yaml

2. Input file: /scratch1/NCEPDEV/da/Praveen.Kumar/JEDI/IODA_CONVR_ADPUPA_PB_2020110112/testinput/gdas.t12z.adpupa.prepbufr

3. Queries: /scratch1/NCEPDEV/da/Praveen.Kumar/JEDI/IODA_CONVR_ADPUPA_PB_2020110112/testinput/queries_adpupa_prepbufr

[...]

PR #1235, opened 2023-05-04
Fix for group_by bug in timeoffset variable.

rmclaren:
## Description
TimeoffsetVariable did not set group_by fields correctly in the Query Info objects it returned to identify the queries it needs. This PR fixes that.

### Issue(s) addressed
#1128 

New unit test test_iodaconv_bufr_ncep_prepbufr_adpupa passes.


emilyhcliu:
Tested the bug fix from this PR.  
It worked!!.
The conversion of time difference to dateTime variable works in conjunction with  "group_by_field"

BenjaminRuston (approved):
we should move this forward but will not the `bufr_ncep_prepbufr_adpupa.yaml` does not successfully decode a prepBUFR or adpupa file from the NCAR RDA archive

PR #1239, opened 2023-05-09
Bugfix/install libdir

### LINKED ISSUE #1187: Python converters still not installed consistently
## Description

Sorry to be beating the dead horse on this after the PR was merged that got tests passing...

After #1161 there is now the ability to `pip install` the ioda-converters, and scripts work with that approach. The install, however, instead of installing something broad like `pyiodaconv` takes all of the python subdirectories in `src/` and treats each as a python package. This means that there are installed libraries like so:
```
    /home/cmartin/.local/lib/python3.7/site-packages/compo/*
    /home/cmartin/.local/lib/python3.7/site-packages/conventional/*
    /home/cmartin/.local/lib/python3.7/site-packages/gnssro/*
    /home/cmartin/.local/lib/python3.7/site-packages/goes/*
    /home/cmartin/.local/lib/python3.7/site-packages/gsi_ncdiag/*
    /home/cmartin/.local/lib/python3.7/site-packages/hdf5/*
    /home/cmartin/.local/lib/python3.7/site-packages/ioda_converters-1.0.0.dist-info/*
    /home/cmartin/.local/lib/python3.7/site-packages/land/*
    /home/cmartin/.local/lib/python3.7/site-packages/lib_python/*
    /home/cmartin/.local/lib/python3.7/site-packages/marine/*
    /home/cmartin/.local/lib/python3.7/site-packages/ncep/*
    /home/cmartin/.local/lib/python3.7/site-packages/orig/*
    /home/cmartin/.local/lib/python3.7/site-packages/single_observation/*
    /home/cmartin/.local/lib/python3.7/site-packages/ssec/*
    /home/cmartin/.local/lib/python3.7/site-packages/wrfda_ncdiag/*
```
And that means things have to be imported like:
`import lib_python`

I find this concerning for a few reasons. 1) It's not clear what these libraries are, 2) something like 'hdf5' could conflict with a generic HDF5 library and 3) lib_python means nothing when these are all python libraries...

This is hacked together in the ctests by adding to `$PYTHONPATH` the `src/` directory because what used to work still places certain scripts into `build/lib/pyiodaconv`. These are inconsistent depending on how the converters are built/installed.

Instead, I would think this should be refactored such that there is just one installed library:
```
    /home/cmartin/.local/lib/python3.7/site-packages/pyiodaconv/*
```
and then the `import` statements should be consistent whether it is installed via pip or with ecbuild bundles.

`import pyiodaconv.lib_python`

## Requirements

IODA-converters should be installable and usable with both `pip` and `ecbuild` approach without having to add `src/` to `$PYTHONPATH`

All of the following work without hacks
`pip install`
`ecbuild`

## Dependencies
None

### LINKED ISSUE #1238: [Bug] pyiodaconv is not installed in the correct place
         All of ioda converters python scripts are installed in lib/pyiodaconv forcing users to set the python path twice which is not ideal.

## Expected behavior

The correct installation directory is lib/python(version)/pyiodaconv

PatNichols:
## Description

Install  ioda-conveters python in the proper place. some subdirectory of lib/python(version)/pyiodaconv

- fixes #1238 
- fixes #1187 

Dependencies:

- [ ] merge with JCSDA-internal/ioda/pull/1017


gthompsnJCSDA:
assuming this is in Draft form because there are likely other files in each subdir (such as `src/conventional`) that do not yet appear changed for the IODA path stuff.

PatNichols:
> assuming this is in Draft form because there are likely other files in each subdir (such as `src/conventional`) that do not yet appear changed for the IODA path stuff.

@gthompsnJCSDA  It's in draft form due to ctests not being able to find the bufr python module. Everything else works fine :) . Several of the files in src end up in bin/ so there is no reason yet to change that. The rest end up in /lib/pyiodaconv which  I want to change /lib/python(version)/pyiodaconv.

src/compo/tropess_co_nc2ioda.py
```diff
@@ -19,14 +19,9 @@
 import math
 from numpy import log as ln
 
-IODA_CONV_PATH = Path(__file__).parent/"@SCRIPT_LIB_PATH@"
-if not IODA_CONV_PATH.is_dir():
-    IODA_CONV_PATH = Path(__file__).parent/'..'/'lib-python'
-sys.path.append(str(IODA_CONV_PATH.resolve()))
-
-import ioda_conv_engines as iconv
+import lib_python.ioda_conv_engines as iconv
```
CoryMartin-NOAA, line 22: we should really change this to be `import pyiodaconv.ioda_conv_engines` or something similar. Like I've said before `lib_python` is not a sufficient name, it would be like '#include cpp`

PatNichols: @CoryMartin-NOAA it's doable. I will need to change the directory name from lib_python to something else. Let me give it a look.

BenjaminRuston: @CoryMartin-NOAA fully agree lump it in this one or do in a subsequent PR could be considered

test/CMakeLists.txt
```diff
@@ -442,9 +442,12 @@ set(IODA_CONV_COMP_TOL "0.5e-4")
 #  Dom's fix to get right path
 #set(IODACONV_PYTHONPATH "$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
 
+set(IODACONV_PYTHONPATH "$(CMAKE_BIL$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
```
BenjaminRuston, line 445: this looks like a typo and like it may not be used?

CMakeLists.txt
```diff
@@ -29,16 +29,17 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
+
 set(pyver python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR})
-set(PYDIR ${pyvers}/pyiodaconv)
+set(PYDIR ${pyver}/pyiodaconv)
 
 # Location of installed python iodaconv libraries
 set( PYIODACONV_BUILD_LIBDIR   ${CMAKE_BINARY_DIR}/lib/${PYDIR} )
-set( PYIODACONV_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/${PYDIR} )
+set( PYIODACONV_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX}/lib/${PYDIR} )
```
srherbener, line 39: I've noticed that `<install-prefix>/lib/python<version>/site-packages` seems to be, typically, where setup.py, spack-stack and pip install user built packages. Should we follow that convention here?

What do others think about using the site-packages location?I can go either way (what is in this PR vs site-packages).

I think what is much more important is that we eventually end up installing all ioda-converters and ioda python packages in the same place and that place be findable without having to use PYTHONPATH settings.

CoryMartin-NOAA (approved):
Thanks so much for making this change, it's much appreciated

CMakeLists.txt
PatNichols: @CoryMartin-NOAA @srherbener I checked the Python way of doing this. If you use pip install or there ilk, things will be installed in the site-packages directory of PYTHONHOME. This directory is always in PYTHONPATH unless you clobber this variable it which is not advised at all. There no reason to set site-packages in cmake since it comes for free with a pip install and well it might be confusing to most outside python users. So it looks like we are doing it the "python way"

srherbener: Thanks for following up on this!

srherbener (approved):

BenjaminRuston (approved):
think this is good to go forward understanding we will do another ioda-converter PR to rename the `lib_python` directory to `pyiodaconv` and then modify the `__init__.py` 

to just contain this directory:
`from .pyiodaconv import *`

test/CMakeLists.txt
```diff
@@ -440,11 +440,11 @@ set(IODA_CONV_COMP_TOL "0.5e-4")
 #        ...
 #        )
 #  Dom's fix to get right path
-#set(IODACONV_PYTHONPATH "$<TARGET_FILE_DIR:_ioda_python>/../:$ENV{PYTHONPATH}")
 
 list(APPEND python_path
   "${CMAKE_CURRENT_SOURCE_DIR}/../src"
```
BenjaminRuston, line 445:
```suggestion
```

BenjaminRuston:
on the MacOS M1 spack-stack 1.4.0
```
100% tests passed, 0 tests failed out of 296

```

PatNichols:
@srherbener @BenjaminRuston This is weird.
ubuntu src$ git merge develop
Already up to date.
No conflicts on mac or ubuntu. I will kick off teests on CI again. Perhaps it's a another hiccup with github . BTW github desktop is pure evil....

PR #1243, opened 2023-05-10
update cmake flags and add one for Cray FORTRAN

### LINKED ISSUE #1242: update cmake flags adding Cray FORTRAN
## Description
keep the flags in sync between ioda-converters and ioda

BenjaminRuston:
## Description

update compiler flags keeping in sync with IODA

### Issue(s) addressed

- fixes #1242

### DIFF
--- a/cmake/compiler_flags_Cray_Fortran.cmake
+++ b/cmake/compiler_flags_Cray_Fortran.cmake
@@ -1,10 +1,7 @@
-# (C) Copyright 2009-2016 ECMWF.
+# (C) Copyright 2023 JCSDA
 # 
 # This software is licensed under the terms of the Apache Licence Version 2.0
 # which can be obtained at http://www.apache.org/licenses/LICENSE-2.0. 
-# In applying this licence, ECMWF does not waive the privileges and immunities 
-# granted to it by virtue of its status as an intergovernmental organisation nor
-# does it submit to any jurisdiction.
 
 ####################################################################
 # FLAGS COMMON TO ALL BUILD TYPES
--- a/cmake/compiler_flags_Cray_Fortran.cmake
+++ b/cmake/compiler_flags_Cray_Fortran.cmake
@@ -1,44 +1,56 @@
 # (C) Copyright 2009-2016 ECMWF.
-#
+# 
 # This software is licensed under the terms of the Apache Licence Version 2.0
-# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
-# In applying this licence, ECMWF does not waive the privileges and immunities
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0. 
+# In applying this licence, ECMWF does not waive the privileges and immunities 
 # granted to it by virtue of its status as an intergovernmental organisation nor
 # does it submit to any jurisdiction.
 
 ####################################################################
 # FLAGS COMMON TO ALL BUILD TYPES
 ####################################################################
+set( CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -emf -rmoid -lhugetlbfs")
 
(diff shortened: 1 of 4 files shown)


cmake/compiler_flags_Intel_Fortran.cmake
```diff
@@ -1,40 +1,44 @@
...
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation nor
+# does it submit to any jurisdiction.
 
 ####################################################################
 # FLAGS COMMON TO ALL BUILD TYPES
 ####################################################################
 
-set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -g -traceback")
+set( CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -D_REAL8_")
```
BenjaminRuston, line 13: @srherbener do you know why the *normal* Fortran flags would have `-g -traceback` for iodaconv these should be in `DEBUG` no?

cmake/compiler_flags_Intel_Fortran.cmake
```diff
@@ -1,40 +1,44 @@
-# (C) Copyright 2020-2020 NOAA/NWS/NCEP/EMC
+# (C) Copyright 2009-2016 ECMWF.
```
BenjaminRuston, line 1: @srherbener and @PatNichols what should we put here?

cmake/compiler_flags_GNU_Fortran.cmake
```diff
@@ -1,41 +1,40 @@
...
+# does it submit to any jurisdiction.
 
 ####################################################################
 # FLAGS COMMON TO ALL BUILD TYPES
 ####################################################################
 
-set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -g -fbacktrace")
-
-if(CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
-    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fallow-argument-mismatch")
```
BenjaminRuston, line 13: @srherbener  and @PatNichols  do we want to have `-fallow-agrument-mismatch` ?

cmake/compiler_flags_Cray_Fortran.cmake
```diff
@@ -0,0 +1,53 @@
+# (C) Copyright 2023 JCSDA
```
BenjaminRuston, line 1: @PatNichols please make consistent

PatNichols (approved):
Looks good.

cmake/compiler_flags_Intel_Fortran.cmake
srherbener: I agree with applying the `-traceback` option for only the debug build. The tracebacks from the optimized builds are usually unintelligible.

cmake/compiler_flags_Intel_Fortran.cmake
srherbener: I think the copyright notices look good now. Thanks for fixing those!

fcvdb (approved):
Thanks Ben

cmake/compiler_flags_GNU_Fortran.cmake
srherbener: I agree with removing the `-fallow-argument-mismatch` option. I can't think of a case where you would want to allow argument mismatching (of which I think could lead to some quite difficult debugging efforts).

srherbener (approved):
Thanks for fixing this! I tested ioda-bundle on my Mac, and ioda-bundle built successfully and all of the ioda and ioda-converters tests passed.

PR #1247, opened 2023-05-18
Improved BUFR python example (IODA vars need params)

rmclaren:
## Description

This pull request improves an testinput BUFR python example that writes data into an IODA Netcdf4 file. It turns out that if you don't set the parameter on each variable it (ioda) will pick **0** to represent the fill_value (treated as missing value by Netcdf). 

**Note**: this pull request only affects an example test file and associated output data.

## Associated Bug
https://github.com/JCSDA-internal/ioda/issues/989


emilyhcliu (approved):
Looks good to me.  Thanks for spotting the issue and fixing it.

PR #1252, opened 2023-05-26
Light version

fabiolrdiniz:
## Description

Provide a detailed description of what this PR does. What problem does it fix? What new capability does it add?

## Issue(s) addressed

Resolves #<issue_number>

## Dependencies

- [ ] ...

## Impact

## Checklist


fcvdb (approved):
Thanks Fabio!

BenjaminRuston (approved):
thanks,,, this is now removing things not available in the file,, right @fabiolrdiniz 

the approach is we can set these in UFO at run time as parameters in the yaml

fabiolrdiniz:
Yes, the solar angles we should be able to get through a filter in UFO, but the sensor angles are not clear to me how to extract from the file (I see 28 angles for 24 channels).

Issue #1260, opened 2023-06-06
Query python ResultSet::get_datetime bugs


rmclaren:
## Current behavior (describe the bug)

Query python ResultSet::get_datetime method is returning a regular numpy array not a masked array (with valid fill value etc...). Also the Minute and Second fields are always 0.

BenjaminRuston:
@rmclaren going to close this one assume merge fixed this, please re-open or open new issue as necessary

PR #1261, opened 2023-06-06
Fix some datetime issues.

rmclaren:
## Description

Fixes issues with the Query Python ResultSet get_datetime method. It was returning regular numpy arrays, not masked arrays. Also the minute and second components were always 0.

## Issue(s) addressed

#1260

### DIFF
--- a/test/testinput/bufr_query_python_test.py
+++ b/test/testinput/bufr_query_python_test.py
@@ -38,7 +38,8 @@ def test_basic_query():
     assert len(rad_all.shape) == 2
 
     datetimes = r.get_datetime('year', 'month', 'day', 'hour', 'minute', 'second')
-    assert datetimes[5] == np.datetime64('2020-10-26T21:00:00')
+    assert datetimes[5] == np.datetime64('2020-10-26T21:00:01')
+    assert datetimes.fill_value == np.datetime64('1970-01-01T00:00:00')
 
 
 def test_string_field():
--- a/src/bufr/BufrParser/Query/ResultSet.cpp
+++ b/src/bufr/BufrParser/Query/ResultSet.cpp
@@ -15,6 +15,11 @@
 
 #ifdef BUILD_PYTHON_BINDING
     #include <time.h>
+    #include <pybind11/pybind11.h>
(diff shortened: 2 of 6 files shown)


test/testinput/bufr_query_python_test.py
```diff
@@ -38,7 +38,8 @@ def test_basic_query():
     assert len(rad_all.shape) == 2
 
     datetimes = r.get_datetime('year', 'month', 'day', 'hour', 'minute', 'second')
-    assert datetimes[5] == np.datetime64('2020-10-26T21:00:00')
+    assert datetimes[5] == np.datetime64('2020-10-26T21:00:01')
+    assert datetimes.fill_value == np.datetime64('1970-01-01T00:00:00')
```
nicholasesposito, line 42: Am I understanding this to mean the fill_value is 0?

rmclaren: You shouldn't need to use the fill_value directly unless you are writing datetime data to a netcdf file (see **test/testinput/bufr_query_python_to_ioda_test.py**). The fill values for datetimes are 0 seconds after Jan 1, 1970 (0:00).

BenjaminRuston: we'll need to keep talking about this as a `0` valid and system may be run at or before the ctime epoch

stick a pin in this one ;0)

BenjaminRuston (approved):
working here, fine to move forward

emilyhcliu:
I am testing the bug fix now....

test/testinput/bufr_query_python_test.py
rmclaren: @BenjaminRuston The other possible **fill value** we might consider is the max number of int64 (seconds since the epoch). This will give you a date in the far future... Would this be preferable?

emilyhcliu:
> I am testing the bug fix now....

The fix worked well.  The dateTime values accurate and the type (int64) is correct.

emilyhcliu (commented):
Tested and Approved!

nicholasesposito (approved):

PR #1279, opened 2023-06-27
update to new Bresch decoder from main repo

### LINKED ISSUE #1277: update of obs2ioda
## Description

the NCAR obs2ioda has fallen behind and should be brought up to date so other repairs and fixes can be brought in and routine maintenance can be performed

PatNichols:
## Description

This PR will merge changes from the Jamie Bresch obs2ioda branch into ioda converters.

## Issue(s) addressed

Resolves #1277 

## Impact

No expected impacts to jedi codes.

## Checklist


PatNichols:
> I am working on adding "Launchtime" to radiosonde data. will commit a little later.

thanks @hailingz !

climbfuji:
@PatNichols I just pushed commit https://github.com/JCSDA-internal/ioda-converters/pull/1279/commits/b7d0c584661594f8fe92267159c6e5ea7c82994b, which fixes a compiler error on AWS Parallel Cluster with `intel@2021.6.0`:
```
/usr/bin/ld: failed to convert GOTPCREL relocation; relink with –no-relax
```

PatNichols:
> @PatNichols I just pushed commit [b7d0c58](https://github.com/JCSDA-internal/ioda-converters/commit/b7d0c584661594f8fe92267159c6e5ea7c82994b), which fixes a compiler error on AWS Parallel Cluster with `intel@2021.6.0`:
> 
> ```
> /usr/bin/ld: failed to convert GOTPCREL relocation; relink with –no-relax
> ```

BenjaminRuston:
@hailingz you push now has it changing 855 files!

I had already tried to put in your prepBUFR changes already:
https://github.com/JCSDA-internal/ioda-converters/pull/1279/commits/2569c1c1abbb852a389c888de80773a44560efcf

note that has commit I did previously

hailingz:
OOPS, I did not mean it. how could I reverse my commit then? @BenjaminRuston 
I saw your changes but there is one missing.

PatNichols:
@hailingz Just let me know what needs to be changed. There isn't much change from the ncar obs2ioda so it shouldn't be a big deal to change this.

hailingz:
> @hailingz Just let me know what needs to be changed. There isn't much change from the ncar obs2ioda so it shouldn't be a big deal to change this.

sorry for messing up. 
The only change I meant to make is replace line 187 with
if (ncname == 'dateTime' .or. ncname == 'LaunchTime') then

Thanks!

BenjaminRuston:
@hailingz yeah I don't think this was your change not sure what's up

hailingz:
> @hailingz yeah I don't think this was your change not sure what's up

I do not know what happened. Did what I usually do to commit and push :(

PatNichols:
> > @hailingz yeah I don't think this was your change not sure what's up
> 
> I do not know what happened. Did what I usually do to commit and push :(

@hailingz I had to revert your changes. Every file in the tree except ncio_mod were deleted :)

PatNichols:
@hailingz @BenjaminRuston Fixed with revert and changed to ncio_mod to what it should be?
Also changed the ncio_mod from *.h5 to *.nc4 ( I think that's right).

cmake/compiler_flags_GNU_Fortran.cmake
```diff
@@ -8,25 +8,25 @@
 # FLAGS COMMON TO ALL BUILD TYPES
 ####################################################################
 
-set( CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -g -ffree-line-length-none")
+set( CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -g -ffree-line-length-none -mcmodel=medium")
```
BenjaminRuston, line 11: think the `mcmodel` is a Intel compiler flag not `gnu` I do not see it in the gnu flags

PatNichols: @BenjaminRuston it's a GNU flag. Both intel and gnu used the same backend up until recently. Intel is moving to clang (may have already done so with OneAPI). It's a cpu dependent  flag for arm and x86.
https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html

climbfuji: @PatNichols You probably saw that pushed those flags for Intel only directly in the CMakeLists in the obs2ioda subdirectory. Feel free to take those out and make them consistent with the gnu flags, or vice versa. Thanks!

PatNichols:
There is a gotcha for the clang build. Since we have to use the fortran compiler from gnu, the "-mcmodel=medium" flags causes a problem for a  linking library built with clang since they cannot be linked due to the conflicting memory models.

cmake/compiler_flags_GNU_Fortran.cmake
PatNichols: @climbfuji Thanks! I have migrated those changes into our cmake/compiler_flags.

PatNichols:
> There is a gotcha for the clang build. Since we have to use the fortran compiler from gnu, the "-mcmodel=medium" flags causes a problem for a linking library built with clang since they cannot be linked due to the conflicting memory models.

This is fixed now. We should at some point separate compiler flags for macos apple clang and clang on linux which are not the same system.

PatNichols:
This PR still needs an updated testoutput file to compare to for ctests.

PatNichols:
@BenjaminRuston  got rid of the mopitt file.

PR #1291, opened 2023-07-13
Update ATMS converter yaml for EWOK ingest suite

ashley314:
## Description

An issue was found when running `bufr2ioda.x` with `test/testinput/bufr_ncep_atms_remap.yaml` on AWS Pcluster. For some reason the compiled code looks for the `sensor: atms` field. This does not happen when running on my mac. 

## Impact

I tested the additional line on the my working copy of ioda-converters on my mac with no change in capabilities. It also then allows `bufr2ioda.x` to run on the pcluster. Note, I am not sure who all uses this template (maybe ctests) but I think the impact should be minimal across the board if there is any.

## Checklist


ashley314:
@BenjaminRuston @PatNichols When you get a minute can you take a look at this? I havn't had any issues and tested on the pcluster and my mac. Although I do not know who or what also uses `test/testinput/bufr_ncep_atms_remap.yaml`. Thanks!

srherbener (approved):
I think this change is fine, but I don't have knowledge of the impact (if any) of making this change. It would be good to have someone with knowledge of this ATMS converter to review this.

test/testinput/bufr_ncep_atms_remap.yaml
```diff
@@ -179,4 +180,5 @@ observations:
           source: variables/remappedBT
           longName: "3-by-3 Averaged Brightness Temperature"
           units: "K"
+          range: [120, 500]
```
BenjaminRuston, line 183: @emilyhcliu this is the range in a prior ATMS converter the high value is a bit high, and the low is actually a bit close.

would you be OK with range: [80, 380]

or do you think this is a nothing-burger ;0)

BenjaminRuston (approved):
this should be good to go @smaticka this will add the `range` attribute for you as well

PR #1305, opened 2023-07-20
Fix for missing gsl-lite dependency

rhoneyager-tomorrow:
## Description

Hi @srherbener, @PatNichols, and @rmclaren! Just passing through and exploring the new Skylab release.

Many of the bufr-related codes include gsl-lite headers ([example](https://github.com/JCSDA-internal/ioda-converters/blob/develop/src/bufr/BufrParser/Query/DataProvider/DataProvider.h#L16)). The CMake code does not specify gsl-lite as a dependency, and this causes a compilation error. I am rather surprised that CI didn't catch this. This was noticed using CMake 3.23.1 / ecbuild 3.7.2 with Spack stack 1.4.0. 

There are two issues here:
1. `print_queries.x` and the Python `bufr` lib (both in `src/bufr/CMakeLists.txt`) both forget to link to gsl-lite.  The check for gsl-lite is also missing from the top-level CMakeLists.txt when setting `iodaconv_bufr_python_ENABLED`. 
2. The `ingester` executable *should* have transitively imported gsl-lite since it is [linked publicly with ioda-engines](https://github.com/JCSDA/ioda/blob/develop/src/engines/ioda/CMakeLists.txt#L257C1-L257C57). But, it somehow didn't, and there might be a CMake bug hidden somewhere.

Also note: it is a bit confusing to call `find_package( bufr QUIET )` and later define a Python library that is also named `bufr`.

## Checklist


PatNichols (approved):
thanks Ryan! I did not even notice the gsl-lite dependency.

BenjaminRuston:
@danholdaway and @CoryMartin-NOAA adding you notice the added dependency assume OK for you NOAA EMC, NASA GMAO this too should be in the stack already

CoryMartin-NOAA:
Thanks @BenjaminRuston this is already a requirement for IODA I believe, so it should be no additional dependency for our purposes.

CoryMartin-NOAA (approved):

rhoneyager-tomorrow:
For context, gsl-lite was added to the stack in 2020 to backport a feature from C++20 ([`std::span`](https://en.cppreference.com/w/cpp/container/span)). It's required for ioda. Ron's code has always depended on it.

It's always hard to detect issues like this. Compilers have their own search locations outside of where CMake and ecbuild tell them to look for header files, and there's probably a copy of gsl-lite floating around somewhere in the CI container image.

srherbener:
Hi @rhoneyager-tomorrow! Good to hear from you. Hope things are going well.

Thanks for fixing this!

PR #1307, opened 2023-07-20
Git-lfs error bugfix for mopitt file

PatNichols:
## Description

This removes a file that slipped through our git-lfs filter. 

## Issue(s) addressed

Resolves #1306 

## Dependencies

None

## Impact

Expected impact on downstream repositories:

None

## Checklist

- [x ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have run the unit tests before creating the PR

PR #1309, opened 2023-07-21
Bugfix/git lfs bug mopitt 2

PatNichols:
## Description

This adds in the mopitt_co test and file that was removed to fix the git-lfs bug

## Issue(s) addressed

Resolves #1307 

## Dependencies

None

## Impact

None


test/CMakeLists.txt
```diff
@@ -1143,17 +1143,17 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_tropomi_no2_tropo
...
-#                          -r 2021092903 2021092921"
-#                  mopitt_co.nc ${IODA_CONV_COMP_TOL_ZERO})
+ecbuild_add_test( TARGET  test_${PROJECT_NAME}_mopitt_co
+                  TYPE    SCRIPT
+                  ENVIRONMENT "PYTHONPATH=${IODACONV_PYTHONPATH}"
+                  COMMAND bash
+                  ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                          netcdf
+                          "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/mopitt_co_nc2ioda.py
+                          -i testinput/mopitt_co.h5
```
BenjaminRuston, line 1153: we need to add it back as `.he5`

do whatever we need to do as these are correct file extensions.. think it involves re-writing the git lfs log

PR #1323, opened 2023-08-09
Change Ozone Variable Name and Turn on Error

danholdaway:
## Description

Some small fixes needed for current GMAO testing. 

- The variable for instruments like MLS should be ozoneProfile, rather than ozoneTotal.
- Some of our testing relies on use of the GSI initial error. We are working to set the error in the YAML instead but right now the error isn't being set. Either it should be set to something sensible or the variable removed entirely. Here we set it to something sensible.

## Issue(s) addressed

N/A

## Dependencies

N/A

## Impact

N/A

## Checklist

- [NA] I have made corresponding changes to the documentation


karpob:
To match the new convention you'll also need to change ozoneLayer to ozoneTotal in the If statement

src/gsi_ncdiag/gsi_ncdiag.py
gmao-jjin3, line 1: I made an update in the condition to save GSI "Forecast_adjusted" in radiance data.

gmao-jjin3 (approved):

BenjaminRuston:
seeing two failed tests:
```
iodaconv_gsi_ncdiag_coding_norms
test_iodaconv_gsidiag_rad
```
--

[iodaconv_gsi_ncdiag_coding_norms](http://cdash.jcsda.org:8080/test/1x2521753)
[test_iodaconv_gsidiag_rad](http://cdash.jcsda.org:8080/test/12521810)

PatNichols:
@danholdaway do you need to update the output file?

BenjaminRuston:
update of `testOutput` likely needed

danholdaway:
OK thanks. I will take a look.

src/gsi_ncdiag/gsi_ncdiag.py
danholdaway: Thanks @gmao-jjin3

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -1795,9 +1796,9 @@ def toIODAobs(self, OutDir, clobber=True):
                 LocVars.append(ncv)
 
         nlocs = self.nobs
-        vname = "ozoneTotal"
+        vname = "ozoneProfile"
```
emilyhcliu, line 1799: One question: Is this variable defined for ozone measurement with units in mol/mol?

karpob: yes. MLS and OMPS-LP are profile measurements that are mixing ratio, and not DU.

karpob: Actually, you're probably right, it should be ppmv, but last I checked the variable mole_fraction_ozone_in_air was in ppmv. Depending on your point of view, this could also be considered wrong.

danholdaway: Apologies, I'm not following. Is there something to change?

emilyhcliu: In IODA convention table, there are several ozone-related variables with different units:
- ozoneTotal in DU
- ozoneLayer in DU
- ozoneColumn in mol/m2
- ozoneProfile in mol/mol
- ozoneSurface in kg/kg

So, depending on the ozone obs used, the ozone variables should be named based on their units.

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -1803,12 +1806,8 @@ def toIODAobs(self, OutDir, clobber=True):
         varDict[vname]['errKey'] = vname, iconv.OerrName()
         varDict[vname]['qcKey'] = vname, iconv.OqcName()
         VarDims[vname] = ['Location']
-        if (self.sensor in oz_lev_sensors):
-            varAttrs[varDict[vname]['valKey']]['units'] = 'mol mol-1'
-            varAttrs[varDict[vname]['errKey']]['units'] = 'mol mol-1'
-        else:
-            varAttrs[varDict[vname]['valKey']]['units'] = 'DU'
-            varAttrs[varDict[vname]['errKey']]['units'] = 'DU'
+        varAttrs[varDict[vname]['valKey']]['units'] = units_values[vname]
```
danholdaway, line 1809: @emilyhcliu @karpob I switched to just using units_values to set the units so now it just depends on the name. Then I added the complete list of ozone variables to units_values. I don't think this is a change from what was happening before but perhaps the code is a bit clearer?

karpob:
@danholdaway I think you might need to indent twice for that "if" line continuation coding norm error.

danholdaway:
I think it's ready now. Thanks all.

gmao-jjin3:
@PatNichols @BenjaminRuston  I think this one needs to be merged ASAP, so that we check it out and re-create our data.

BenjaminRuston (approved):
this looks ready

BenjaminRuston:
@jeromebarre just FYI, note this was merged please let us know if you see anything that may need attention

src/gsi_ncdiag/gsi_ncdiag.py
jeromebarre: I think in the future we could get rid of ozoneSurface as ozoneProfile would be the same, i.e. a vertical interpolation of a point on the geoval profile. To clear the confusion given the comments above, the unit is not that important as DU to **ppmV** (**volume** mixing ratio in mol/mol) to **ppm** (**mass** mixing ratio in kg/kg) is a matter of a single coefficient multiplication. FYI, Some surface measurements, i.e. Europe for ozone are reported in ug/m3. Conversion is then not straightforward and requires P and T within the measurements. This is not that relevant here. Ideally, JEDI should use mixing ratios and not densities. Please try to add people from the JCSDA atmospheric composition team for trace gas and aerosol related PRs. Thanks!

PR #1334, opened 2023-08-18
bug fix to make QCPath values out of 0 to 31 valid

ypwang19:
## Description
In the the VIIRS AOD Level 2 data file, QCPath is set with a valid range of 0-31. The values out of 0-31 are not read in properly by the converters, so that the retrieval surface is not correctly identified. This PR is to fix this minor bug in reading QCPath, so that values out of 0-31 are valid. 

## Dependencies

- None.

## Checklist

### DIFF
--- a/src/compo/viirs_aod2ioda.py
+++ b/src/compo/viirs_aod2ioda.py
@@ -99,6 +99,7 @@ class AOD(object):
             vals = ncd.variables['AOD550'][:].ravel()
             errs = ncd.variables['Residual'][:].ravel()
             qcpath = ncd.variables['QCPath'][:].ravel()
+            qcpath = np.where(qcpath>-1,qcpath,np.nan)
             qcall = ncd.variables['QCAll'][:].ravel().astype('int32')
             obs_time = np.full(np.shape(qcall), base_datetime, dtype=object)
             if self.mask == "maskout":
--- a/src/compo/viirs_aod2ioda.py
+++ b/src/compo/viirs_aod2ioda.py
@@ -98,8 +98,13 @@ class AOD(object):
             lats = ncd.variables['Latitude'][:].ravel()
             vals = ncd.variables['AOD550'][:].ravel()
             errs = ncd.variables['Residual'][:].ravel()
-            qcpath = ncd.variables['QCPath'][:].ravel()
-            qcpath = np.where(qcpath>-1, qcpath, np.nan)
+
+            # QCPath is the flag for retrieval path. The valid range is 0-127 in the
+            # ATBD: https://www.star.nesdis.noaa.gov/jpss/documents/ATBD/ATBD_EPS_Aerosol_AOD_v3.4.pdf.
+            # QCPath's valid range in the input file is not correct, so we define the valid range here.
+            qcpath = ncd.variables['QCPath'][:].data.ravel()
+            qcpath = np.ma.masked_array(qcpath, np.logical_or(qcpath < 0, qcpath > 127))
+
             qcall = ncd.variables['QCAll'][:].ravel().astype('int32')
             obs_time = np.full(np.shape(qcall), base_datetime, dtype=object)
             if self.mask == "maskout":
--- a/src/compo/viirs_aod2ioda.py
+++ b/src/compo/viirs_aod2ioda.py
@@ -99,7 +99,7 @@ class AOD(object):
             vals = ncd.variables['AOD550'][:].ravel()
             errs = ncd.variables['Residual'][:].ravel()
             qcpath = ncd.variables['QCPath'][:].ravel()
-            qcpath = np.where(qcpath>-1,qcpath,np.nan)
+            qcpath = np.where(qcpath>-1, qcpath, np.nan)
             qcall = ncd.variables['QCAll'][:].ravel().astype('int32')
             obs_time = np.full(np.shape(qcall), base_datetime, dtype=object)
             if self.mask == "maskout":
--- a/test/testoutput/viirs_aod.nc
+++ b/test/testoutput/viirs_aod.nc
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b20ef47567288c23913bd1768e4cb39c061e584b39a076af5d16e6b01a1e9508
+oid sha256:9e44cadb8e1b55c333ee01338afafe83cbd73bc0637b42917dc0cf1e0e3d0025
 size 14272


src/compo/viirs_aod2ioda.py
```diff
@@ -99,6 +99,7 @@ def _read(self):
             vals = ncd.variables['AOD550'][:].ravel()
             errs = ncd.variables['Residual'][:].ravel()
             qcpath = ncd.variables['QCPath'][:].ravel()
+            qcpath = np.where(qcpath>-1,qcpath,np.nan)
```
CoryMartin-NOAA, line 102:
```suggestion
            qcpath = np.where(qcpath>-1, qcpath, np.nan)
```
I hope this suggestion makes it pass coding norms, may need to adjust the greater than too

jeromebarre (changes requested):
This PR replaces the negative values in QCPath to `np.nan`. I am not sure from which values QC path is supposed to range. Cannot really find the documentation of the product that late on a Friday night. Would be good to have a link to the product documentation in the converter description (see for example trace gas converters). For the negative values, they might mean something. If they are -999 or something like that, we would might not want to throw NaNs but we should filter out those values using the already existing mask in the converter. Thanks a lot!

jeromebarre:
And the coding norms tests is still failing. Just run the first test (compo coding norms) with the -VV option. It specify where your syntax is failing the test. I usually do that right before or right after I issue a PR in iodaconv. Thanks again.

ypwang19:
> This PR replaces the negative values in QCPath to `np.nan`. I am not sure from which values QC path is supposed to range. Cannot really find the documentation of the product that late on a Friday night. Would be good to have a link to the product documentation in the converter description (see for example trace gas converters). For the negative values, they might mean something. If they are -999 or something like that, we would might not want to throw NaNs but we should filter out those values using the already existing mask in the converter. Thanks a lot!


The variable in the data file has the following attributes:
```
byte QCPath(Rows, Columns) ;
                QCPath:long_name = "Flags for retrieval path (0-No/1-Yes): bit 0: retrieval over water; bit 1: over bright land; bit 2: over glint water; bit 3: retrieval with SW scheme over land; bit 4: retrieval with SWIR scheme over land; bit 5: retrieval over bright-land algorithm, bit 6: dark land retrieval over bright land" ;
                QCPath:coordinates = "Longitude Latitude" ;
                QCPath:_FillValue = -128b ;
                QCPath:valid_range = 0b, 31b ;
                QCPath:units = "1" ;
```
The QCpath is supposed to range 0-127. The attribute `QCPath:valid_range = 0b, 31b` in the data file is a mistake, based on the communication with Hongqing from NESDIS data product team, and based on their document: https://www.star.nesdis.noaa.gov/jpss/documents/ATBD/ATBD_EPS_Aerosol_AOD_v3.4.pdf

Would it be better to set values below 0 to its missing value -128 (right after the the read in line) like this:
` qcpath = np.where(qcpath >= 0.0, qcpath, ncd.variables['QCPath']._FillValue)
`
and let the converter do the further quality control job? This line is just making values out of the "0-31 valid range" be valid, instead of "--".
I will also add some description near this line in the converter.

jeromebarre:
> > This PR replaces the negative values in QCPath to `np.nan`. I am not sure from which values QC path is supposed to range. Cannot really find the documentation of the product that late on a Friday night. Would be good to have a link to the product documentation in the converter description (see for example trace gas converters). For the negative values, they might mean something. If they are -999 or something like that, we would might not want to throw NaNs but we should filter out those values using the already existing mask in the converter. Thanks a lot!
> 
> The variable in the data file has the following attributes:
> 
> ```
> byte QCPath(Rows, Columns) ;
>                 QCPath:long_name = "Flags for retrieval path (0-No/1-Yes): bit 0: retrieval over water; bit 1: over bright land; bit 2: over glint water; bit 3: retrieval with SW scheme over land; bit 4: retrieval with SWIR scheme over land; bit 5: retrieval over bright-land algorithm, bit 6: dark land retrieval over bright land" ;
>                 QCPath:coordinates = "Longitude Latitude" ;
>                 QCPath:_FillValue = -128b ;
>                 QCPath:valid_range = 0b, 31b ;
>                 QCPath:units = "1" ;
> ```
> 
> The QCpath is supposed to range 0-127. The attribute `QCPath:valid_range = 0b, 31b` in the data file is a mistake, based on the communication with Hongqing from NESDIS data product team, and based on their document: https://www.star.nesdis.noaa.gov/jpss/documents/ATBD/ATBD_EPS_Aerosol_AOD_v3.4.pdf
> 
> Would it be better to set values below 0 to its missing value -128 (right after the the read in line) like this: `qcpath = np.where(qcpath >= 0.0, qcpath, ncd.variables['QCPath']._FillValue)` and let the converter do the further quality control job? This line is just making values out of the "0-31 valid range" be valid, instead of "--". I will also add some description near this line in the converter.

We can either use the mask or set it as a value that we know how to filter in the obsspace during a JEDI execution. If you choose to do the latter please provide a test/example in UFO. Thanks!

ypwang19:
> 

OK, I will mask out values out of 0-127. The QCPath will not go to the observation file or go through JEDI execution. It is only used to identify surface to assign observation error.

CoryMartin-NOAA:
So my understanding of the issue here was that QCPath was not being properly converted from a bitmap to an integer. The QCPath is only used in computing the observation errors, otherwise it is ignored and not passed to the output file. Thus this PR is only changing the assigned initial errors and not masking out any observations.

jeromebarre:
Ok sorry, I got confused and went too fast on this PR review. So QCpath is used to define the error here and not stored in the ioda file. So this PR changes only the way the error is specified. Meaning more data points now fall under: https://github.com/JCSDA-internal/ioda-converters/blob/d58dce0e5ef429fd3e58cde42512de004f6a2017/src/compo/viirs_aod2ioda.py#L135 I presume.
One of my question still remains though. Don't we want to get rid of the datapoints that have a wrong QCPath?

ypwang19:
> Ok sorry, I got confused and went too fast on this PR review. So QCpath is used to define the error here and not stored in the ioda file. So this PR changes only the way the error is specified. Meaning more data points now fall under:
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/d58dce0e5ef429fd3e58cde42512de004f6a2017/src/compo/viirs_aod2ioda.py#L135
> 
> I presume.
> One of my question still remains though. Don't we want to get rid of the datapoints that have a wrong QCPath?

Yes, more data fall under dark land error. The QCPath data is fine, the issue is they put a wrong valid range attribute to this variable by mistake. We fix it in this PR.

jeromebarre:
Thanks for clarifying. The PR description was unclear to me at first.

PR #1335, opened 2023-08-21
Without this the code does not compile

rtodling:
## Description

Without the fix here - declaration of variable - this code did not compile ... not sure how this passed any test!

## Issue(s) addressed

Resolves #<issue_number>

## Dependencies

- NOT SURE.

## Impact

Without it: code does not compile

## Checklist

There are too many kinds.[f,F]0 around ... this should be converted in iodaconv_kinds.f90 and all codes referring to the original kinds should be changed accordingly.


PatNichols (approved):
Thanks for catching this.

climbfuji:
@PatNichols @srherbener There is a discussion for a similar PR in vader: https://github.com/JCSDA-internal/vader/pull/194#issuecomment-1686501789

PatNichols:
@rtodling @climbfuji @srherbener There is a problem defining long this way. It should be defined as c_long or just use the iso_fortran_env types instead of doing this at all. Intel's math library needs special flags if the underlying memory model is not what it seems. My own preference is to define these types directly in the source code and forgo having such a header since any c interfaces need to be well defines in terms of types ( is this 64bit vs 32 vs 128 etc) to work correctly.

PatNichols:
> I think this is the issue: there are two source files in the ioda converter code that define the same module, kinds.
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/aa38b19a061d61cc9796e7329494497f71575546/src/ncar-bufr2nc-fortran/kinds.f90#L1
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/aa38b19a061d61cc9796e7329494497f71575546/src/bufr/BufrParser/Exports/Variables/Transforms/atms/kinds.F90#L1
> 
> These appear to have been created from a copy-paste of the same module definition and they are slightly different. The ncar-bufr2nc-fortran version contains the i_long definitions and the bufr version does not.
> 
> I suspect the build success/failure behavior on different platforms is coming from the order in which the modules are searched for.
> 
> I think the proper solution is to delete the bufr kinds.F90 file (without i_long) and move the ncar-bufr2nc-fortran kinds.f90 file (with i_long) to a common area where both source code subdirectories can share the common kinds module.

@srherbener Just to add confusion, I think oops defines their own kinds too. Also as I mentioned in my previous comment the intel math library need some flags under certain conditions. int32 and int64 make sense in fortran not long and llong.

BenjaminRuston:
@rtodling we are tempted to just follow @srherbener suggestion.

@PatNichols will go ahead and branch off your branch and move the type kinds out of atms and ncar-bufr2ioda and consolidate them in a higher level location. then you can verify if this fixes the issue for you, does this sound OK?

danholdaway:
@PatNichols @BenjaminRuston can we fix this with a fair degree of urgency please? This is impacting our ability to build on Discover so puts all of us at GMAO on hold.

srherbener:
@PatNichols thanks for creating PR #1338. Does PR #1338 supersede this PR, ie can we close this PR? Thanks!

PatNichols:
> @PatNichols thanks for creating PR #1338. Does PR #1338 supersede this PR, ie can we close this PR? Thanks!

@rtodling @srherbener The plan is to merge #1338 into this one. Then merge into develop. @rtodling could you review the #1338? Thanks

srherbener (approved):

src/utils/iodaconv_kinds.f90
```diff
@@ -1,4 +1,13 @@
...
+!/*
+! * (C) Copyright 2009-2032 UCAR
+! *
+! * This software is licensed under the terms of the Apache Licence Version 2.0
+! * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+! * In applying this licence, ECMWF does not waive the privileges and immunities
+! * granted to it by virtue of its status as an intergovernmental organisation nor
+! * does it submit to any jurisdiction.
+! */
+module iodaconv_kinds
```
climbfuji, line 10: Thanks for fixing that name!

climbfuji (approved):
I haven't run any tests, but from the conversation and parsing the code changes this all looks good.

PR #1338, opened 2023-08-22
Bugfix/unify kinds

### LINKED ISSUE #1336: kinds.f90 appears in 2 different places in ineqivalent form

@rtodling reports that ioda converters does not compile with the kinds.f90 in bufr. A separate file of the same name exists in ncar-bufr. We need to have only 1 kinds.f90 in the repo and it needs to compile.

PatNichols:
## Description

This unifies the kinds.f90 into 1 file that resides in a new utils directory. At some point we should move all code that is not sensor specific here to make the repo easier to maintain and alleviate technical debt.

## Issue(s) addressed

Resolves #1336


BenjaminRuston:
@rtodling, @rmclaren and @srherbener does this satisfy?
It consolidates the kinds.[fF]90 files into a single one in src/utils/iodaconv_kinds.F90 which then can be maintained

srherbener (approved):
This looks great. Thanks for doing this consolidation!

BenjaminRuston:
@CoryMartin-NOAA something popped up for Ricardo, drove us to consolidate the FORTRAN kinds in iodaconv this look fine to you

CoryMartin-NOAA (approved):

BenjaminRuston (approved):
this should work @rtodling solve your problem

rtodling:
This is the way to go. Thanks.

PR #1350, opened 2023-08-31
bug fixes for the tempo converter

jeromebarre:
## Description

In the proxy data there are a lot of inconsistencies in how the mask is specified from a file to another.
The changes allow the converter to run on proxy files and produce ioda files for our tests period.

## Dependencies

None

## Impact

None

## Checklist

- [ x] I have run the unit tests before creating the PR

### DIFF
--- a/src/compo/tempo_nc2ioda.py
+++ b/src/compo/tempo_nc2ioda.py
@@ -124,7 +124,6 @@ class tempo(object):
                 sfp = np.ma.array(sfp, mask=mask)
                 ak = ncd.groups['support_data'].variables['surface_pressure'].Eta_A
                 bk = ncd.groups['support_data'].variables['surface_pressure'].Eta_B
-                print(np.shape(mask))
                 preslev = hPa2Pa * np.transpose(ak[:, np.newaxis] + np.outer(bk, sfp))
                 preslev = np.ma.array(preslev, mask=np.repeat(mask, levels+1))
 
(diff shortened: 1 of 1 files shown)


src/compo/tempo_nc2ioda.py
```diff
@@ -121,6 +124,7 @@ def _read(self):
                 sfp = np.ma.array(sfp, mask=mask)
                 ak = ncd.groups['support_data'].variables['surface_pressure'].Eta_A
                 bk = ncd.groups['support_data'].variables['surface_pressure'].Eta_B
+                print(np.shape(mask))
```
BenjaminRuston, line 127: is this print still needed, if so a little more description would likely make it even more helpful ;0)

BenjaminRuston (approved):
bring on some TEMPO data

src/compo/tempo_nc2ioda.py
jeromebarre: no it was just for testing, removing it now

src/compo/tempo_nc2ioda.py
```diff
@@ -70,6 +70,7 @@ def _read(self):
         # loop through input filenames
         first = True
         for f in self.filenames:
+            print(f)
```
mer-a-o, line 73: there's one more print here.

mer-a-o (approved):

PR #1351, opened 2023-09-01
Add table of central wavenumber to iasi converter

### LINKED ISSUE #1313: Cleanup plain old radiance vs spectral radiance vs scaled radiance
The names here are confusing and are not consistent across the converters, ufo transformations, and the IODA conventions. They are also not consistently named within the broader field.

The [IODA conventions spec](https://docs.google.com/spreadsheets/d/1fHb2QIHOwgp4d6plyn9Dj9HZFZsDRxcPaUV6hdTb5BU/edit#gid=0) and [Wikipedia](https://en.wikipedia.org/wiki/Radiance) define radiance following a common convention in radiometry. It is the radiant flux emitted, reflected, transmitted or received by a given surface, per unit solid angle per unit projected area. Units are *W m-2 sr-1*.

Spectral radiance is the radiance of a surface per unit frequency, or wavelength, or wavenumber. There are also scaled radiances, which represent compressed data that need to be transformed to get regular radiances. These are all distinct variables in the [BUFR code tables](https://confluence.ecmwf.int/display/ECC/WMO%3D35+element+table), and these all have different units.

Here are a few spots where these are confused:

- In [bufr_ncep_mtiasi.yaml](https://github.com/JCSDA-internal/ioda-converters/blob/develop/test/testinput/bufr_ncep_mtiasi.yaml#L239), it is unclear if a spectral radiance or a regular radiance is being used. The comments say [spectral radiance](https://github.com/JCSDA-internal/ioda-converters/blob/develop/test/testinput/bufr_ncep_mtiasi.yaml#L89C15-L89C37), but [SCRA](https://www.nco.ncep.noaa.gov/sib/jeff/bufrtab_tableb.html) is a plain old scaled radiance. Spectral radiance (SPRD) would be a different mnemonic.

- [bufr_iasi.yaml](https://github.com/JCSDA-internal/ioda-converters/blob/develop/test/testinput/bufr_iasi.yaml#L55) is writing scaled radiance as a regular radiance. These have different units (`W m-2 sr-1` vs `W m-2 sr-1 m`). Comments in yaml do not get carried through to end users, and they end up with a variable that does not contain what they expect.

- In [UFO's SatBrightnessTempFromRad transform](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/jedi-components/ufo/variabletransforms/transforms/btfromrad.html), the radiance variable is being treated as a spectral radiance variable. This is doubly awkward as a user needs to manually specify if the spectral radiance is expressed in terms of frequency, wavelength, or wavenumber.

@gthompsnJCSDA / @emilyhcliu / @BenjaminRuston would you mind disambiguating these cases? If users are writing spectral radiances and scaled radiances in IODA files, then these variables should be documented in the IODA conventions.

smaticka:
## Description
Rather than pulling central wavenumber per experiment in the skylab yaml, we are adding a static table of central wavenumber values to be exported during the ioda convertion. 

Need to add an 'assign method' in ioda-converter yaml for iasi that looks up a table of values. 

## Issue(s) addressed

Resolves #1313


test/testinput/bufr_ncep_mtiasi.yaml
```diff
@@ -20,6 +20,72 @@ observations:
...
+                     74825.0, 74875.0, 74925.0, 75050.0, 75125.0, 75175.0, 75275.0, 75300.0, 75325.0, 75450.0,
+                     75525.0, 75600.0, 75725.0, 75900.0, 75950.0, 76275.0, 76400.0, 76550.0, 77200.0, 77350.0,
+                     78125.0, 78275.0, 78450.0, 78625.0, 78750.0, 78800.0, 78925.0, 79075.0, 79325.0, 80100.0,
+                     80625.0, 81025.0, 81175.0, 82100.0, 82950.0, 83375.0, 84400.0, 86150.0, 87125.0, 87500.0,
+                     90150.0, 90625.0, 91725.0, 91925.0, 92500.0, 92800.0, 93800.0, 94250.0, 94325.0, 95025.0,
+                     96250.0, 96550.0, 97925.0, 99700.0, 99825.0, 99975.0, 100075.0, 100150.0, 100225.0, 100325.0,
+                     100475.0, 100525.0, 100600.0, 100725.0, 100825.0, 100975.0, 101050.0, 101200.0, 101325.0, 101450.0,
+                     101550.0, 101650.0, 101825.0, 101875.0, 102025.0, 102100.0, 102200.0, 102225.0, 102300.0, 102425.0,
+                     102500.0, 102625.0, 102700.0, 102775.0, 102875.0, 102900.0, 103000.0, 103100.0, 103175.0, 103300.0,
+                     103475.0, 103675.0, 103825.0, 103950.0, 104050.0, 104100.0, 104150.0, 104625.0, 105125.0, 105450.0,
```
rmclaren, line 50: Needs to go under `globals` in the ioda section.

Are these values going to be the same for all IASIA files?

rmclaren (changes requested):

rmclaren:
If you use gettab on this BUFR file you get the following mnemonics:

`gettab  test/testinput/gdas.t12z.mtiasi.tm00.bufr_d | grep WAVE `

```
| WVNM     | 006029 | WAVE NUMBER                                              |
| LOGRCW   | 025076 | LOG-10 OF (TEMPERATURE-RADIANCE CENTRAL WAVENUMBER)      |
```

but when you run `print_queries.x test/testinput/gdas.t12z.mtiasi.tm00.bufr_d` you find out these mnemonics are not available in the test dataset...

You can add a global like you tried, but I'm not sure this this is the right answer as I'm guessing these wavenumbers will be different for different IASI files (?). For more flexibility you could read the data using the python API.

test/testinput/bufr_ncep_mtiasi.yaml
rmclaren: I think we were talking about an extension that could get values from a text file in the future, but better bet right now might be to use python to read the file instead. Please see the following examples:

https://github.com/JCSDA-internal/ioda-converters/blob/develop/test/testinput/bufr_query_python_to_ioda_test.py
https://github.com/JCSDA-internal/ioda/blob/develop/src/engines/test/ioda/python/test_ioda_obsspace.py

BenjaminRuston: these values are the same for all `IASI` MetOp-A, -B and -C

rmclaren: Cool! So this should work as long as it's moved into the ioda::globals section of the YAML configuration file.

smaticka:
> If you use gettab on this BUFR file you get the following mnemonics:
> 
> `gettab test/testinput/gdas.t12z.mtiasi.tm00.bufr_d | grep WAVE `
> 
> ```
> | WVNM     | 006029 | WAVE NUMBER                                              |
> | LOGRCW   | 025076 | LOG-10 OF (TEMPERATURE-RADIANCE CENTRAL WAVENUMBER)      |
> ```
> 
> but when you run `print_queries.x test/testinput/gdas.t12z.mtiasi.tm00.bufr_d` you find out these mnemonics are not available in the test dataset...
> 
> You can add a global like you tried, but I'm not sure this this is the right answer as I'm guessing these wavenumbers will be different for different IASI files (?). For more flexibility you could read the data using the python API.

Thanks for checking this @rmclaren. The input test data file hasn't been updated yet, but was the next planned task. Just wanted to get the draft created for conversation on the table entry to begin. I'll work on the new file next.

smaticka:
> Cool. I should probably update the print_queries.x output to show the description information so that you don't need to switch back and forth... Please be aware that the `gettab` output often does not really reflect the data in the file (lists subsets that don't exists, often does not list subsets actually in the file). `print_queries.x` will show you whats actually available in the file.

Gotcha. that makes sense then b/c `WVNM` was still in the output of `gettab`, but it's not in the `print_queries.x` output.

test/testinput/bufr_ncep_mtiasi.yaml
```diff
@@ -6,7 +6,7 @@
 observations:
   - obs space:
       name: bufr
-      obsdatain: "./testinput/gdas.t12z.mtiasi.tm00.bufr_d" 
+      obsdatain: "test/testinput/gdas.mtiasi.t00z.20220216.bufr"
```
PatNichols, line 9: @smaticka  Looks like the path is wrong here. Perhaps ./testinput/gdas.mtiasi.t00z.20220216.buf ?

smaticka: hmm, when I tried that it failed, but maybe I was in the wrong directory. 
I'll try again from `test`

rmclaren: When you run the unit tests the working directory is set to  **_\<ioda-converters dir\>_/test**. The paths need to be relative to this directory.

test/testinput/bufr_ncep_mtiasi.yaml
```diff
@@ -6,7 +6,7 @@
 observations:
   - obs space:
       name: bufr
-      obsdatain: "./testinput/gdas.t12z.mtiasi.tm00.bufr_d" 
+      obsdatain: "./testinput/gdas.mtiasi.t00z.20220216.bufr"
```
fabiolrdiniz, line 9: Do you know why the original file doesn't work? Please, notice that your added file (`test/testinput/gdas.mtiasi.t00z.20220216.bufr`) has 779 MB, it's too large. I'd revert it to use the original file or reduce the new one.

```suggestion
      obsdatain: "./testinput/gdas.t12z.mtiasi.tm00.bufr_d" 
```

fabiolrdiniz (commented):
Thanks for working on it, @smaticka. I suspect your code will have a global attribute with the `sensorCentralWavenumber` values. However, I believe what you are looking for is a `MetaData` channel-dimensioned variable with those values (`MetaData/sensorCentralWavenumber(Channel)`).

@rmclaren, is there a way of passing a float vector straight to an IODA variable? We know she can do that through a filter when using this file (UFO level), but I believe people would like to have it added at this stage. The solution via YAML at runtime opens the door to have different `sensorCentralWavenumber` though (which was a concern above).

test/testinput/bufr_ncep_mtiasi.yaml
PatNichols: @fabiolrdiniz this is my lack of knowledge showing but does the (*).bufr_d differ from (?).bufr in some way. Why do we have two diferent suffixes for a bufr file ( prep bufr vs. ncep bufr?)

BenjaminRuston:
@smaticka this likely will be moved to future work as updates to the converter are needed to make this happen.  Should plan this either with Ron as part for bufr2ioda.x,,, or with SteveH so yaml can be used to write MetaData

BenjaminRuston:
in short we may want to close this

smaticka:
Closing to reopen later once 2d ability is allowed for MetaData

BenjaminRuston:
@smaticka yes this is so confusing, but yes this ioda-converter PR can likely close again

we'll use the table of wavenumbers in a ufo instrumentTests ctest with a reduced ioda file

PR #1352, opened 2023-09-01
Refactor the bottle neck in combining the geovals

danholdaway:
## Description

The `combine_obsspace` code for contains a nasty bottle neck when combining GeoVaLs, where there is a rather inefficient nested loop. For GMAO when converting all the of the observations from GSI to IODA format and including the GeoVaLs it was taking around 50 minutes. After this change it's more like 8 minutes.

### DIFF
--- a/src/gsi_ncdiag/combine_obsspace.py
+++ b/src/gsi_ncdiag/combine_obsspace.py
@@ -276,21 +276,53 @@ def combine_obsspace(FileList, OutFile, GeoDir):
         GeoVarData31 = np.dstack(GeoVarData31)
         GeoVarIdx31 = np.dstack(GeoVarIdx31)
         GeoVarUnique31 = np.ones((len(idx), len(GeoVarData31[0]), len(GeoVarData31[0, 0, :])))*np.abs(nc.default_fillvals['f4'])
-        for ii, jj in np.ndindex(GeoVarData2.shape):
-            j = GeoVarIdx2[ii, jj]
-            i = inv[ii]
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):
Much cleaner, yes I never liked the loops and it's not surprising that this would be much more efficient. Thanks for this!

danholdaway:
> Much cleaner, yes I never liked the loops and it's not surprising that this would be much more efficient. Thanks for this!

Yes, thanks ChatGPT&copy;

src/gsi_ncdiag/combine_obsspace.py
```diff
@@ -276,21 +276,53 @@ def combine_obsspace(FileList, OutFile, GeoDir):
...
-                GeoVarUnique31[i, kk, j] = GeoVarData31[ii, kk, jj]
+
+        # Data2
+        # -----
+
+        # Create a mask to check for the conditions where GeoVarData2 is not equal to the fill values
+        non_fill_mask = (GeoVarData2 != nc.default_fillvals['i4']) & (GeoVarData2 != np.abs(nc.default_fillvals['f4']))
+
+        # Use np.where to find the indices that meet the condition
+        ii, jj = np.where(non_fill_mask)
```
BenjaminRuston, line 287: are you worried about putting a trap in here for when no valid data or at least all set to fill val?

BenjaminRuston (approved):
overall this looks fine and likely a bit faster, the numpy where also not blindingly fast

src/gsi_ncdiag/combine_obsspace.py
danholdaway: We could get a bit of speed up to avoid the `where` in those situations but I'm not esp. worried since we don't need the GeoVaLs beyond the short term

PR #1355, opened 2023-09-07
tropics converter update

BenjaminRuston:
## Description

this updates the TROPICS converter for L1B files v3 and maintains backward support for v2

## Issue(s) addressed

Resolves #1354 

## Checklist

- [x] I have performed a self-review of my own code

Issue #1354, opened 2023-09-06
TROPICS converter update for newer calibration


BenjaminRuston:
## Description

update the TROPICS converter for pathfinder data with new calibration

## Requirements

ioda python interface TROPICS data from NASA DISC 

## Dependencies
PR #1355 
PR  #1358

danholdaway:
The default is to build develop but any branch can be used.

PR #1358, opened 2023-09-12
Adding ctest for the updated TROPICS converter

smaticka:
## Description

The `ioda-converter` for `TROPICS` was updated to account for the new calibration by NASA. 
In this PR, a `ctest` is created to test that the converter properly works after building `jedi`. 

## Issue(s) addressed

Contributes to #1354 

## Dependencies

Ioda-converter update PR #1355 

## Checklist

- [x] reduce raw data so files aren't too large
- [x] add `test/testinput` file
- [x] add `test/testoutput` file
- [x] add `ctest` to the `test/CMakeList.txt` file
- [x] add converter to list which copies over `tropics_2ioda.py` into the `build/bin/` directory.
- [x] rebuild and confirm it passes.


test/CMakeLists.txt
```diff
@@ -96,6 +97,7 @@ list( APPEND test_input
 )
 
 list( APPEND test_output
+  testoutput/tropics_20220116.nc4
```
PatNichols, line 100: @smaticka  is this supposed to be tropics_20220216.nc??? The test is failing because it can't find a file.

smaticka: Yes I think that's it! Thanks @PatNichols

test/CMakeLists.txt
```diff
@@ -895,6 +897,18 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_sondetac
...
+                  ENVIRONMENT "PYTHONPATH=${IODACONV_PYTHONPATH}"
+                  COMMAND bash
+                  ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                          netcdf
+                  "${CMAKE_BINARY_DIR}/bin/tropics_2ioda.py
+                  -i testinput/tropics_20220216.nc
+                  -o testrun/tropics_20220216.nc4
+                  -d 2022021600"
+                  tropics_20220216.nc4 ${IODA_CONV_COMP_TOL})
+
```
huishao-r, line 911: Seems you used ".nc4" here. Should they be consistent with the testinput and output?

smaticka: I was using the original script from Ben I believe. It took in an nc file and output an nc4. Do you mean lines 907-908? Or do you mean 908 and 910?

huishao-r: I thought testoutput should be .nc,  which @PatNichols pointed out and you agreed?

BenjaminRuston: this all looks correct now,,, it's a bit confusing that the name of the input and the testoutput and testrun are all so similar, but if the testouput and testrun have some other convention maybe using `20220216T00Z_PT6H` or something make those two files look different

PatNichols: @smaticka @BenjaminRuston @huishao-r  The error I saw was that name of the file in the CMakefile did not match the one that was added in this PR. Hence the ctest could not find anything  to compare against.

PatNichols:
@smaticka The file not found error is gone. Now it is seeing differences between the testrun output file and the "gold" standard file in testoutput. 
DIFFER : LENGTHS OF RECORDS : Location (567) <> Location (2268)
DIFFER : LENGTHS : DIMENSION : Location : 567 <> 2268
DIFFER : VARIABLE : Location : ATTRIBUTE : suggested_chunk_dim : VALUES : 567 <> 2268
DIFFER : VARIABLE : Location : SIZE : 567 <> 2268
DIFFER : VARIABLE : dateTime : SIZE : 567 <> 2268
DIFFER : VARIABLE : latitude : SIZE : 567 <> 2268
DIFFER : VARIABLE : longitude : SIZE : 567 <> 2268
...more....

smaticka:
Thanks for the comments everyone. This test passes now! 
The `testoutput` file corresponded to a previous ioda-conversion. 

It should be read for review

PatNichols (approved):
thanks Sam

test/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@
 ################################################################################
 
 list( APPEND test_input
+  testinput/tropics_20220216.nc
```
fabiolrdiniz, line 11:
```suggestion
  testinput/tropics_2022021600.nc
```

fabiolrdiniz (approved):
Thanks, @smaticka! I'd add something informing the (approximate) hour to the name of the raw & converted files (I see that the recently added AMSR2 & GMI are lacking that). No need to hold the merge of this though... Thanks again!

huishao-r (approved):
The names are a little confusing. Other than that, it is ok.

BenjaminRuston (approved):
perfect passed with new ugly names hyphens and all

PR #1367, opened 2023-09-14
replacing test files with smaller data case

smaticka:
## Description

A new `ufo ctest` was made for `gmi`, but the `geovals` file needed for that test was too large. The raw data was reduced and converted and then run in `skylab` to generate smaller geovals files. In order to keep the workflow consistent using the `ioda ctest` `testoutput` file as the `testinput` file for the `ufo ctest`, the `ioda-converter` test files needed to be replaced.

The `iodaconv ctest` passes with the new files, by running `ctest -R test_iodaconv_gmi_gpm` in the `build` directory. 

## Issue(s) addressed
Resolves #<issue_number>

## Checklist

- [x] The `ctest` passes with new files


PatNichols (approved):
Thanls!

fcvdb (approved):
Thanks Sam!

PatNichols:
@smaticka Do you think we should take this out of draft status. It looks like it's ready to go.

PatNichols:
@smaticka Can we take this out draft status??

PR #1375, opened 2023-09-26
replacing tropics ctest files with smaller data case

smaticka:
A new `ufo ctest` was made for `tropics`, but the `geovals` file needed for that test was too large. The raw data was reduced and converted and then run in `skylab` to generate smaller geovals files. In order to keep the workflow consistent using the `ioda ctest` `testoutput` file as the `testinput` file for the `ufo ctest`, the `ioda-converter` test files needed to be replaced.

The `iodaconv ctest` passes with the new files, by running `ctest -R test_iodaconv_tropics -VV` in the `build` directory. 

## Issue(s) addressed
Resolves #<issue_number>

## Checklist

- [x] The `ctest` passes with new files

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -8,7 +8,7 @@
 ################################################################################
 
 list( APPEND test_input
-  testinput/TROPICS01.L1B.V03-04.ST20220216-012443.ET20220216-025955.nc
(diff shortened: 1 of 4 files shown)


PatNichols (approved):
Looks good

fcvdb (approved):
Thanks Sam!

PR #1379, opened 2023-09-28
Revert "Refactored QueryRunner::collectData"

danholdaway:
Reverts JCSDA-internal/ioda-converters#1221

This does not compile on Discover. I presume because we don't have the version of BUFR lib that's needed (`get_irf_f` not found). Fences need to be put in place in the CMake in case the user does not have a modern enough version of the dependent library before merging. 

We are using the exact version of spack-stack that is written in the documentation.


rmclaren:
The error `get_irf_f` would indicate that the NCEPLIB_bufr lib you are using is not the correct one. Must be 12.0.0 for that PR.

rmclaren:
@danholdaway Your suggested change sounds good to me.

danholdaway:
OK if you can make a PR with the minimum version you think it needed then we can close this and merge that instead.

danholdaway:
@rmclaren I think you want to make a new PR with just the change you just pushed rather than pushing it to this branch that reverts all the changes. And I think you want that change to be:

```cmake
ecbuild_find_package( bufr VERSION 12.0.0 QUIET )
```

CoryMartin-NOAA:
We should hold off on any code changes requiring spack-stack 1.5.0 until after we have funds appropriated for FY24, as it is likely work at both GMAO and EMC would be derailed until folks like Dan or myself can update our workflows to use the new stack. So by that I mean revert this PR for now.

PatNichols:
@CoryMartin-NOAA Understood. I will revert this.

PatNichols (approved):

CoryMartin-NOAA:
Will approve but need to remove the commit that @rmclaren added in here I think before it actually is a 'revert' PR.

srherbener (commented):
My preference would be to close this PR and go with #1379, but I'm fine not doing so if this isn't agreeable with everyone.

srherbener:
> We should hold off on any code changes requiring spack-stack 1.5.0 until after we have funds appropriated for FY24, as it is likely work at both GMAO and EMC would be derailed until folks like Dan or myself can update our workflows to use the new stack. So by that I mean revert this PR for now.

One thing to note is that spack-stack-1.4.1 (patch release) contains bufr@12.0.0 so that's an option, as long as spack-stack-1.4.1 is acceptable for moving through a potential government shutdown.

CMakeLists.txt
```diff
@@ -87,7 +87,7 @@ find_package( MPI )
 
 find_package( ioda QUIET )
 
-find_package( bufr QUIET )
+find_package( bufr 12 QUIET )
```
danholdaway, line 90:
```suggestion
find_package( bufr QUIET )
```

CoryMartin-NOAA:
I just chatted with @nicholasesposito and I think we are okay with knowing that ioda-converters may not be complete until we can update the default GDAS modules to 1.5.0 as long as 1) it builds without BUFR12 and 2) the instructions for using the spack stack modules for the Skylab release are sufficient. So basically, I think it's okay to not merge this and make the other change if that works for @danholdaway

danholdaway:
I reverted the single commit that ron made to this revert PR so we can go ahead with reverting things. Then Ron can revert this revert and re-add the commit he made to this revert pr. Make sense?

danholdaway:
Yes @CoryMartin-NOAA that is fine with me. We can merge https://github.com/JCSDA-internal/ioda-converters/pull/1381 and close this.

Issue #221, opened 2019-10-07
Missing variables in GSI ncdiag files cause inconsistent format in IODA-converted files


svahl991:
Our group has been performing MPAS model experiments using the 1-month of cycling data provided in the ncdiag files at http://data.jcsda.org/gsidiags'+yyyymmdd+'.tar.

We take these ncdiag files and use the ''proc_gsi_ncdiag.py" converter script followed by the "combine_conv.py" to create a IODA file containing all the radiosonde conventional obs (t, q, u, and v) for each time period, that we then use for our experiments.

We noticed that that for one single time period in the entire month of data we were missing the specific_humidity (q) variable in the converted & combined file. To be precise, the problem is not that there are missing values in in the specific_humidity variable, but that the variable is missing entirely from the file. This causes problems for our post-processing scripts that are equipped to handle missing values in a variable, but expect the variable to be present in the file and populated with missing values. 

The time period with the issue is the '6Z' window of April 30, 2018. After some detective work (described in detail below), we found that this situation came about because the source 'conv_q' ncdiag file for that time period contained no specific humidity information at all, and it appears to be the only time period in the entire month where this occurs.

Since this issue appears to be caused not by an error, but simply because of missing data, somewhere the scripts that process these data are going to need to handle this situation.

The question is which scripts should deal with it. Should our post-processing script need to test that all the same variables are present in all the converted/combined files? Or should the issue be dealt with "upstream", so that the files that come out of the convert and combine process are uniform in format? The latter seems more desirable, but maybe more difficult?

I'm interested to hear your thoughts @CoryMartin-NOAA about the difficulty of solving this issue in the scripts that perform the convert & combine process.

As mentioned earlier, I'm including below the text of an email I wrote which contains all the details on the research I've done on this issue, in case it is helpful for understanding:

=======================
JJ, Junmei,

I've been looking into the case of the missing specific humidity data in the 2018043006 time window of our 1-month data set. I think I have figured out what is happening.

I started my investigation by looking at the source GSI ncdiag files. I thought maybe one of the ncdiag files would be missing in the '06Z' directory. But that was not the case. All the same files were present in that directory as all the others.

[...]

PR #1394, opened 2023-10-09
Add option to save LSW in gnssro bufr2ioda

### LINKED ISSUE #1393: Add option to write out LSW in GNSSRO burf2ioda
## Description
LSW is given in the CDAAC processed WMO bufr files. Add the option to save this as a separate variable 
## Requirements

Code runs and outputs file with expected variables 

LSW is output correctly when using new flag. Output is unchanged without flag

## Dependencies
none

haydenlj:
## Description

Adds the option to save LSW from CDAAC processed bufr files as its own variable when a flag is set.

## Issue(s) addressed

Resolves #1393 

## Dependencies

none

## Impact

none

## Checklist

### DIFF
--- a/src/gnssro/gnssro_bufr2ioda.py
+++ b/src/gnssro/gnssro_bufr2ioda.py
@@ -288,7 +288,7 @@ def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None):
         bang[bangid] = float_missing_value
         bang_err[bangid] = float_missing_value
 
-        lsw = bang_err/bang * 100   # CDACC saves LSW in radians as bang_err. Convert to percentage 
+        lsw = bang_err/bang * 100   # CDACC saves LSW in radians as bang_err. Convert to percentage
         lsw[bang == float_missing_value] = float_missing_value
         obs_data[('localSpectralWidth', "MetaData")] = assign_values(lsw)
     return obs_data
--- a/src/gnssro/gnssro_bufr2ioda.py
+++ b/src/gnssro/gnssro_bufr2ioda.py
@@ -284,7 +284,7 @@ def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None):
             obs_data[k] = obs_data[k][good]
 
     if addLSW:
-        lsw = bang_err/bang * 100
+        lsw = bang_err/bang * 100   # CDACC saves LSW in radians as bang_err. Convert to percentage 
         obs_data[('localSpectralWidth', "MetaData")] = assign_values(lsw)
 
     return obs_data
--- a/src/gnssro/gnssro_bufr2ioda.py
+++ b/src/gnssro/gnssro_bufr2ioda.py
@@ -284,9 +284,10 @@ def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None):
             obs_data[k] = obs_data[k][good]
 
     if addLSW:
+        bang[bang == 0] = float_missing_value
+        bang_err[bang == 0] = float_missing_value
         lsw = bang_err/bang * 100   # CDACC saves LSW in radians as bang_err. Convert to percentage 
         obs_data[('localSpectralWidth', "MetaData")] = assign_values(lsw)
-
     return obs_data
 
 
--- a/src/gnssro/gnssro_bufr2ioda.py
+++ b/src/gnssro/gnssro_bufr2ioda.py
@@ -44,6 +44,7 @@ def main(args):
 
     dtg = datetime.strptime(args.date, '%Y%m%d%H')
     qc = args.qualitycontrol
+    addLSW = args.localspectralwidth
 
     # read / process files in parallel
     pool_input_01 = args.input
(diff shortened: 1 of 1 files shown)


src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -278,6 +283,11 @@ def get_obs_data(bufr, profile_meta_data, add_qc, record_number=None):
         for k in obs_data.keys():
             obs_data[k] = obs_data[k][good]
 
+    if addLSW:  
+        lsw = bang_err/bang * 100
+
+        obs_data[('localSpectralWidth', "ObsError")] = assign_values(lsw)
```
fabiolrdiniz, line 289:
```suggestion
        lsw = bang_err/bang * 100
        obs_data[('localSpectralWidth', "ObsError")] = assign_values(lsw)
```

fabiolrdiniz (approved):
Thanks, @haydenlj!

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -110,11 +111,15 @@ def main(args):
     VarAttrs[('longitude', 'MetaData')]['_FillValue'] = float_missing_value
     VarAttrs[('height', 'MetaData')]['_FillValue'] = float_missing_value
 
+    if addLSW:
+        VarAttrs[('localSpectralWidth', 'ObsError')]['units'] = 'Percentage'
```
hailingz, line 115: If I understand correctly, this will be localSpectralWidth@ObsError. That will be interpreted as the obs error of localSpectralWidth?

haydenlj: What would you say is the best location for this variable @hailingz? ObsValue? MetaData? @BenjaminRuston @huishao-r which is the best place for it?

huishao-r: This is a hard one to decide. LSW actually is considered as QC flags in the latest discussion of the new bufr table, instead of putting it under observation error section. What do we have under preQC currently?

fabiolrdiniz: In the IODA world, isn't it more like a `MetaData` variable (instead of `ObsVaue`, `ObsError`, `PreQC`)? With that saved as a `MetaData` variable, it can be used later in QC procedures to derive other quantities. If the `PreQC` route is chosen, it implies that a decision will be needed on deciding which classes are going to be considered as "good" (remember, `PreQC` is an integer-based variable). If it's kept as a float in the `MetaData` group, then the user is free to decide what is considered "good" later on when using the file.

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -278,6 +283,10 @@ def get_obs_data(bufr, profile_meta_data, add_qc, record_number=None):
         for k in obs_data.keys():
             obs_data[k] = obs_data[k][good]
 
+    if addLSW:  
+        lsw = bang_err/bang * 100
+        obs_data[('localSpectralWidth', "ObsError")] = assign_values(lsw)
```
huishao-r: Would this cause an issue if _bang_ is zero?

haydenlj: Thanks for pointing that out, I fixed it

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -417,5 +430,10 @@ def concat_obs_dict(obs_data, append_obs_data):
         help='turn on quality control georeality checks',
         default=False, action='store_true', required=False)
 
+    optional.add_argument(
+        '-l', '--localspectralwidth',
```
BenjaminRuston, line 434:
```suggestion
        '-lsw', '--localspectralwidth',
```

let's maybe have this be a slightly longer short name

BenjaminRuston (approved):
great this seems to be working for me as well tried on a spire file with and without the `-lsw` switch

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -278,6 +283,14 @@ def get_obs_data(bufr, profile_meta_data, add_qc, record_number=None):
         for k in obs_data.keys():
             obs_data[k] = obs_data[k][good]
 
+    if addLSW:
+        bangid = bang == 0
+        bang[bangid] = float_missing_value
+        bang_err[bangid] = float_missing_value
+
+        lsw = bang_err/bang * 100   # CDACC saves LSW in radians as bang_err. Convert to percentage
+        lsw[bang == float_missing_value] = float_missing_value
```
hailingz, line 292: This missing value is actually taken care in line #221.
but leaving as it is does not hurt.

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -278,6 +283,14 @@ def get_obs_data(bufr, profile_meta_data, add_qc, record_number=None):
...
+        obs_data[('localSpectralWidth', "MetaData")] = assign_values(lsw)
```
hailingz, line 293: Since this is normalized by observation and the unit is %, perhaps localSpectralWidth does not reflect what it is. Any name convention? @huishao-r  probably has a better idea.

hailingz (commented):
Another question but not closely related to this PR particularly: lines 203 and 235, it seems like the "_conf" variables are never used. Do you want keep that? probably this is a question to @BenjaminRuston . again, not added by this PR, does not need to address right away.

Issue #1378, opened 2023-09-26
[Bug] Test failures on Mac M1 (arm64 architecture)


srherbener:
## Current behavior (describe the bug)

I'm seeing the following test failures on my Mac M1 when using the develop branch:
```
  1 706:test_iodaconv_bufr_ncep_1bamua2ioda
  2 707:test_iodaconv_bufr_ncep_1bamua2ioda_n15
  3 708:test_iodaconv_bufr_ncep_esamua2ioda
  4 712:test_iodaconv_prepbufr_ncep_api_adpsfc2ioda
  5 713:test_iodaconv_prepbufr_ncep_api_sfcshp2ioda
  6 725:test_iodaconv_bufr_ncep_mtiasi
  7 726:test_iodaconv_bufr_ncep_atms
  8 732:test_iodaconv_bufr_ncep_prepbufr_adpupa_api
  9 753:test_iodaconv_prepbufr_conv
 10 754:test_iodaconv_mhs_conv
 11 755:test_iodaconv_amsua_conv
```

These are a mix of what looks like precision/tolerance issues, fill values and shared library load issues:
```
30143 DIFFER : VARIABLE : sensorViewAngle : POSITION : [818] : VALUES : 31.659 <> 31.659
30144 DIFFER : VARIABLE : sensorViewAngle : POSITION : [834] : VALUES : -15.003 <> -15.003
30145 DIFFER : VARIABLE : sensorViewAngle : POSITION : [835] : VALUES : -11.67 <> -11.67
30146 DIFFER : VARIABLE : sensorViewAngle : POSITION : [836] : VALUES : -8.337 <> -8.337
30147 DIFFER : VARIABLE : sensorViewAngle : POSITION : [837] : VALUES : -5.004 <> -5.004
30148 DIFFER : VARIABLE : sensorViewAngle : POSITION : [838] : VALUES : -1.671 <> -1.671
30149 DIFFER : VARIABLE : sensorViewAngle : POSITION : [839] : VALUES : 1.662 <> 1.662
30150 DIFFER : VARIABLE : sensorViewAngle : POSITION : [843] : VALUES : 14.994 <> 14.994
30151 DIFFER : VARIABLE : sensorViewAngle : POSITION : [844] : VALUES : 18.327 <> 18.327
30152 DIFFER : VARIABLE : sensorViewAngle : POSITION : [845] : VALUES : 21.66 <> 21.66
[...]

PR #1397, opened 2023-10-12
Fixed overflow issues with int64 fill value handling

srherbener:
## Description

This PR fixes ctests that exhibited overflow errors when handling int64 fill values. The typical way this shows up is when converting the date time values in the time offset format (float32) into date time values in the epoch format (int64).

This involves an update of test reference files for three tests:

- test_iodaconv_prepbufr_ncep_api_adpsfc2ioda
- test_iodaconv_prepbufr_ncep_api_sfcshp2ioda
- test_iodaconv_bufr_ncep_prepbufr_adpupa_api

I verified that the only thing that changed in the test reference files was the values of the fill values used for dateTime and releaseTime variables (which are stored in the epoch format in int64 data type). Also verified that the fill value spec in the test reference files for dateTime and releaseTime are consistent with the actual values in the variable data (so the ioda reader will be able to recognize these as missing).

## Issue(s) addressed

Partially addresses #1378

## Dependencies

None

## Impact

The only impact is for cases where the actual value of missing values in dateTime  and releaseTime matter. The expectation is that this means no impact.

## Checklist

- [x] I have made corresponding changes to the documentation (code comments)
- [x] Test on HPC platform - Orion, GNU, OpenMPI

Note that differences between Orion and the Mac M1 were what triggered looking into this, so it would be good to verify that both of these get the same results.


test/testinput/bufr_ncep_prepbufr_adpupa.py
```diff
@@ -67,10 +68,19 @@ def test_bufr_to_ioda(DATA_PATH, OUTPUT_PATH, date):
...
     # Time variable
+    # The time is entering in as a float32 value representing an offset from the cycle
+    # time in hours (so fractions of hours can exist). This needs to be first converted
+    # to seconds while still a float32, then converted to an int64 for the dateTime variable.
+    # Another consideration is that the get function returns a masked array with an
+    # appropriate fill value assigned. When converting to an int64, the fill value needs
+    # to get updated, and then before writing into the output ioda file, the masked array
+    # function filled() needs to be called which will convert the values marked invalid
+    # to the fill value.
     hrdr = r.get('timeOffset')
```
rmclaren, line 79: delete this line

rmclaren (approved):
@srherbener Thanks so much for fixing these!

test/testinput/bufr_ncep_prepbufr_adpupa.py
srherbener: Thanks for catching this. I've removed the line.

nicholasesposito (approved):
Looks fine as a fix to the v1 scripts. For the record, there are v2 scripts that I don't believe have the same issue. They're in branches on GDASApp but I don't know if there is a plan to publish them at the moment.

PatNichols (approved):
Thanks @srherbener !

PR #1408, opened 2023-10-26
generic gnssro ncepbufr2ioda converter 

### LINKED ISSUE #1409: adding an generic gnssro converter: gdas bufr to ioda
## Description

The current Fortran converter in develop that converts a NCEP (or gdas) bufr file to IODA was written following mostly (if not completely) the read_gps.f90 routine in https://github.com/NOAA-EMC/GSI/blob/develop/src/gsi/read_gps.f90

A more generic converter is needed so the observations can be compared to those generated by other centers, like the ODB data Met Office uses. Specifically, ODB reads/uses quality flags which is done differently in the current converter (or read_gps.f90). 

We probably want to keep both for now  for different applications and consider consolidate a bit later.

#https://github.com/JCSDA-internal/ioda-converters/pull/1408 is merged

## Dependencies
N/A

hailingz:
## Description

This is a revised version of  the NCEP GNSSRO Bufr2ioda converter after discussion with  @neill-b 
1. To make the Bufr2ioda converter more generic, this PR adds two variables, "qualityFlags" and "timeOffset", to metadata. 
- qualityFlags is an integer adding up the 16-digit QC flag reported in the Bufr data. details are referred to the ROM SAF's documentation. It contains a new of flags, such as bending angle non-nominal, refractivity non-monial, descending flag, L2C GPS signal, etc, for sanity checks.
- timeOffset is the time increment relative to the start time of the occultation in second. That can be used when time interpolation is involved.

2. Meantime, this PR keeps some basic sanity checks (such as lat/lon range) and removes all others, and removes the obs error specification as well. Both the removed sanity checks and obs error assignment can be done via UFO filters.

3. "second" of observation time was set to zero before and now is read from file. "dateTime" is consistent with Met Office's ODB data now.

4. README file in src/gnssro was updated. Simple descriptions were provided for the FIVE GNSSRO converters in this folder. Please edit for the python one if you want @BenjaminRuston  

To test, one has to build the ioda-bundle with this branch: [feature/gnssro_metoffice_generic](https://github.com/JCSDA-internal/ioda-converters/tree/feature/gnssro_metoffice_generic)

The ctest is:
         ctest -R gnssro_bufr_generic_conv

Rather you can look at (ncdump) the output file in test/testoutput/gnssro_obs_3prof_generic_2022090100.nc4

NOTE: Neill and I compared the obs accounts for each QC categories by missions. The attached tables compare those numbers among the ODB data (LEFT column), the NC file  (MIDDLE) created by running  gnssro_bufr2ioda_generic.x and the file (RIGHT) created by running tgnssro_bufr2ioda.x 
The new generic converters generated identical observations with the ODB file!

On the other hand,  "gnssro_bufr2ioda.x", which applies more sanity checks and does not use second report in consistence with GSI, can be probably kept in this repository. 

Screenshot 2023-10-30 at 3 41 13 PM
Screenshot 2023-10-30 at 3 41 03 PM

## Issue(s) addressed

Resolves #1409

## Dependencies
N/A

## Impact
N/A

## Checklist


src/gnssro/gnssro_bufr2ioda_generic.f90
```diff
@@ -0,0 +1,496 @@
...
+   call check(nf90_def_var_deflate(grpid_metadata, varid_asce,       &
+                                    & shuffle=1, deflate=1, deflate_level=deflate_level))
+   call check(nf90_put_att(grpid_metadata, varid_asce, "longname", "the original occultation ascending/descending flag"))
+   call check(nf90_put_att(grpid_metadata, varid_asce, "valid_range", int((/0, 1/))))
+   call check(nf90_put_att(grpid_metadata, varid_asce, "flag_values", int((/0, 1/))))
+   call check(nf90_put_att(grpid_metadata, varid_asce, "flag_meanings", "descending ascending"))
+   call check(nf90_put_att(grpid_metadata, varid_asce, "units", "1"))
+   call check(nf90_def_var_fill(grpid_metadata, varid_asce, 0, i_missing))
+
+   call check(nf90_def_var(grpid_metadata, "process_center", NF90_INT, nlocs_dimid, varid_ogce))
```
neill-b, line 335: I believe the name should be `dataProviderOrigin` rather than `process_center`.

src/gnssro/gnssro_bufr2ioda_generic.f90
```diff
@@ -0,0 +1,496 @@
...
+
+   call check(nf90_def_var(grpid_metadata, "satelliteIdentifier", NF90_INT, nlocs_dimid, varid_said))
+   call check(nf90_def_var_fill(grpid_metadata, varid_said, 0, i_missing))
+   call check(nf90_def_var_deflate(grpid_metadata, varid_said,       &
+                                    & shuffle=1, deflate=1, deflate_level=deflate_level))
+   call check(nf90_put_att(grpid_metadata, varid_said, "longname", &
+      & "Low Earth Orbit satellite identifier, e.g., COSMIC2=750-755"))
+   call check(nf90_put_att(grpid_metadata, varid_said, "units", "1"))
+
+   call check(nf90_def_var(grpid_metadata, "occulting_sat_is", NF90_INT, nlocs_dimid, varid_siid))
```
neill-b, line 313: I think this should be named `instrumentIdentifier`.

neill-b (commented):
I have a couple of (very) minor comments on the code.  However, comparing the generic reference file to the "normal" reference file there are some bigger differences.  The lat/lon for the first observation is set to 1E11, as are a few other observations, and the sequence number appears to have an offset (the second and third profiles start at the wrong point).

src/gnssro/gnssro_bufr2ioda_generic.f90
hailingz: Thanks for the catch. 
Making notes here: this also need to be reflected in "ioda-upgrade-v2-to-v3"
by ioda/share//ioda/yaml/validation/ObsSpace.yaml

src/gnssro/gnssro_bufr2ioda_generic.f90
hailingz: Done! thanks!

hailingz:
"process_center (dataProviderOrigin)" and "occulting_sat_is (instrumentIdentifier)" are not listed in ioda/share//ioda/yaml/validation/ObsSpace.yaml 
That is why they could not be upgraded with "ioda-upgrade-v2-to-v3.x" 

I do not think the old version of GNSSRO Fortran converter is used much so probably no changes need to be done in ObsSpace.yaml. Though a few of ctests still rely on the IODA upgrader but definitely neither "dataProviderOrigin" nor "instrumentIdentifier" is used in any of those tests.

I just make notes here @srherbener

neill-b:
> "process_center (dataProviderOrigin)" and "occulting_sat_is (instrumentIdentifier)" are not listed in ioda/share//ioda/yaml/validation/ObsSpace.yaml That is why they could not be upgraded with "ioda-upgrade-v2-to-v3.x"

I think the naming-conventions sprint looked hard for old variable names in ufo test reference files.  So, perhaps `process_center` and `occulting_sat_is` were missed because they were in a repo which wasn't looked at.  This **might** mean that there are other files present with unusual variable names.

huishao-r (commented):
@hailingz why do we want to have two converters for bufr2ioda? Are you planning to remove the old one? Also, I would prefer to keep sanity check and obs error assignment as optional (same as the current python converter I think), if we are going to remove the older converter (I wouldn't like to keep both). The obs error part is due to the fact we need obsfunction to allow certain obs errors in (including the new hybrid 3CH Lynsey is working on). The fast tract to allow it in is through converter (but we do want to change it to UFO side later).

neill-b:
Hi @huishao-r - removing the sanity checks was something that I requested.  In the ODB reader (as far as possible) we avoid checks, and try and put all the QC into the ufo filters.  So, for comparison we'd like the converter to behave in a similar way.  The easiest way to achieve this is to create a separate file and remove the additional QC checks from that code.  The additional checks could be put as optional with some extra work from @hailingz, but it would make our life easier to avoid that work.

srherbener:
> "process_center (dataProviderOrigin)" and "occulting_sat_is (instrumentIdentifier)" are not listed in ioda/share//ioda/yaml/validation/ObsSpace.yaml That is why they could not be upgraded with "ioda-upgrade-v2-to-v3.x"
> 
> I do not think the old version of GNSSRO Fortran converter is used much so probably no changes need to be done in ObsSpace.yaml. Though a few of ctests still rely on the IODA upgrader but definitely neither "dataProviderOrigin" nor "instrumentIdentifier" is used in any of those tests.
> 
> I just make notes here @srherbener

@hailingz thank you for the notes. I think in the long run the ObsSpace.yaml file should be updated so that the files can be properly validated. I think can be done in a subsequent PR however so it would be good to create a ZenHub issue for this. Thanks!

srherbener (commented):
I think in the long run it would be good to settle on one converter since the two converters (current converter and the new generic converter) are close in functionality. Would it work to proceed with the two converters until the sanity checks, etc. have been ported to UFO filters, then remove the old converter? Would this cover @huishao-r needs to have the sanity checks, etc. available for example? Thanks!

huishao-r:
@neill-b @srherbener Thanks for the comments. This comes into an ongoing discussion on how to reduce the use of obsSpace and memory. Doing it in the converter is not optimal. But I don't know where to put them yet, to be honest.

hailingz:
> oceed with the two converters until the sanity checks, etc. have been ported to UFO filters, then remove the old converter? Would t

Hi @srherbener Thank you for your comments. 
I agree with you that we can go with two converters now and look for merging them to one once we have more ideas. 
The main issue is that different centers read/write data differently in terms of the checks. Another thing in this PR is that it reads "SECOND" from the bufr report while the old one following GSI ignores the SECOND by setting it to zero in converter. It might not be scientifically important but can be center dependent.

@huishao-r 
We may want to talk it?

neill-b (approved):
I'm happy with these changes - I have checked that the observations from the converter are identical to what we get via the ODB reader (starting with a Met Office ODB file).  It would make sense for @hailingz and @huishao-r to open an issue to look at migrating towards the generic converter using ufo filters.  The issue of reducing the memory requirement is a much more pressing issue for hyper-spectral radiances, so we can follow what is developed for them.

BenjaminRuston:
thanks @hailingz and @neill-b we should choose which is the default (NOAA?) and then add a switch to do the changes in your steps 1. and 2. above, the fix to `seconds` to the `dateTime` should be kept regardless

but have one routine that can output either style ODB or NOAA

also good time to add a usage clause, so call without arguments and it spits out usage clause rather than having to read the code.. I can help with this.

hailingz:
> a usage clause

Hello @BenjaminRuston Thank you so much for your very useful comments.
I have just pushed a couple commits which added (1) "a usage clause", and  (2) optional argument so one can choose apply more sanity checks or not. Also descriptions are updated in the top of the codes.  
Could you please take a look?

PatNichols:
@hailingz is it possible to combine the generic and other into one file?  In some sense it looks like we are repeating ourselves. thanks

hailingz:
> @hailingz is it possible to combine the generic and other into one file? In some sense it looks like we are repeating ourselves. thanks

Yes @PatNichols 
The old one was removed and the generic one is renamed to gnssro_bufr2ioda.f90
README is updated.

hailingz:
Thank you @BenjaminRuston for the "ready to merge" label. Please let me know if anything is still needed so we can get this merged soon. @PatNichols

BenjaminRuston:
@hailingz I tried this converter the resulting file is much smaller than before these changes

is this expected looks like the gsi refractivity and bending angle errors were removed

BenjaminRuston:
@hailingz here results through skylab for the `new` file versus what we have archived using the previous version of the decoder 

image

if this is expected or you're happy let's merge this in

hailingz:
Thank you @BenjaminRuston for the plots. The new plot should be identical to the old one.
I have added another check in the "addChecks". Could you please check the results again?
Thank you so much!

BenjaminRuston (approved):
yes still working thanks for getting this in

PR #1411, opened 2023-10-31
ncdiag radiance bias addition to ioda output

### LINKED ISSUE #1410: gsi_ncdiag add total bias to output
## Description
the current `gsi_ncdiag.py` does not seem to add `GsiBc` in the ioda output
compute this using  `Obs_Minus_Forecast_adjusted` and `Obs_Minus_Forecast_unadjusted` with missing if these do not exist in the input diag* ncDiag files

## Requirements

ncDiag files from GSI

compute the total Obs bias and put into output IODA file

BenjaminRuston:
## Description

This adds the `GsiBc` to the IODA output from ncDiag

## Issue(s) addressed

Resolves #1410 

## Checklist

### DIFF
--- a/src/gsi_ncdiag/gsi_ncdiag.py
+++ b/src/gsi_ncdiag/gsi_ncdiag.py
@@ -1349,13 +1349,13 @@ class Radiances(BaseGSI):
                 pass
         ncout.close()
 
-    def toIODAobs(self, OutDir, ObsBias, QCVars, TestRefs, clobber=True):
+    def toIODAobs(self, OutDir, ObsBias, TotalBias, QCVars, TestRefs, clobber=True):
         """ toIODAobs(OutDir,clobber=True)
         output observations from the specified GSI diag file
         to the JEDI/IODA observation format
         """
 
-        print("Input Parameters: ObsBias=%s QCVars=%s TestRefs=%s" % (ObsBias, QCVars, TestRefs))
+        print("Input Parameters: ObsBias=%s TotalBias=%s QCVars=%s TestRefs=%s" % (ObsBias, TotalBias, QCVars, TestRefs))
         # set up a NcWriter class
         outname = OutDir + '/' + self.sensor + '_' + self.satellite + \
             '_obs_' + self.validtime.strftime("%Y%m%d%H") + '.nc4'
@@ -1593,14 +1593,15 @@ class Radiances(BaseGSI):
                 outdata[gvname] = np.reshape(tmp, (nlocs, nchans))
                 VarDims[gvname] = ['Location', 'Channel']
 
-        # brute force for now add option hook and default to missing
-        key1 = 'Obs_Minus_Forecast_adjusted'
-        key2 = 'Obs_Minus_Forecast_unadjusted'
-        tmp = self.var(key1) - self.var(key2)
-        iodavar = 'GsiBc'
-        gvname = "brightnessTemperature", iodavar
-        outdata[gvname] = np.reshape(tmp, (nlocs, nchans))
-        VarDims[gvname] = ['Location', 'Channel']
+        if (TotalBias):
+            # compute final bias correction
+            key1 = 'Obs_Minus_Forecast_adjusted'
+            key2 = 'Obs_Minus_Forecast_unadjusted'
+            tmp = self.var(key1) - self.var(key2)
+            iodavar = 'GsiBc'
+            gvname = "brightnessTemperature", iodavar
+            outdata[gvname] = np.reshape(tmp, (nlocs, nchans))
+            VarDims[gvname] = ['Location', 'Channel']
 
         # brightness temperature variables
         value = 'brightnessTemperature'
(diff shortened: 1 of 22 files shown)


src/gsi_ncdiag/proc_gsi_ncdiag.py
```diff
@@ -92,11 +92,13 @@ def run_radiances_obsdiag(radfile, outdir):
                 help="Path to directory to output observations")
 ap.add_argument("-d", "--obsdiag_dir",
                 help="Path to directory to output observations")
-ap.add_argument("-b", "--add_obsbias", default=False,
+ap.add_argument("-b", "--add_obsbias", default=False, action='store_true',
```
BenjaminRuston, line 95: if you didn't already know, @danholdaway and @CoryMartin-NOAA this changes the arguments so instead of passing:
`-b True`

you would now pass:
`-b`

without it still defaults to false presence of the switch turns it to true. This may interrupt scripting you have in place so let me know or suggest remove of the `action='store_true'` and we can rollback

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -1593,6 +1593,16 @@ def toIODAobs(self, OutDir, ObsBias, QCVars, TestRefs, clobber=True):
                 outdata[gvname] = np.reshape(tmp, (nlocs, nchans))
                 VarDims[gvname] = ['Location', 'Channel']
 
+        if (TotalBias):
```
CoryMartin-NOAA, line 1596: Just noting that you may wish to do this for aircraft T as well as Ps, both of which have bias corrections applied.

src/gsi_ncdiag/ncdiag_to_feedback.py
```diff
@@ -0,0 +1,101 @@
...
+    # open analysis departure
+    g = h5py.File(oman, 'r')
+    # duplicate (destructive) the analysis departure
+    shutil.copyfile(oman, output_file)
+    # open output ioda in read write
+    k = h5py.File(output_file, 'r+')
+
+    # start appending data to output_file
+    f.copy(f['GsiHofXBc'], k, 'hofx0')
+    g.copy(g['GsiHofXBc'], k, 'hofx1')
```
CoryMartin-NOAA, line 31: This should be hofx2 to be "correct", given that it is after the second (final) minimization loop

CoryMartin-NOAA (approved):

src/gsi_ncdiag/gsi_ncdiag.py
BenjaminRuston: thanks Cory, think I will just add to all the class definitions

src/gsi_ncdiag/ncdiag_to_feedback.py
BenjaminRuston: noted

danholdaway:
Thanks for flagging this @BenjaminRuston. I have the corresponding change on the Swell side so from my perspective feel free to merge when ready. Does this replace the way we currently produce the bias correction coeffs using the text files from GSI?

CoryMartin-NOAA:
@danholdaway I'm under the impression this is the bias itself and not the coeffs, which would still be the other converters (which are changing as part of the VarBC generalization)

danholdaway (approved):

danholdaway:
OK, thanks @CoryMartin-NOAA. Will the generalization changes be backward compatible?

smaticka (approved):
This works for me. Thanks @BenjaminRuston !

I added a description to the definition in ncdiag_to_feedback file. Please correct it if there are any mistakes.

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -1765,18 +1792,34 @@ def toGeovals(self, OutDir, clobber=True):
...
+                    key1 = 'Obs_Minus_Forecast_adjusted'
+                    key2 = 'Obs_Minus_Forecast_unadjusted'
+                    if key1 in self.df.variables.keys() and key2 in self.df.variables.keys():
+                        tmp = self.var(key1) - self.var(key2)
+                        iodavar = 'GsiBc'
+                        gvname = vname, iodavar
+                        outdata[gvname] = np.reshape(tmp, dim)
+                        VarDims[gvname] = dim_names
+                    else:
+                        print(f' ... can not add total bias missing {key1} or {key2} from input file')
```
BenjaminRuston, line 1816: @danholdaway and @CoryMartin-NOAA wasn't having luck running Ozone with ncDiag:

```
Traceback (most recent call last):
  File "/work2/noaa/jcsda/bruston/work/JCSDA/JEDI/ioda-bundle-gnu/iodaconv/src/gsi_ncdiag/gsi_ncdiag.py", line 1862, in toIODAobs
    tmp = self.var('Input_Observation_Error')
```

CoryMartin-NOAA: I don't think we've ever tried the operations diags for ozone through the converter, it's possible there is a slightly different variable name that needs to be changed.

PR #1414, opened 2023-11-09
Pass TotalBias argument to Ozone.toGeovals

danholdaway:
## Description

Our nightly picked up a bug in the Ozone converter code. TotalBias is an undefined variable.

### DIFF
--- a/src/gsi_ncdiag/gsi_ncdiag.py
+++ b/src/gsi_ncdiag/gsi_ncdiag.py
@@ -1745,7 +1745,7 @@ class Ozone(BaseGSI):
         self.nobs = len(df.dimensions['nobs'])
         self.df = df
 
-    def toGeovals(self, OutDir, clobber=True):
+    def toGeovals(self, OutDir, TotalBias=False, clobber=True):
         """ toGeovals(OutDir,clobber=True)
         if model state fields are in the GSI diag file, create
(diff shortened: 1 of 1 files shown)


BenjaminRuston (approved):
yes thought something like this may happen,,, thanks @danholdaway looks good

CoryMartin-NOAA (approved):

BenjaminRuston:
Wait,, think the correct approach is none of the `toGeovals` should have `TotalBias`

so we should remove this section:

```
                if (TotalBias):
                    # compute final bias correction
                    key1 = 'Obs_Minus_Forecast_adjusted'
                    key2 = 'Obs_Minus_Forecast_unadjusted'
                    if key1 in self.df.variables.keys() and key2 in self.df.variables.keys():
                        tmp = self.var(key1) - self.var(key2)
                        iodavar = 'GsiBc'
                        gvname = vname, iodavar
                        outdata[gvname] = np.reshape(tmp, dim)
                        VarDims[gvname] = dim_names
                    else:
                        print(f' ... can not add total bias missing {key1} or {key2} from input file')
```

BenjaminRuston:
@danholdaway please put `ready to merge` if this fix works for you and we'll get it in

danholdaway:
Thanks @BenjaminRuston. Removing it makes sense to me.

PR #1430, opened 2023-12-08
 option to retrieve only bending angle also add sub sat id

### LINKED ISSUE #1429: update gnssro converter to new BUFR template including satellite sub id
## Description

the WMO BUFR templates have been updated allowing for the inclusion of the satellite sub identifier.

when using the ecCodes interface this is index using: `satelliteSubIdentifier`

further add an option to only retrieve bending angle ignoring `refractivity` and retrieved profiles

successful creation of IODA file for UFO

## Dependencies
ecCodes 2.32.1 or greater

BenjaminRuston:
add an option to the call to allow for bending angle only to be retrieved from the input BUFR file

also leveraging the new WMO BUFR template retrieve the sub satellite ID
t fix? What new capability does it add?

## Issue(s) addressed

Resolves #1429 

## Dependencies

ecCodes 

## Checklist

### DIFF
--- a/src/gnssro/gnssro_bufr2ioda.py
+++ b/src/gnssro/gnssro_bufr2ioda.py
@@ -45,6 +45,7 @@ def main(args):
     dtg = datetime.strptime(args.date, '%Y%m%d%H')
     qc = args.qualitycontrol
     addLSW = args.localspectralwidth
+    only_bang = args.onlybendingangle
 
     # read / process files in parallel
     pool_input_01 = args.input
@@ -53,7 +54,7 @@ def main(args):
     obs_data = {}
     # create a thread pool
     with ProcessPoolExecutor(max_workers=args.threads) as executor:
-        for file_obs_data in executor.map(read_input, pool_inputs, repeat(qc), repeat(addLSW)):
+        for file_obs_data in executor.map(read_input, pool_inputs, repeat(qc), repeat(addLSW), repeat(only_bang)):
             if not file_obs_data:
                 print(f"INFO: non-nominal file skipping")
                 continue
@@ -119,7 +120,7 @@ def main(args):
     writer.BuildIoda(obs_data, VarDims, VarAttrs, GlobalAttrs)
 
 
-def read_input(input_file_and_record, add_qc, addLSW):
+def read_input(input_file_and_record, add_qc, addLSW, only_bang):
     """
     Reads/converts input file(s)
 
@@ -140,8 +141,10 @@ def read_input(input_file_and_record, add_qc, addLSW):
     codes_set(bufr, 'unpack', 1)
 
     profile_meta_data = get_meta_data(bufr)
+    if not profile_meta_data:
+        return None
 
-    obs_data = get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=record_number)
+    obs_data = get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=record_number, only_bang=only_bang)
 
     return obs_data
 
@@ -163,7 +166,7 @@ def get_meta_data(bufr):
     hour = codes_get(bufr, 'hour')
     minute = codes_get(bufr, 'minute')
     second = codes_get(bufr, 'second')  # non-integer value
-    second = round(second)
+    second = np.floor(second)
 
     # get string date, translate to a datetime object, then offset from epoch
     dtg = ("%4i-%.2i-%.2iT%.2i:%.2i:%.2iZ" % (year, month, day, hour, minute, second))
@@ -174,7 +177,7 @@ def get_meta_data(bufr):
     return profile_meta_data
 
 
-def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None):
+def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None, only_bang=False):
 
     # allocate space for output depending on which variables are to be saved
     obs_data = {}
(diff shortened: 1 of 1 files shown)


src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -320,6 +331,7 @@ def def_meta_data():
         # "timeIncrement": 'timeIncrement',
         "earthRadiusCurvature": 'earthLocalRadiusOfCurvature',
         "satelliteIdentifier": 'satelliteIdentifier',
+        "satelliteSubIdentifier": 'satelliteSubIdentifier',
```
BenjaminRuston, line 334: the addition of this line will make this converter **not** backward compatible
will need to add handling at line:
https://github.com/JCSDA-internal/ioda-converters/blob/40531cd029207f88a7c015b78736e5904079bb66/src/gnssro/gnssro_bufr2ioda.py#L160

BenjaminRuston: have a way to handle this now

BenjaminRuston:
@huishao-r these changes are backward compatible for `satelliteSubIdentification` so earlier files that don't have it will still decode.

also this adds the `--only_bending_angle` option if one does not want `refractivity` and the `profiles`

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -227,12 +233,15 @@ def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None):
     obs_data[('bendingAngle', "PreQC")] = np.full(krepfac[0], 0, dtype=ioda_int_type)
 
     # (geometric) height is read as integer but expected as float in output
-    height = codes_get_array(bufr, 'height', ktype=float)
-
-    # get the refractivity
-    refrac = codes_get_array(bufr, 'atmosphericRefractivity')[0::2]
-    refrac_err = codes_get_array(bufr, 'atmosphericRefractivity')[1::2]
-    refrac_conf = codes_get_array(bufr, 'percentConfidence')[sum(krepfac[:1])+1:sum(krepfac[:2])+1]
+    if only_bang:
```
fcvdb, line 236: The refractivity is overwritten with the bending angle. This could be confusing later on. Could we fill all the `refrac` array with missing values instead?

BenjaminRuston: good idea,,,, or exclude it entirely is the other option and my preference,, do not know if that works 

missing may be a initial step

huishao-r:
> When I ran this, I got failure. The old version is still working to me. I'll try to reinstall my code and see whether there is some other updates causing this.

I pushed a code change to allow the height to read from impact height for only_bang option. It works for me, instead of the original PR code. Will see whether CI tests pass or not.

PatNichols:
@huishao-r The same ctest is still failing and there is a coding norms failure:

14 - iodaconv_gnssro_coding_norms (Failed)
130 - test_iodaconv_generic_gnssro_bufr (Failed)

huishao-r:
> @huishao-r The same ctest is still failing and there is a coding norms failure:
> 
> 14 - iodaconv_gnssro_coding_norms (Failed) 130 - test_iodaconv_generic_gnssro_bufr (Failed)

Got it. I am running ctest locally so that I can see what went wrong. Will push changes once I get the fix.

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -227,12 +235,20 @@ def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None):
...
+        refrac_err = codes_get_array(bufr, 'atmosphericRefractivity')[1::2]
 
-    # get the refractivity
-    refrac = codes_get_array(bufr, 'atmosphericRefractivity')[0::2]
-    refrac_err = codes_get_array(bufr, 'atmosphericRefractivity')[1::2]
-    refrac_conf = codes_get_array(bufr, 'percentConfidence')[sum(krepfac[:1])+1:sum(krepfac[:2])+1]
+    try:
+        height = codes_get_array(bufr, 'geopotentialHeight', ktype=float)
+    except Exception as e:
+        height = codes_get_array(bufr, 'height', ktype=float)
```
huishao-r, line 252: height = codes_get_array(bufr, 'geopotentialHeight', ktype=float)

This line caused the failure of the ctest failure: test_iodaconv_generic_gnssro_bufr
with the PR code, _height_ becomes _geopotentialHeight_, which differs from the reference height. We need to update the reference or figure out how to deal with a data set with both GH and height.

src/gnssro/gnssro_bufr2ioda.py
```diff
@@ -200,6 +206,8 @@ def get_obs_data(bufr, profile_meta_data, add_qc, addLSW, record_number=None):
     impact = codes_get_array(bufr, 'impactParameter')[offset::drepfac[0]]
     bang = codes_get_array(bufr, 'bendingAngle')[offset*2::drepfac[0]*2]
     bang_err = codes_get_array(bufr, 'bendingAngle')[offset*2+1::drepfac[0]*2]
+    if only_bang:
+        bang_err[:] = 0.003
```
huishao-r, line 210: I tested the decoded data and found that missing bang_error doesn't cause an issue while running H(x). So this line seems not necessary.

src/gnssro/gnssro_bufr2ioda.py
PatNichols: @huishao-r Seems we should keep both or get rid of one of them. Correct?

PatNichols: @huishao-r The code as is reads in the geopotentialHeight only if it is present and ignores the height.

PatNichols: @huishao-r @BenjaminRuston @fcvdb I switched the order above and the test now pass. Can you take a look and see if this is just fluke?

PatNichols: @huishao-r @fcvdb It seems to read in both height and geopotential height if only_bang is false. If that is true then it only reads in height. What would be good to know is:
1)  Is there always a height or always a geopotential height.
2)  If they are both present which is the one that matters if only_bang is true. Given that geopotential is height above mean sea level which is wrong for regions nearer to the poles ( the earth is not a sphere) then perhaps height is what we want.

PatNichols (approved):

fcvdb (approved):
I tested on the following WMO files for which refractivity is missing:
```
grace_d20210806_s012751_edzw.bfr
spi_d20220901_s000300_e001300_iuti.bfr
```
The decoder completed with the option `--onlybendingangle`, and thrown an error without the option. When the decoder completes, the `atmosphericRefractivity` field is filled with missing values.

I then tested this WMO file `gras_mta_d20210806_s010638_r14183_ekmi.bfr ` that contains atmospheric refractivity. 
Without the option `--onlybendingangle`, the decoder completes and fills in the `atmosphericRefractivity` field with the input values.
With the option `--onlybendingangle`, the decoder completes and fills in the `atmosphericRefractivity` field with missing values.

The decoder behaves as expected.

src/gnssro/gnssro_bufr2ioda.py
BenjaminRuston: we need to open an issue an can remove this line

Issue #1427, opened 2023-12-04
[Bug] bufr2ioda.x Crash When Requesting a Non-Existent Query for SATWND Conversion


spanNOAA:
## Current behavior (describe the bug)
The program usually issues a warning message for a non-existing query and then moves on to the next one. But in the case of SATWND, it unexpectedly leads to a segmentation fault when requesting a non-existing query.

## To Reproduce

> What computer are you running on?
Hera
> What compilers/modules are you using?
spack-stack-1.5.0
> Steps to reproduce the behavior

I have two examples to request non-existing queries for AIRCFT and SATWND.
Both of them are on Hera: /scratch2/BMC/wrfruc/Sijie.Pan/iodaconv/SATWND_issue

## Expected behavior

When running the "bufr2ioda.x bufr_ncep_prepbufr_aircft.yaml" command, numerous warning messages appear indicating non-existent queries. 
Conversely, executing "bufr2ioda.x bufr_ncep_prepbufr_satwnd.yaml" results in an immediate "Segmentation fault."

rmclaren:
The query `*/PRSLEVEL/Z___INFO/Z__EVENT/ZOB` should be `*/Z___INFO/Z__EVENT/ZOB` as PRSLEVEL is not a repeatable sequence (see the output for print_queries.x)... I know this is confusing as PRSLEVEL is usually present for the other data types in this file... So that will solve the issue, but I still need to look at why its crashing instead of giving a warning.

rmclaren:
Fixed in pull request

PatNichols:
@spanNOAA Can we close this one now?

spanNOAA:
Yes. Thanks!

Sijie

PR #1435, opened 2023-12-17
Validate path for inappropriate node types.

rmclaren:
## Description

We were not checking to make sure the query path string does not contain node types it should not (like non-repeating sequences). The paths would be recognized but lead to a crash instead of warning the user about the invalid path for the subset. This pull request aims to fix this...

## Issues

#1427


srherbener (approved):
Thanks for fixing this! Tested on my Mac and all ioda-converter tests passed.

PatNichols:
@rmclaren Could you explain the problem more in depth.

test/testinput/bufr_ncep_prepbufr_adpsfc.yaml
```diff
@@ -24,6 +24,8 @@ observations:
             query: "*/XOB"
           latitude:
             query: "*/YOB"
+          invalidLongitude:  # make sure that path strings can't contain sequences
```
BenjaminRuston, line 27: @rmclaren is this something that does not exist and tests that it handles it gracefully?

rmclaren: It points to something that exists but the query string includes the non-repeating sequence mnemonic which should not be part of the query string... The old version of the code will crash if you do this (the fixed version detects the issue and warns the user without crashing).

BenjaminRuston: oh, that is subtle but now follow though `XOB` is in the `HEADR` that latter sequence mnemonic should not be in the query

yes not sure how to express this any better in a succinct line

BenjaminRuston (approved):
i follow what this is doing now thanks @rmclaren for the explanation and for the catch

PR #1436, opened 2023-12-28
Adding Muon Space IODA-converter

smaticka:
## Description

Ocean surface data of `wind speed retrieval` (no direction) is available from `Muon Space` as netCDF4 files. 
A converter is needed to bring the data into `IODA` format to be used in `JEDI`. 

- Need to find a way to concatenate the nc4 files before using the ioda-converter or else the converter takes a long time. (~3000 files per day of data). 
- Currently, wind speed is pulled as the data and assumed at 10 m. need to confirm with muon. 

## Issue(s) addressed

Resolves #<issue_number>

## Checklist

- [x] the converter creates an file that can be added in r2d2 and successfully runs in skylab
- [x] the necessary data is transferred. 
- [ ] it runs as quick as possible.


fcvdb:
@smaticka The muon converter fails the `iodaconv_hdf5_coding_norms ` test. Could you run the Python utility to check the coding norms? Thanks.

smaticka:
> @smaticka The muon converter fails the `iodaconv_hdf5_coding_norms ` test. Could you run the Python utility to check the coding norms? Thanks.

I made the changes and it looks like it's passing the tests now.

PatNichols:
@smaticka @fcvdb  should we take this out draft status? It looks ready to do...

fcvdb (approved):
Thanks Sam!

BenjaminRuston:
we need to open an issue and rename this converter

should work for muon, spire, planetiq, CYGNSS from NOAA etc... 

should not contain `muon` in the name of the routines

PR #1447, opened 2024-01-23
augmenting OSW converter for Spire

smaticka:
## Description

Ocean surface data of `wind speed retrieval` (no direction) is available from `Spire` as netCDF files. 
A converter is needed to bring the data into `IODA` format to be used in `JEDI`. 

Command to test:
`python3 $JEDI_SRC/iodaconv/src/hdf5/spire_2ioda.py -i spire_gnss-br_L2_gbrOcn_v02.04_2024-01-09T20-47-47_FM172_G28.nc -o filename_spire.nc4`
(file not on publicly accessible platforms)

## Issue(s) addressed

Partial of https://github.com/JCSDA-internal/ioda-converters/issues/1452

we just going to want a single converter that forks (?) when needed for the different providers CYGNSS, Spire and Muon at this point

## Checklist

- [x] the converter creates an file that can be added in r2d2 and successfully runs in skylab
- [x] the necessary data is transferred.

### DIFF
--- a/src/hdf5/spire_2ioda.py
+++ b/src/hdf5/spire_2ioda.py
@@ -123,7 +123,7 @@ def main(args):
         # Change time reference
         obs_data_append = adjust_dateTime(obs_data_append, dat_ref)
 
-        # Change longitude range 
+        # Change longitude range
         obs_data_append = adjust_longitude(obs_data_append)
 
         # Append to data frame containing all timestamp data
@@ -194,7 +194,7 @@ def main(args):
 def get_data_from_file(afile, col_names, file_name):
 
     # Get instrument reference
-    instrument_ref = afile.attrs['tx_id'].decode('UTF-8') 
+    instrument_ref = afile.attrs['tx_id'].decode('UTF-8')
 
     # Pull each data type (variable) and create a list
     latitude = [afile['sp_lat'][ii] for ii in range(len(afile['sp_lat']))]
@@ -218,6 +218,7 @@ def get_data_from_file(afile, col_names, file_name):
 def adjust_dateTime(obs_DF, dat_ref):
     obs_DF['dateTime'] = obs_DF['dateTime']+int(dat_ref)
     return obs_DF
+
 
 def adjust_longitude(obs_DF):
     obs_DF['longitude'] = (obs_DF['longitude']+180).mod(360) - 180
--- a/src/hdf5/spire_2ioda.py
+++ b/src/hdf5/spire_2ioda.py
@@ -216,7 +216,7 @@ def get_data_from_file(afile, col_names, file_name):
 
 
 def adjust_dateTime(obs_DF, dat_ref):
-    obs_DF['dateTime'] = obs_DF['dateTime']+int(dat_ref)
+    obs_DF['dateTime'] = obs_DF['dateTime']+dat_ref
     return obs_DF
 
 
--- a/src/hdf5/spire_2ioda.py
+++ b/src/hdf5/spire_2ioda.py
@@ -126,6 +126,9 @@ def main(args):
         # Change time reference
         obs_data_append = adjust_dateTime(obs_data_append, dat_ref)
 
+        # Change longitude range 
+        obs_data_append = adjust_longitude(obs_data_append)
+
         # Append to data frame containing all timestamp data
         obs_data = pd.concat([obs_data, obs_data_append], ignore_index=True)
 
@@ -217,6 +220,10 @@ def get_data_from_file(afile, col_names, file_name):
 
 def adjust_dateTime(obs_DF, dat_ref):
     obs_DF['dateTime'] = obs_DF['dateTime']+int(dat_ref)
+    return obs_DF
+
+def adjust_longitude(obs_DF):
+    obs_DF['longitude'] = (obs_DF['longitude']+int(180)) % 360 - int(180)
     return obs_DF
 
 
--- a/src/hdf5/spire_2ioda.py
+++ b/src/hdf5/spire_2ioda.py
@@ -223,13 +223,13 @@ def adjust_dateTime(obs_DF, dat_ref):
     return obs_DF
 
 def adjust_longitude(obs_DF):
-    obs_DF['longitude'] = (obs_DF['longitude']+int(180)) % 360 - int(180)
+    obs_DF['longitude'] = (obs_DF['longitude']+180) % 360 - 180
     return obs_DF
 
 
 def quality_control(obs_data):
-    # Apply initial QC for physically possible values (wind, lat, lon) need to add swells
-    wind_range = [-100, 100]
+    # Apply initial QC for physically possible values (wind, lat, lon)
+    wind_range = [0, 100]
     lat_range = [-90, 90]
     lon_range = [-180, 180]
     # Replace with None to be filled with missing value later
--- a/src/hdf5/spire_2ioda.py
+++ b/src/hdf5/spire_2ioda.py
@@ -221,7 +221,7 @@ def adjust_dateTime(obs_DF, dat_ref):
 
 
 def adjust_longitude(obs_DF):
-    mask = obs_DF['longitude']>180
+    mask = obs_DF['longitude'] > 180
     obs_DF.loc[mask, 'longitude'] = obs_DF['longitude'].loc[mask].values - 360
     return obs_DF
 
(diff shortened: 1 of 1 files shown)


src/hdf5/spire_2ioda.py
```diff
@@ -219,6 +222,10 @@ def adjust_dateTime(obs_DF, dat_ref):
     obs_DF['dateTime'] = obs_DF['dateTime']+int(dat_ref)
     return obs_DF
 
+def adjust_longitude(obs_DF):
+    obs_DF['longitude'] = (obs_DF['longitude']+int(180)) % 360 - int(180)
```
fabiolrdiniz: Thanks, @smaticka. Not clear why `int()`?

smaticka: it might not matter. It was mostly because an `int` takes less storage space (bytes) than a `float` and if it's being added and then subtracted, I believe it reduces the amount of computational error (though probably very minor).

PatNichols:
@smaticka Need to fix coding norms.

src/hdf5/spire_2ioda.py
PatNichols: @smaticka it's redundant. int(180) is the same as 180. it just looks weird.

PatNichols: if ObsDF is a float see my suggestion.

src/hdf5/spire_2ioda.py
```diff
@@ -0,0 +1,281 @@
...
+    obs_data_append = pd.DataFrame(data_lists, columns=col_names)
+    return obs_data_append
+
+
+def adjust_dateTime(obs_DF, dat_ref):
+    obs_DF['dateTime'] = obs_DF['dateTime']+int(dat_ref)
+    return obs_DF
+
+def adjust_longitude(obs_DF):
+    obs_DF['longitude'] = (obs_DF['longitude']+int(180)) % 360 - int(180)
```
PatNichols:
```suggestion
    obs_DF['longitude'] = (int(obs_DF['longitude'])+180) % 360 - 180
```

src/hdf5/spire_2ioda.py
```diff
@@ -0,0 +1,281 @@
...
+    return obs_DF
+
+
+def quality_control(obs_data):
```
gthompsnJCSDA, line 229: Not sure how wind speed could be a negative value.  Am I mis-interpreting the information?  Is the velocity in this data feed a value of wind `towards` the instrument and `away` from the instrument such as in doppler radar?  That's the only reason a negative value should be possible, right?

src/hdf5/spire_2ioda.py
gthompsnJCSDA: I full agree with Pat on unimportant usage of `int` casting.  When not needed, simply don't cast.

gthompsnJCSDA (dismissed):

src/hdf5/spire_2ioda.py
smaticka: Thanks @PatNichols. Not sure on c++ behavior, but is `180` automatically an `int` since there's no decimal (i.e. `180.` would behave as a float). I don't think we want to cast the longitude into an `int` though. The decimal place in latitude in needed to be retained.

smaticka: so maybe this is fine?
```obs_DF['longitude'] = (obs_DF['longitude']+180) % 360 - 180```

File: src/hdf5/spire_2ioda.py (hunk 3 above)
Comment (smaticka) [reply src/hdf5/spire_2ioda.py:229]: The speed should definitely be positive. Negative would only be if it's an error, but correct, a range of `[-100, 100]' is not needed. '[0, 100]' will suffice.

File: src/hdf5/spire_2ioda.py (hunk 2 above)
Comment (PatNichols) [reply src/hdf5/spire_2ioda.py]: @smaticka if you don't cast obs_DF to an int then the mod operator will be wrong. Probably what you want is math.fmod(obs_DF['longitude']+180,360) ?

Comment (smaticka) [reply src/hdf5/spire_2ioda.py]: @PatNichols oh nice, yeah I think that's what I need. I need longitude to stay a float (then add exactly 180 to it, then take the modulus with 360, then subtract 180 from the result of the modulus, and I need that result to be a float with the decimal precision). So would it be this?:
`obs_DF['longitude'] = math.fmod(obs_DF['longitude']+180,360) - 180`

Comment (smaticka) [reply src/hdf5/spire_2ioda.py]: weirdly it worked before.. when I added int(180) to the float.. I checked the plots. The idea was to shift the data into the proper range (from 0, 360 to -180, 180). Red is original, blue is after (y is longitude, x is index number)
image

Comment (PatNichols) [reply src/hdf5/spire_2ioda.py]: Why not  (value - 180) ? Assuming value is between 0 to 360 -> -180 to 180

Comment (smaticka) [reply src/hdf5/spire_2ioda.py]: b/c 0 to 180 retain their values. but the range 181 to 360 becomes -180 to -1 (359 becomes -1 for example)

Comment (smaticka) [reply src/hdf5/spire_2ioda.py]: `math.fmod()` doesn't work on pandas series, so I used the pandas series method. (though I don't understand why the original method wasn't ok).

File: src/hdf5/spire_2ioda.py [hunk 4]
```diff
@@ -0,0 +1,278 @@
...
+    obs_data_append = pd.DataFrame(data_lists, columns=col_names)
+    return obs_data_append
+
+
+def adjust_dateTime(obs_DF, dat_ref):
+    obs_DF['dateTime'] = obs_DF['dateTime']+int(dat_ref)
+    return obs_DF
+
+def adjust_longitude(obs_DF):
+    obs_DF['longitude'] = (obs_DF['longitude']+180).mod(360) - 180
```
PatNichols: @smaticka I still think you are making this harder than it needs to be. If the longitude varies from 0 to 360 then
just subtracting 180 will give you what you need.

src/hdf5/spire_2ioda.py
```diff
@@ -0,0 +1,278 @@
...
+    data_lists = list(zip(latitude, longitude, dateTime, sensorIdentification,
+                          height, windSpeedAt10M))
+
+    # All observation data for this file to append to the master dataframe
+    obs_data_append = pd.DataFrame(data_lists, columns=col_names)
+    return obs_data_append
+
+
+def adjust_dateTime(obs_DF, dat_ref):
+    obs_DF['dateTime'] = obs_DF['dateTime']+int(dat_ref)
```
PatNichols:
```suggestion
    obs_DF['dateTime'] = obs_DF['dateTime']+dat_ref
```

PatNichols: the cast is not needed

src/hdf5/spire_2ioda.py
smaticka: 90 - 180 = -90 and that's incorrect (it should remain 90). please see my comment above: https://github.com/JCSDA-internal/ioda-converters/pull/1447#discussion_r1471656269

src/hdf5/spire_2ioda.py
PatNichols: @smaticka hmmm I get it now... Too little sleep.
What you want then is 
      if ( long > 180):
          long = -360 + long

src/hdf5/spire_2ioda.py
```diff
@@ -0,0 +1,279 @@
...
+    return obs_data_append
+
+
+def adjust_dateTime(obs_DF, dat_ref):
+    obs_DF['dateTime'] = obs_DF['dateTime']+dat_ref
+    return obs_DF
+
+
+def adjust_longitude(obs_DF):
+    obs_DF['longitude'] = (obs_DF['longitude']+180).mod(360) - 180
```
PatNichols:
```suggestion
    if ( obs_DF['longitude'] > 180.0):
         obs_DF['longitude'] =  -360.0 + obs_DF['longitude']
```

PatNichols: that gives you the proper answer, I think. No mod or casting needed.

src/hdf5/spire_2ioda.py
smaticka: @PatNichols 
An `if` statement would require a `for` loop. `mod`, which is a `pandas` `method`, is presumably vectorized, which should be more efficient than a for loop. Is there a problem with me using the `mod` method?

fcvdb (approved):
The decoder works. Thanks Sam!

PatNichols (approved):

PR #1454, opened 2024-02-02
Ioda-converters YAML for [3D]-RTMA ADPUPA BUFR

PraveenKumar-NOAA:
## Description

This PR adds an ioda-converters YAML for the [3D]-RTMA ADPUPA BUFR data.

The following files are new/ modified:

testinput/bufr_ncep_rtma_adpupa.yaml
testinput/rtma.t00z.adpupa.tm00.bufr_d
testoutput/rtma.t00z.adpupa.tm00.nc
CMakeLists.txt

The ioda-validation test is performed and following results were obtained:

**Final results:**
_errors: 0
warnings: 0_

The validation of the output ioda obs is also performed in: https://github.com/JCSDA-internal/ioda-converters/issues/1451

## Issue(s) addressed

Resolves #1450 

## Dependencies

None

## Impact

None

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have run the unit tests before creating the PR

PR #1455, opened 2024-02-05
Merge all OSW converters

### LINKED ISSUE #1452: Add converter for ocean surface wind from retrieval (CYGNSS, Muon, Spire)
## Description

Ocean surface (10 m) wind speed data is available from `CYGNSS NOAA`, `Muon`, and `Spire`. A converter is first needed to bring it into IODA-format. Since it is only wind speed (no direction), an `obs function` is also needed in `UFO` to estimate the wind velocity components. 

Wind speed (no direction) is retrieved for **CYGNSS:** using the NOAA GMF v1.0 and the NOAA trackwise debiasing algorithm; **Muon**:  using neural nets; and **Spire**: using the empirical relation between Sigma0 and background ECMWF 10m wind). 

Each source provides netCDF files. 

## Dependent PRs
CYGNSS converter: https://github.com/JCSDA-internal/ioda-converters/pull/1449
Muon converter: https://github.com/JCSDA-internal/ioda-converters/pull/1436
Spire converter: https://github.com/JCSDA-internal/ioda-converters/pull/1447

Should these be combined into 1?

smaticka:
## Description

Currently a separate converter exists for each `CYGNSS`, `Muon`, and `Spire` data. This PR is to merge the 3. (and then remove the originals). 

Example of how to run:
`python3 $JEDI_SRC/iodaconv/src/hdf5/osw_2ioda.py -i filename_input.nc -o filename_output.nc`

## Issue(s) addressed

Resolves #1452 

## Checklist

- [x] ensure the converter produces the same results as the originals
- [x] remove unneeded headers
- [x] remove the 3 original converters (muon_2ioda.py, cygns_*, spire_*)


smaticka:
the converter produces the same results as the original 3 converters, aside for the global attribute labeling and the unused variables in the Muon converter. 

the 3 original converters need to be deleted in this PR as well, but are left in for now in case people want them for testing/comparison.

src/hdf5/osw_2ioda.py
```diff
@@ -0,0 +1,315 @@
+#!/usr/bin/env python3
+#
+# (C) Copyright 2020-2023 UCAR
```
PatNichols: 2024 ?

BenjaminRuston:
@smaticka was able to run this for muon data again after my changes please put it through the paces
but think this is pretty close, we will want to remove:
`{spire,muon,cygnss}_2ioda.py` files as part of this PR as well

smaticka:
@BenjaminRuston I had to change the way you were appending strings, but now it works for me. I tested on the cygnss data.

src/hdf5/osw_2ioda.py
```diff
@@ -146,7 +146,7 @@ def main(args):
     GlobalAttrs['datetimeReference'] = datetime.fromtimestamp(obs_data['dateTime'].min()).strftime("%Y-%m-%dT%H:%M:%SZ")
     # append platform to long description
     long_description = add_long_description(osw_source)
-    GlobalAttrs['platformLongDescription'].append(long_description)
+    GlobalAttrs['platformLongDescription'] + long_description
```
BenjaminRuston: this isn't correct should be `+=` to add to the string, tried with `+=` and got the expected output for global attribute `platformLongDescription` 

pushing up that fix

BenjaminRuston:
@srherbener think this is ready to go now.  Do have a question for you.   Try to use the IODA missing values for `int` and `float` but wondering is there one for string as well see:

https://github.com/JCSDA-internal/ioda-converters/blob/33289418f9eb2b64364d884adda0a9ae652bfb6b/src/hdf5/osw_2ioda.py#L68

also more trivial bits, but would maybe like to consolodate these too into the pyiodaconv/def_jedi_utils:
https://github.com/JCSDA-internal/ioda-converters/blob/33289418f9eb2b64364d884adda0a9ae652bfb6b/src/hdf5/osw_2ioda.py#L75

can we use these `ioda_int_type` rather than `np.int32` ?
https://github.com/JCSDA-internal/ioda-converters/blob/33289418f9eb2b64364d884adda0a9ae652bfb6b/src/pyiodaconv/def_jedi_utils.py#L23

src/hdf5/osw_2ioda.py
```diff
@@ -70,11 +66,6 @@
 
 # Assign missing value details for the variables
 string_missing_value = '_'
```
srherbener, line 68: The `iconv.get_default_fill_val` function is listed below. It should be mimicking the netcdf default fill values which are good for a ioda file. The ioda reader will convert any specified fill values to the internal JEDI missing values, so might as well use the netcdf defaults. Here's the code:

https://github.com/JCSDA-internal/ioda-converters/blob/4af7be2f4a8fcdcf0e08b2518ff3c17775b4169d/src/pyiodaconv/ioda_conv_engines.py#L42-L69

It's not very clear but the np.S1 data type can be used for a string and you get an empty string back (\x00 -> null byte). Note that np.U1 or np.object can also be used to return a string containing a null byte. I believe that np.S1 is a 1 character string in ASCII, and np.U1 is a 1 character string in unicode, and np.object is a catch all type that we are using for datetime objects and if not a datetime, then it falls back on the null byte.

I think it would be good to continue using the `iconv.get_default_fill_val` function, and if you want it to take a python data type argument, then this function could be updated accordingly. We originally used numpy datatypes because all of the python converters were based on numpy arrays (which may not be true now). Also if you want to change the string default fill value to an underscore, that could be updated in the function too. Using the function helps keep things consistent across all the python converters.

srherbener (commented):
My comment is assuming that this converter is based on numpy arrays. If not maybe it makes more sense to create two default fill value functions, one for numpy arrays (which is what `iconv.get_default_fill_val` is doing and another for python objects (that takes python data types as an argument). What I'm getting at is to wrap up items like the default fill value into functions that get shared across the converters to help consistency and elimination of duplicated code, etc.

srherbener:
> @srherbener think this is ready to go now. Do have a question for you. Try to use the IODA missing values for `int` and `float` but wondering is there one for string as well see:
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/33289418f9eb2b64364d884adda0a9ae652bfb6b/src/hdf5/osw_2ioda.py#L68
> 
> also more trivial bits, but would maybe like to consolodate these too into the pyiodaconv/def_jedi_utils:
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/33289418f9eb2b64364d884adda0a9ae652bfb6b/src/hdf5/osw_2ioda.py#L75
> 
> can we use these `ioda_int_type` rather than `np.int32` ?
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/33289418f9eb2b64364d884adda0a9ae652bfb6b/src/pyiodaconv/def_jedi_utils.py#L23

I think we should use numpy data types when the converter is numpy array based, and python data types when the converter is python object based (eg, lists).

src/hdf5/osw_2ioda.py
BenjaminRuston: great thanks @srherbener think I will create a new issue and associated PR that adds all the types here:
https://github.com/JCSDA-internal/ioda-converters/blob/4af7be2f4a8fcdcf0e08b2518ff3c17775b4169d/src/pyiodaconv/def_jedi_utils.py#L22

it's missing `long` and will discuss with you about adding the null byte for missing string in that issue.  Also will then modify all the routines in this directory (and a few others) that are using this template

you're good to move forward with this PR correct?

BenjaminRuston (approved):
this is working as expected thanks for merging these @smaticka

src/hdf5/osw_2ioda.py
srherbener: I think it makes sense to create the new issue, merge this PR in now and address the data type issues later. I'll hit the approve button.

srherbener (approved):

src/hdf5/osw_2ioda.py
smaticka: ah, oops! you're right, thanks for catching it.

BenjaminRuston:
@rmclaren  and @CoryMartin-NOAA just for reference this was merged after the PR:
https://github.com/JCSDA-internal/ioda-converters/pull/1454

and passed the tests as well,,

PR #1461, opened 2024-02-13
Update cris to cris-fsr in NCAR converter

ashley314:
## Description

Updating `cris` to `cris-fsr` for the NCAR's obs2ioda converter. 

## Issue(s) addressed

Addresses https://github.com/JCSDA-internal/skylab/issues/327

## Dependencies

- [x] https://github.com/JCSDA-internal/SIMOBS/pull/235
- [x] https://github.com/JCSDA-internal/skylab/pull/341
- [x] https://github.com/JCSDA-internal/ewok/pull/929

## Impact

Changes the observation ingest suite, so we will test with that CI.

## Checklist


src/ncar-bufr2nc-fortran/define_mod.f90
```diff
@@ -343,7 +343,7 @@ subroutine set_name_sensor(instid, sensor)
       case (297); sensor = 'ahi'
       case (207); sensor = 'seviri'
       case (420); sensor = 'airs'
-      case (620); sensor = 'cris'
+      case (620); sensor = 'cris-fsr'
```
BenjaminRuston, line 346: this seems to be not flexible for `cris` and `cris-fsr` which changed a specific crossover date

BenjaminRuston:
ok @ashley314 this is essentially hardcoded to the `cris-fsr`

it is a hard date crossover where NOAA switched processing, unfortunate as the sensor ID seems to be the same 
`      case (620); sensor = 'cris-fsr'`

ashley314:
> ok @ashley314 this is essentially hardcoded to the `cris-fsr`
> 
> it is a hard date crossover where NOAA switched processing, unfortunate as the sensor ID seems to be the same ` case (620); sensor = 'cris-fsr'`

@BenjaminRuston When you say "sensor ID" are you referring to `case (620)`? I did not know how to check that so what I have here was a guess and it seemed to work

BenjaminRuston:
@ashley314 to be thorough we can get a `cris` file before the change to `cris-fsr` and see if it to is caught by the `case(620)`

if it has a unique identifier that would be great, but I'm guessing they may very well be the same

ashley314:
@BenjaminRuston I was able to convert a CrIS file from 2020-01-01 using this converter. But that file seems to contain both the old CrIS and CrIS-fsr. For our skylab purposes, I don't think files older than that will need to be converted

src/ncar-bufr2nc-fortran/define_mod.f90
climbfuji: Can we address this question/concern first before I review the set of PRs?

ashley314: Yes. @BenjaminRuston I commented below, that I was able to convert a CrIS file from 2020-01-01 using this converter. But that file seems to contain both the old CrIS and CrIS-fsr. For our skylab purposes, I don't think files older than that will need to be converted. Do you agree or have concerns with that?

BenjaminRuston: again this converter will hopefully deprecate and we'll move to the EMC one so yes agree for now let's move forward and cross idea of handling older CrIS data when we get our re-analysis funding

PatNichols:
@ashley314 Do we have to merge this with the other PRs or is it independent?

src/ncar-bufr2nc-fortran/define_mod.f90
BenjaminRuston: these numbers for obserrors are `431` long which is `fsr` we would need another vector for the `cris` data for truncated anyway.. 

in short this is hardcoded to `fsr` so this should be fine

BenjaminRuston (approved):
this is correct as this decoder is hardcoded in the vector to cris-fsr

ashley314:
@PatNichols It should be merged with the other PRs. Thanks for checking!

smaticka:
When I ran the `ingest-observations.yaml` after pulling this branch on `iodaconv` and in `ewok` and then running `make`, it completes the ingest experiment. but when I did the same thing, but also pull the branch for `skylab`, I get an error. I'll write it on the skylab PR.

ashley314:
Thanks @smaticka, I followed up in that PR. There is a [line](https://github.com/JCSDA-internal/ewok/blob/bae54b6040ee6205fb506e8285cd717d6194d4e2/src/ewok/suites/ingestObservations.py#L32) in the `EWOK` PR and you shouldn't be getting that error. Can you check your `EWOK` repo and do a`git pull` to make sure the branch is up to date and matches what is in `feature/obsingest_cris-fsr`?

smaticka (approved):
Thanks @ashley314 . it worked for me when I pulled all 3 branches and ran `ingest-observations.yaml` for just `cris`

cmgas:
If somebody can merge this I will merge the other two!

climbfuji:
> If somebody can merge this I will merge the other two!

Ready whenever you are

cmgas:
Merged!

PR #1466, opened 2024-02-16
add support for both h5 and n4 CYGNSS files

### LINKED ISSUE #1465: CYGNSS wind files not in hdf5 format netcdf4 only
## Description

the most recent CYGNSS data is not following hdf5 format but is decodable using the netCDF4 library

## Requirements

h5py and netCDF4 python libraries

decodes both hdf5 supported and netCDF4 only supported CYGNSS files

BenjaminRuston:
## Description

support CYGNSS file in both h5 and netcdf4 formats

## Issue(s) addressed

Resolves #1465 

## Checklist

### DIFF
--- a/src/hdf5/osw_2ioda.py
+++ b/src/hdf5/osw_2ioda.py
@@ -97,20 +97,26 @@ def main(args):
 
         # Open file
         # (need to move file and dat ref into get_data_from_file, but then need to move adjust data_append inside also.)
-        file = h5py.File(file_name, 'r')
+        nc4 = False
+        if h5py.is_hdf5(file_name):
+            file = h5py.File(file_name, 'r')
+        else:
+            import netCDF4 as nc
+            nc4 = True
+            file = nc.Dataset(file_name, 'r')
 
         # Figure out what source the file is coming from (e.g. cygnss, muon, spire)
-        osw_source = get_data_source(file)
+        osw_source = get_data_source(file, nc4=nc4)
 
         # Get reference time and convert to epoch time
-        dat_ref = get_reference_time(file, osw_source)
+        dat_ref = get_reference_time(file, osw_source, nc4=nc4)
 
         if 'obs_data' not in locals():
             # initialize the DF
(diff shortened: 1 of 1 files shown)


src/hdf5/osw_2ioda.py
```diff
@@ -187,7 +193,13 @@ def main(args):
     writer.BuildIoda(ioda_data, VarDims, varAttrs, GlobalAttrs)
 
 
-def get_data_source(afile):
+def get_data_source(afile, nc4=False):
+    if nc4:
+        if  'title' in afile.ncattrs() and 'CYGNSS' in afile.getncattr('title'):
+            return 'CYGNSS'
+        else:
+            return 'unsupported file type'
```
BenjaminRuston, line 201: should probably have the decoder fall over if this `unsupported` is returned

smaticka:
@BenjaminRuston, as discussed, rather than making all of these additions, we could instead run this `os` command to make the files compatible with the current converter. 

`nccopy -k 'netCDF-4' cyg.ddmi.s20240109-000001-e20240109-235959.l2.wind_trackgridsize25km_NOAAv1.2_L1a21.d21.nc cyg_copy.nc`

@ashley314. Wanted to keep you in the loop on this converter change (which we won't do). Essentially the newer `cygnss` files are not compatible with hd5f python package, but the os/terminal command above will convert the files to then be used by the current converter as is. Ben said this `nccopy -k 'netCDF-4'` will belong in the ingest script, so wanted you to be aware that this will be needed as a check. Some sort of check similar to lines 100-103 that Ben added might be needed and then as an `else` option, using `nccopy` to convert. (`nccopy -k` can also check the filetype I believe). etc etc :)

smaticka:
@BenjaminRuston can we close this now?

ashley314:
Thanks @smaticka for the information. I'll make note of it for when we add these obs to the suite

BenjaminRuston:
yes thanks @smaticka agree this should be done on the ingest step, and will discuss with the data provider if they can't remedy this directly on production

PR #1475, opened 2024-02-29
[Bugfix] Allow treatment of radar echo below detectable limit but present in data as compared to regions of no mrms data coverage

### LINKED ISSUE #1474: [Bug] MRMS radar converter needs to distinguish between data void regions and below detectable limit signal
Line 227 of `mrms_grib2ioda.py` is masking data outside the range of (-25, 80) dBZ so if there is a special value for data void regions separately from the points for which we know that the signal is present but below the detectable limit, then the distinction is lost during the format conversion.  When this happens we eliminate the `Location` (lat,lon) variable where we still want to compute the HofX, so we need to keep these points from getting masked away.

## To Reproduce

This happens behind the scenes so there is no special way to reproduce - you simply have to know what is done to the special values using the existing code.

## Expected behavior

Provide the points of signal is present but below the detectable limit so the converter outputs those locations.

gthompsnJCSDA:
## Description

The MRMS converter was chopping off data with reflectivity less than -25 dBZ.  Instead, we can capture the special values of -99 as valid data areas at below detectable signal limits.  That leaves only the other missing value of -999 in areas that are not covered by the radar composite - either terrain blocked or geometry of radar scans.  This Bugfix now allows users to include or exclude based on the lower limit of reflectivity (variable called `min_dbz` can be set using the `--skip` command line option during data conversion).  By default it is set to -35 dBZ and the special value of -99 is changed to be -34.5 for keeping those points for the output file.  In practice, it appears there is never echo in the MRMS product below -30 dBZ so the special value keeps data in a tighter range than -99.

## Issue(s) addressed

Resolves #1474 

## Dependencies

Should be no dependencies

## Impact

Should be no impact to other repos.

## Checklist

### DIFF
--- a/src/mrms/mrms_grib2ioda.py
+++ b/src/mrms/mrms_grib2ioda.py
@@ -194,6 +194,8 @@ def read_grib(input_file, obsvars):
     # Use a mask to remove the missing value points entirely from the dataset.
     # This is FLAWED because there could be different missing values on different
     # height levels, or potentially different missing values depending on product.
+    # Currently -999 values are outside radar coverage and -99 values are used
+    # for valid signal below minimum detectable threshold for reflectivity.
     '''
 
     mask = None
@@ -220,8 +222,13 @@ def read_grib(input_file, obsvars):
 
         Z = Z.reshape(ni*nj).astype('float')
         if obsvar == 'reflectivity':
-            mask = np.logical_and(Z >= -25, Z <= 80)
+            mask_inside = np.logical_and(Z <= -98.5, Z > -99.5)  # Capture -99 as special
+            Z[mask_inside] = -34.5                               # and reset to -34.5
+            mask = np.logical_and(Z >= -35, Z <= 80)
             Z = Z[mask]
+        else:
+            print(f"CAUTION: no added processing done for this product: {obsvar}")
+
         Z = Z.tolist()
         print(f"DEBUG: min, max of data: {min(Z)}, {max(Z)}")
         mrms_data[obsvar].extend(Z)
--- a/src/mrms/mrms_grib2ioda.py
+++ b/src/mrms/mrms_grib2ioda.py
@@ -232,7 +232,7 @@ def read_grib(input_file, obsvars, min_dbz):
             Z = Z.reshape(ni*nj).astype('float')
             if obsvar == 'reflectivity':
                 mask_inside = np.logical_and(Z <= -98.5, Z > -99.5)  # Capture -99 as special
-                Z[mask_inside] = -34.5                               # and reset to -34.5
+                Z[mask_inside] = -34.9                               # and reset to -34.9
                 mask = np.logical_and(Z >= min_dbz, Z <= 80)
                 Z = Z[mask]
             else:
--- a/src/mrms/mrms_grib2ioda.py
+++ b/src/mrms/mrms_grib2ioda.py
@@ -60,7 +60,8 @@ obsValName = iconv.OvalName()
 obsErrName = iconv.OerrName()
 qcName = iconv.OqcName()
 
-float_missing_value = -999.0    # or netCDF value,  nc.default_fillvals['f4']
+grib_missing_value = -999.0
+float_missing_value = nc.default_fillvals['f4']
 int_missing_value = nc.default_fillvals['i4']
 double_missing_value = nc.default_fillvals['f8']
 long_missing_value = nc.default_fillvals['i8']
@@ -79,7 +80,7 @@ dtypes = {'string': object,
           'double': np.float64}
 
 
-def main(file_names, output_file):
+def main(file_names, output_file, min_dbz):
 
     # Initialize
     varDict = defaultdict(lambda: DefaultOrderedDict(dict))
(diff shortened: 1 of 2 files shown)


src/mrms/mrms_grib2ioda.py
```diff
@@ -168,84 +168,112 @@ def main(file_names, output_file):
...
+            lons = eccodes.codes_get_array(gid, "longitudes")
+
+            Z = eccodes.codes_get_double_array(gid, 'values')
+
+            Z = Z.reshape(ni*nj).astype('float')
+            if obsvar == 'reflectivity':
+                mask_inside = np.logical_and(Z <= -98.5, Z > -99.5)  # Capture -99 as special
+                Z[mask_inside] = -34.5                               # and reset to -34.5
+                mask = np.logical_and(Z >= min_dbz, Z <= 80)
+                Z = Z[mask]
```
fabiolrdiniz: If the `-999` and `-99` values are well-known numbers for this dataset, I suggest keeping them as they are (except for `-999` because the information is useless).

I don't know what would be the consequences down the road, but my take would be to have a mask that satisfies the three conditions. Not tested, but something like below:

```suggestion
                mask = np.logical_and(Z >= min_dbz, Z <= 80)
                mask = np.logical_or(Z == -99, mask)
                Z = Z[mask]
```

This would keep `-99` in the dataset and it would allow the user to play with specific filters later.

fabiolrdiniz (commented):
Thanks, @gthompsnJCSDA!

src/mrms/mrms_grib2ioda.py
```diff
@@ -60,7 +60,8 @@
 obsErrName = iconv.OerrName()
 qcName = iconv.OqcName()
 
-float_missing_value = -999.0    # or netCDF value,  nc.default_fillvals['f4']
+grib_missing_value = -999.0
```
fabiolrdiniz, line 63: It seems to be a "missing value" for this specific dataset. Perhaps rename it to `mrms_grib_missing_value`?!

chengsiliu:
I agree with the proposed code change.

I'd like to highlight a practical aspect concerning the direct assimilation of negative dBZ values using variational techniques. We've observed that including a significant number of negative dBZ values tends to slow down the convergence of the cost function significantly. Adjusting from 10 dBZ to -20 dBZ, compared to adjusting from 30 dBZ to 0 dBZ, results in a relatively minor change in the mixing ratio. Yet, it occupies a similar portion of the cost function adjustment effort. To ensure faster convergence and efficient iteration, we typically treat negative dBZ (except for missing value) values as zero dBZ in our preprocessing procedures in practice. This approach may compromise accuracy slightly, but it significantly enhances efficiency.

chengsiliu (approved):

gthompsnJCSDA:
@chengsiliu   The timing of your review comment was impeccable.   I was formulating an Email discussion to you when this comment arrived.  Let's continue the discussion in Email.  Obviously, we can permit whatever values we want to pass through a converter and deal with them differently in QC.  That is ultimately a user decision in my opinion, and we should not be short-cutting the data because of past work if the future provides a means to discover something new and the code can operate the same regardless.

src/mrms/mrms_grib2ioda.py
gthompsnJCSDA: I propose keeping my change of -99 to a value near -35 for physical reasons and matching with long-standing radar reflectivity code introduced in WRF and adopted by MPAS and UFS-GFS/RRFS.

fabiolrdiniz: Sure!

fabiolrdiniz:
@gthompsnJCSDA, please, adjust your reference file if you are happy with the changes in the converter. The unit test is failing with:

```
ECCODES ERROR   :  Truncating time: non-zero seconds(41) ignored
DIFFER : LENGTHS OF RECORDS : Location (5419491) <> Location (249186)
DIFFER : LENGTHS : DIMENSION : Location : 5419491 <> 249186
DIFFER : VARIABLE : Location : SIZE : 5419491 <> 249186
DIFFER : VARIABLE : height : ATTRIBUTE : _FillValue : VALUES : 9.96920997e+36 <> -999
DIFFER : VARIABLE : latitude : ATTRIBUTE : _FillValue : VALUES : 9.96920997e+36 <> -999
DIFFER : VARIABLE : longitude : ATTRIBUTE : _FillValue : VALUES : 9.96920997e+36 <> -999
DIFFER : VARIABLE : dateTime : SIZE : 5419491 <> 249186
DIFFER : VARIABLE : height : SIZE : 5419491 <> 249186
DIFFER : VARIABLE : latitude : SIZE : 5419491 <> 249186
DIFFER : VARIABLE : longitude : SIZE : 5419491 <> 249186
DIFFER : VARIABLE : reflectivity : SIZE : 5419491 <> 249186
DIFFER : VARIABLE : reflectivity : SIZE : 5419491 <> 249186
DIFFER : VARIABLE : reflectivity : SIZE : 5419491 <> 249186
```

fabiolrdiniz (approved):
Thanks, @gthompsnJCSDA!

fabiolrdiniz:
> @gthompsnJCSDA, please, adjust your reference file if you are happy with the changes in the converter. The unit test is failing with:
> 
> ```
> ECCODES ERROR   :  Truncating time: non-zero seconds(41) ignored
> DIFFER : LENGTHS OF RECORDS : Location (5419491) <> Location (249186)
> DIFFER : LENGTHS : DIMENSION : Location : 5419491 <> 249186
> DIFFER : VARIABLE : Location : SIZE : 5419491 <> 249186
> DIFFER : VARIABLE : height : ATTRIBUTE : _FillValue : VALUES : 9.96920997e+36 <> -999
> DIFFER : VARIABLE : latitude : ATTRIBUTE : _FillValue : VALUES : 9.96920997e+36 <> -999
> DIFFER : VARIABLE : longitude : ATTRIBUTE : _FillValue : VALUES : 9.96920997e+36 <> -999
> DIFFER : VARIABLE : dateTime : SIZE : 5419491 <> 249186
> DIFFER : VARIABLE : height : SIZE : 5419491 <> 249186
> DIFFER : VARIABLE : latitude : SIZE : 5419491 <> 249186
> DIFFER : VARIABLE : longitude : SIZE : 5419491 <> 249186
> DIFFER : VARIABLE : reflectivity : SIZE : 5419491 <> 249186
> DIFFER : VARIABLE : reflectivity : SIZE : 5419491 <> 249186
> DIFFER : VARIABLE : reflectivity : SIZE : 5419491 <> 249186
> ```

@gthompsnJCSDA, this comment is valid still.

BenjaminRuston:
@gthompsnJCSDA reference file still needs update looks like

gthompsnJCSDA:
> @gthompsnJCSDA reference file still needs update looks like

Fixed the `testoutput` file.  Awaiting CI-tests to ensure all pass.

BenjaminRuston:
@gthompsnJCSDA CI is not working properly do you think this one is ready otherwise

gthompsnJCSDA:
> @gthompsnJCSDA CI is not working properly do you think this one is ready otherwise

Someone needs to explain to me why the JEDI unit test fails, because I don't understand why.

PatNichols:
> > @gthompsnJCSDA CI is not working properly do you think this one is ready otherwise
> 
> Someone needs to explain to me why the JEDI unit test fails, because I don't understand why.

The test did not run. There was an error setting up the CI to test the code. It's fixed now.

PR #1481, opened 2024-03-05
add missing value to ob value and error

### LINKED ISSUE #1480: handle missing value correctly in ssm converter
## Description
the surface soil moisture `ssm` missing value should be specified and set correctly

it appears the missing value in this data stream `-9999` is not correctly set to the IODA missing value

BenjaminRuston:
## Description

properly use the IODA missing value for the observation and error missing values

## Issue(s) addressed

Resolves #1480 

## Checklist

### DIFF
--- a/src/land/smap_ssm2ioda.py
+++ b/src/land/smap_ssm2ioda.py
@@ -164,11 +164,11 @@ def main():
                         type=str, required=True)
     optional = parser.add_argument_group(title='optional arguments')
     optional.add_argument(
-        '-m', '--maskMissing',
+        '--maskMissing',
         help="switch to mask missing values: default=False",
         default=False, action='store_true', required=False)
     optional.add_argument(
-        '-d', '--assumedSoilDepth',
+        '--assumedSoilDepth',
         help="default assumed depth of soil moisture in meters",
         type=float, default=0.025, required=False)
 
--- a/src/land/smap_ssm2ioda.py
+++ b/src/land/smap_ssm2ioda.py
@@ -16,6 +16,8 @@ import pyiodaconv.ioda_conv_engines as iconv
 from collections import defaultdict, OrderedDict
 from pyiodaconv.orddicts import DefaultOrderedDict
 
+float_missing_value = iconv.get_default_fill_val(np.float32)
+
 os.environ["TZ"] = "UTC"
 
 locationKeyList = [
@@ -60,6 +62,8 @@ class smap(object):
             self.varDict[iodavar]['valKey'] = iodavar, iconv.OvalName()
             self.varDict[iodavar]['errKey'] = iodavar, iconv.OerrName()
             self.varDict[iodavar]['qcKey'] = iodavar, iconv.OqcName()
+            self.varAttrs[iodavar, iconv.OvalName()]['_FillValue'] = float_missing_value
+            self.varAttrs[iodavar, iconv.OerrName()]['_FillValue'] = float_missing_value
             self.varAttrs[iodavar, iconv.OvalName()]['coordinates'] = 'longitude latitude'
             self.varAttrs[iodavar, iconv.OerrName()]['coordinates'] = 'longitude latitude'
             self.varAttrs[iodavar, iconv.OqcName()]['coordinates'] = 'longitude latitude'
@@ -98,6 +102,8 @@ class smap(object):
             errs = errs[mask]
             qflg = qflg[mask]
             times = times[mask]
+            fill = (vals < valid_min) | (vals > valid_max) | (vals == _FillValue)
+            vals[fill] = float_missing_value
 
         # file provides yyyy-mm-dd as an attribute
         # str_datetime = ncd.groups['Metadata'].groups['DatasetIdentification'].getncattr('creationDate')
--- a/src/land/smap_ssm2ioda.py
+++ b/src/land/smap_ssm2ioda.py
@@ -102,8 +102,8 @@ class smap(object):
             errs = errs[mask]
             qflg = qflg[mask]
             times = times[mask]
-            fill = (vals < valid_min) | (vals > valid_max) | (vals == _FillValue)
-            vals[fill] = float_missing_value
+        # set fillValue to IODA missing
+        vals[vals == _FillValue] = float_missing_value
 
         # file provides yyyy-mm-dd as an attribute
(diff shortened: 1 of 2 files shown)


src/land/smap_ssm2ioda.py
```diff
@@ -98,6 +102,8 @@ def _read(self):
             errs = errs[mask]
             qflg = qflg[mask]
             times = times[mask]
+            fill = (vals < valid_min) | (vals > valid_max) | (vals == _FillValue)
```
BenjaminRuston: should I take this out of the `if self.mask:` conditional ???

src/land/smap_ssm2ioda.py
```diff
@@ -98,6 +102,8 @@ def _read(self):
...
+            vals[fill] = float_missing_value
```
BenjaminRuston:
```suggestion
        vals[vals == _FillValue] = float_missing_value

```

BenjaminRuston: @fabiolrdiniz think this is the way to do it will try to test it

src/land/smap_ssm2ioda.py
fabiolrdiniz: Maybe take it out of the `if self.mask:` and only check for `fill = (vals == _FillValue)`?! This would keep the same behavior of the code with the fix for what we are looking for.

src/land/smap_ssm2ioda.py
fabiolrdiniz: Oh... just saw this comment. Yes, similar to what I tried to say in my previous comment.

fabiolrdiniz:
@BenjaminRuston, let's remove the `-d` option too.

fabiolrdiniz:
@BenjaminRuston, just pushed that change.

BenjaminRuston:
@YoulongXia-NOAA and @ClaraDraper-NOAA 
do you use the `smap9km_ssm2ioda.py` I just use the file in this PR
should we remove the other one?

YoulongXia-NOAA:
@BenjaminRuston, for 9km smap data, @ClaraDraper-NOAA group is using for testing. For the other SMAP ioda converter, it will use for NRT smap data when EMC move soil moisture DA to ops work. I also developed the ioda converts for other soil moisture data. Which converter do you want to keep and which ones want to remove, you need to talk with @ClaraDraper-NOAA and EMC land DA/DA feds to make the decision.

ClaraDraper-NOAA:
> @BenjaminRuston, for 9km smap data, @ClaraDraper-NOAA group is using for testing. For the other SMAP ioda converter, it will use for NRT smap data when EMC move soil moisture DA to ops work. I also developed the ioda converts for other soil moisture data. Which converter do you want to keep and which ones want to remove, you need to talk with @ClaraDraper-NOAA and EMC land DA/DA feds to make the decision.

@YoulongXia-NOAA I was going to tell Ben to ask you! (or EMC). My expectation is that we should be supporting all of the converters that have been added (as in, I'd be surprised if anything has changed and we don't need them all anymore).

YoulongXia-NOAA:
@ClaraDraper-NOAA, @BenjaminRuston, now land has the converters below: afwa_snod2ioda.py  ims_scf2ioda.py     smap9km_ssm2ioda.py, ascat_ssm2ioda.py, imsfv3_scf2ioda.py, smap_ssm2ioda.py, ghcn_snod2ioda.py  owp_snow_obs.py     smos_ssm2ioda.py. 

 imsfv3_scf2ioda.py and ghcn_snod2ioda.py are used for snow DA, owp_snow_obs.py  developed by NCAR people. smap9km_ssm2ioda.py is used by Clara, smap_ssm2ioda.py will be used by EMC for NRT smap soil moisture.

ascat_ssm2ioda.py and smos_ssm2ioda.py I developed. I am not sure if they will used. If Clara think that they are not useful, they can be removed.  afwa_snod2ioda.py and ms_scf2ioda.py use pygrib to read grib file I developed originally and no one is planned to use. You can remove these two converters as well.  In total you can remove these four files. What do you think?

YoulongXia-NOAA (dismissed):

BenjaminRuston:
appreciate the comments...

would you like me to make corresponding changes to the `9km` at this point let's keep both

regarding the new converters we should open a new issue and get those in and tested

YoulongXia-NOAA:
Okay, please do so. Thank you.

ClaraDraper-NOAA:
> @ClaraDraper-NOAA, @BenjaminRuston, now land has the converters below: afwa_snod2ioda.py ims_scf2ioda.py smap9km_ssm2ioda.py, ascat_ssm2ioda.py, imsfv3_scf2ioda.py, smap_ssm2ioda.py, ghcn_snod2ioda.py owp_snow_obs.py smos_ssm2ioda.py.
> 
> imsfv3_scf2ioda.py and ghcn_snod2ioda.py are used for snow DA, owp_snow_obs.py developed by NCAR people. smap9km_ssm2ioda.py is used by Clara, smap_ssm2ioda.py will be used by EMC for NRT smap soil moisture.
> 
> ascat_ssm2ioda.py and smos_ssm2ioda.py I developed. I am not sure if they will used. If Clara think that they are not useful, they can be removed. afwa_snod2ioda.py and ms_scf2ioda.py use pygrib to read grib file I developed originally and no one is planned to use. You can remove these two converters as well. In total you can remove these four files. What do you think?

We should keep the ASCAT and SMOS converters. @BenjaminRuston  - can you please make the same changes to those, if it's relevant?

BenjaminRuston:
@ClaraDraper-NOAA this PR is just dealing with the SMAP converter and was wondering if we needed both the:
`smap_ssm2ioda.py`  and the  `smap9km_ssm2ioda.py`

so for now made parallel changes.  Will not touch the `ASCAT` and `SMOS` 

but think this is ready and please let me know if this is OK... of note we decided to remove the *shorthand* switches as they could be misinterpreted too easily.

so to run with the *masking* which is typical the `-m` option is removed leaving only the `--maskMissing`
similarly the `-d` (which often mean *date*) but in this case is assumed depth of measurement now can only be triggered using the full name `--assumedSoilDepth`

so please review any scripts you may have in place as these may need to be updated

ClaraDraper-NOAA (approved):
Looks good to me. I like the new expanded input flags.

fcvdb (approved):
Thanks Ben!

PR #1483, opened 2024-03-11
Update the satellite bias coefficient(BC) upgrader following naming conventions

huishao-r:
jedi-ci-build-cache=skip

## Description
During the earlier VarBC code sprint, we decided to update the BC files with new format as well as following the IODA naming conventions. This upgrader was initially developed by Cory updates the BC files for the dimension part as well as some variable name changes. This PR is intended to complete this work, following the discussions with our kinds teams (EMC, MetOffice). 

Please note, this upgrader only works for the BC file in the format being used currently in develop (ufo-data)! The data files in VarBC-sprint (which has partial changes already) shouldnt use the upgrader in this PR. 

The new format will have two-dimension (Channel and Record) arrays for coefficients and errors per predictor. For satellite radiance BC, the dimension Record is not being used at this point and therefore is fixed to be 1. This Record dimension will be adopted in the new future for aircraft BC (or BC for other observation types.). For now, this PR focuses on satellite radiance (or any channel dependent observations). 

Naming convention: 
(1) uses CamelCase, instead of underscore in the variable names, unless some exceptions, e.g. scan_angle_order_1-> sensorScanAngle_order_1
(2) uses existing variable names from observation naming convention: e.g., channels->sensorChannelNumber, cloud_liquid_water->cloudWaterContent

Use of this upgrader: 
python3 satbias_upgrader.py -i input.nc4 -o output.nc4

## Impact

The commit this upgrader wouldn't have any impacts. However, use of this upgrader to BC files in develop would have impacts on coding and tests.

## Checklist

### DIFF
--- a/src/gsi_varbc/satbias_upgrader.py
+++ b/src/gsi_varbc/satbias_upgrader.py
@@ -54,7 +54,8 @@ def satbias_upgrader(infile, outfile):
     'scanAngle': 'sensorScanAngle',
     'zenithAngle': 'sensorZenithAngle',
     'cloudLiquidWater': 'cloudWaterContent',
-    'orbialAngle': 'satelliteOrbitalAngle'
+    'orbialAngle': 'satelliteOrbitalAngle',
+    'emissivity': 'emissivityJacobian'
     }
 
     if 'bias_coefficients' in oldnc.variables.keys():
--- a/src/gsi_varbc/satbias_upgrader.py
+++ b/src/gsi_varbc/satbias_upgrader.py
@@ -60,7 +60,7 @@ def satbias_upgrader(infile, outfile):
         'scanAngle': 'sensorScanAngle',
         'zenithAngle': 'sensorZenithAngle',
         'cloudLiquidWater': 'cloudWaterContent',
-        'orbialAngle': 'satelliteOrbitalAngle',
+        'orbitalAngle': 'satelliteOrbitalAngle',
         'emissivity': 'emissivityJacobian'
     }
 
--- a/src/gsi_varbc/satbias_upgrader.py
+++ b/src/gsi_varbc/satbias_upgrader.py
@@ -5,7 +5,6 @@ import argparse
 import netCDF4 as nc
 import numpy as np
 
-
 def satbias_upgrader(infile, outfile):
     # convert satbias files from old to new format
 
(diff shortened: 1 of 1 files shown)


src/gsi_varbc/satbias_upgrader.py
```diff
@@ -52,19 +55,39 @@ def satbias_upgrader(infile, outfile):
...
         for i, pred in enumerate(predictors):
             temp = pred.split('_')
-            predOut = temp[0] + ''.join(ele.title() for ele in temp[1:])
-            if predOut == 'emissivity':
-                predOut = 'emissivityK'
+            if "order" in temp:
+                idorder = temp.index('order')
+                predOut = temp[0] + ''.join(ele.title() for ele in temp[1:idorder])+'_'+temp[idorder]+'_'+temp[idorder+1]
+            else:
+                predOut = temp[0] + ''.join(ele.title() for ele in temp[1:])
```
CoryMartin-NOAA: This is fine but we could use a dictionary instead of all these `if` statements

CoryMartin-NOAA: Also @ADCollard is going to want that emissivityK thing kept in here.

ADCollard: Yes, came here to say that the predictor in the GSI is `emissivityK` not `emissivity`, so this conversion should be done.  I don't know if anyone uses `emissivity` as a predictor (which would make this line confusing if we had to deal with both cases).

huishao-r: True. It was added on the fly one after another when I was updating the variable names. Let's me try to update these with a dictionary.

huishao-r: I used dictionary to replace all the ifs except the "order" part (I don't know whether there is a different way to do this. Please feel free to sugguest an alternative). 
```
            idorder = temp.index('order') if "order" in temp else None
            predOut = temp[0] + ''.join(ele.title() for ele in temp[1:idorder]) + \
                  ('_' + temp[idorder] + '_' + temp[idorder + 1] if idorder is not None else '')

            # Replace strings using dictionary
            for key, value in replace_dict.items():
                if key in predOut:
                    predOut = predOut.replace(key, value)
```

CoryMartin-NOAA (approved):

PatNichols (approved):

src/gsi_varbc/satbias_upgrader.py
```diff
@@ -40,31 +39,54 @@ def satbias_upgrader(infile, outfile):
...
         nobs_assim_out = newnc.createVariable("numberObservationsUsed", "i4", ("Record", dimname))
         nobs_assim_out[0, :] = nobs_assim_in
 
     # loop through predictors and create predictor variables
+    replace_dict = {
+    'scanAngle': 'sensorScanAngle',
+    'zenithAngle': 'sensorZenithAngle',
+    'cloudLiquidWater': 'cloudWaterContent',
+    'orbialAngle': 'satelliteOrbitalAngle'
+    }
```
ADCollard, line 65: should be `orbitalAngle`.  `emissivity` seems to have disappeared!

PatNichols:
@huishao-r Coding norms error. I will fix it later today.

huishao-r:
@PatNichols You need to merge develop to feature/VarBC-sprint first? Then merge feature/VarBC-sprint to this branch. Otherwise, we will continue to see additional files (from develop) diverges from feature/VarBC-sprint.

ADCollard (changes requested):

src/gsi_varbc/satbias_upgrader.py
huishao-r: Thanks for catching that. Luckily, the data files have been updated with the older version. I can see the emissivity part though like this:
'emissivity': 'emissivityJacobian' (line 64)

huishao-r: oh. We can't see it in this comment section. But if you go to the code review, you will see it. This comment was based on the previous version.

ADCollard: @huishao-r Sorry, I had a pending comment that I just hit submit for.   `emissivityJacobian` is great!

ADCollard (approved):

BenjaminRuston:
@huishao-r if you are happy with this you should be able to merge it since it's not to develop and into a branch

PR #1485, opened 2024-03-14
hack to output impactHeight and correctly sort observations

BenjaminRuston:
## Description

the `gsi_ncdiag.py` did not have the ability to convert the `conv_gps` to IODA
these changes allow this, but had to remove the observation sorting and make the assumption that the `height` is the `impactHeightRO` which appear to be the case 

## Issue(s) addressed

Resolves #<issue_number>

## Checklist

### DIFF
--- a/src/gsi_ncdiag/gsi_ncdiag.py
+++ b/src/gsi_ncdiag/gsi_ncdiag.py
@@ -928,11 +928,9 @@ class Conv(BaseGSI):
 
                 if v == 'bend':
                     # sort record_number
-                    idx = range(self.nobs)
-                    ### record_number = self.var('record_number')[idx]
-                    ### record_number = self.var('sequenceNumber')[idx]
-                    ### id_recordnum_sort = sorted(range(len(record_number)), key=record_number.__getitem__)
-                    print("Sorting ", v, " obs referring to record_number")
+                    record_number = self.var('Observation_Subtype')[idx]
+                    id_recordnum_sort = sorted(range(len(record_number)), key=record_number.__getitem__)
+                    # print("Sorting ", v, " obs referring to Observation_Subtype")
                     # record_number_sorted = [ record_number[ksort] for ksort in id_recordnum_sort ]
                     # print('record_number:', record_number)
                     # print('record_number.size:', record_number.size)
@@ -941,10 +939,10 @@ class Conv(BaseGSI):
                     #         record_number[isort], record_number_sorted[isort] )
 
                     # Shuffle idx referring to sorted record_number's subscripts "id_recordnum_sort".
-                    ### idx_tuples = np.where(idx.data)
-                    ### idx_id = idx_tuples[0]
-                    ### idx_sorted = [idx_id[ksort] for ksort in id_recordnum_sort]
-                    ### idx = idx_sorted
+                    idx_tuples = np.where(idx.data)
+                    idx_id = idx_tuples[0]
+                    idx_sorted = [idx_id[ksort] for ksort in id_recordnum_sort]
+                    idx = idx_sorted
 
                 for o in range(len(outvars)):
                     obsdata = self.var(conv_gsivarnames[v][o])[idx]
--- a/src/gsi_ncdiag/gsi_ncdiag.py
+++ b/src/gsi_ncdiag/gsi_ncdiag.py
@@ -120,6 +120,8 @@ all_LocKeyList = {
     'geoid_height_above_reference_ellipsoid': ('geoidUndulation', 'float'),
     'gnss_sat_class': ('satelliteConstellationRO', 'string'),
     'impact_height': ('impactHeightRO', 'float'),
+    'height': ('impactHeightRO', 'float'),
+    'Height': ('impactHeightRO', 'float'),
     'impact_parameter': ('impactParameterRO', 'float'),
     'Latitude': ('latitude', 'float'),
     'Longitude': ('longitude', 'float'),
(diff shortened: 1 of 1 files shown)


src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -120,6 +120,8 @@
     'geoid_height_above_reference_ellipsoid': ('geoidUndulation', 'float'),
     'gnss_sat_class': ('satelliteConstellationRO', 'string'),
     'impact_height': ('impactHeightRO', 'float'),
+    'height': ('impactHeightRO', 'float'),
+    'Height': ('impactHeightRO', 'float'),
```
BenjaminRuston: I thought these lines would take the key in the file `Height` and then map it to `MetaData/impactHeightRO` but this did not work so was done brute force below

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -926,8 +928,10 @@ def toIODAobs(self, OutDir, TotalBias=False, clobber=True, platforms=None):
 
                 if v == 'bend':
                     # sort record_number
-                    record_number = self.var('record_number')[idx]
-                    id_recordnum_sort = sorted(range(len(record_number)), key=record_number.__getitem__)
+                    idx = range(self.nobs)
```
BenjaminRuston: I removed the sorting as `record_number` is not present in the ncDiag `conv_gps` file

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -1093,6 +1097,10 @@ def toIODAobs(self, OutDir, TotalBias=False, clobber=True, platforms=None):
                             #     tmp = hgt
                             outdata[(loc_mdata_name, 'MetaData')] = tmp
                             varAttrs[(loc_mdata_name, 'MetaData')]['units'] = 'm'
+                        elif v == 'bend':
+                            loc_mdata_name = 'impactHeightRO'
+                            outdata[(loc_mdata_name, 'MetaData')] = self.var(lvar)[idx]
+                            varAttrs[(loc_mdata_name, 'MetaData')]['units'] = 'm'
```
BenjaminRuston, line 1099: here is the lovely mapping is `Height` is in the key and the variable is `bend` put that into `MetaData/impactHeightRO`

BenjaminRuston:
these changes do allow the `conv_gps_ges` and `conv_gps_anl` to be put into a feedback file and produce some routine monitoring.

but should probably flush out a bit more

gnssro cosmic-2 bendingAngle profile 2022021521

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -926,8 +928,10 @@ def toIODAobs(self, OutDir, TotalBias=False, clobber=True, platforms=None):
 
                 if v == 'bend':
                     # sort record_number
-                    record_number = self.var('record_number')[idx]
-                    id_recordnum_sort = sorted(range(len(record_number)), key=record_number.__getitem__)
+                    idx = range(self.nobs)
+                    ### record_number = self.var('record_number')[idx]
+                    ### record_number = self.var('sequenceNumber')[idx]
```
BenjaminRuston: this `sequenceNumber` line I added as well as thought this may work in replacement for `record_number` but did not work

src/gsi_ncdiag/gsi_ncdiag.py
CoryMartin-NOAA: hmm, yeah it should've done that, weird...

src/gsi_ncdiag/gsi_ncdiag.py
CoryMartin-NOAA: This might be something that we added for validation @XuanliLi-NOAA do you know? If so, perhaps we add a (pseudo code) 'if variable in file, sort, otherwise don't'

CoryMartin-NOAA (commented):
Just want clarification on the record sorting before I approve

src/gsi_ncdiag/gsi_ncdiag.py
XuanliLi-NOAA: In conv_gps file, is there a variable called "Observation_Subtype"? If so, you can use it for "record_number". We tried to comment this line out before, but it didn't work because the sorting of record_number not only used in toGeovals but also in toIODAobs (about 150 lines after the toGeovals).

BenjaminRuston: thank you @XuanliLi-NOAA that worked!
I thought the `Observation_Subtype` was receiver ID but indeed it looks like the number to track the occultation profile!

here's an example from `diag_conv_gps_ges.2022021600.nc4`:
```
(Pdb) np.unique(self.var('Observation_Subtype'))
array([   1,    2,    3, ..., 2237, 2238, 2239], dtype=int32)
```

src/gsi_ncdiag/gsi_ncdiag.py
BenjaminRuston: @CoryMartin-NOAA was pretty obvious the `Height` was included later in the `all_LocKeyList` mapping.. so that is why my new (and incorrect mapping -- would mess up other data) didn't work.. and also why it shouldn't be included

so since I'm not seeing `impact_height` in the `conv_gps` looks like this special case to map `Height` may be needed

src/gsi_ncdiag/gsi_ncdiag.py
```diff
@@ -926,9 +926,9 @@ def toIODAobs(self, OutDir, TotalBias=False, clobber=True, platforms=None):
 
                 if v == 'bend':
                     # sort record_number
-                    record_number = self.var('record_number')[idx]
+                    record_number = self.var('Observation_Subtype')[idx]
```
BenjaminRuston, line 929: and @XuanliLi-NOAA here it is!!!
https://github.com/JCSDA-internal/ioda-converters/blob/f02f792dc42ad78dbecf9f83ed8fd6ee3ffffb8e/src/gsi_ncdiag/gsi_ncdiag.py#L119

yes indeed this is what the sort should be on.. I do not know if or what changes are needed in the `geovals` section as do not have any of those files for testing

src/gsi_ncdiag/gsi_ncdiag.py
XuanliLi-NOAA: @BenjaminRuston, just want to add one thing, `Height` in `conv_gps` is computed as **impact_parameter - LOC**, not **impact_parameter - LOC - geoid**.

CoryMartin-NOAA (approved):

XuanliLi-NOAA (approved):
Looks good to me. Approve.

PR #1486, opened 2024-03-14
Windborne data is reverse labelled

### LINKED ISSUE #1489: begin a balloonsonde converter for ingest
## Description

the ballonsonde data provided in `.json` format needs an update to the converter to IODA format for JEDI

there is some mislabeling of the wind components and this drove an update to the API

backward compatibility and handling of the new names 
`speed_y` to `speed_u` 
and
`speed_x` to `speed_v` 

should be handled

smaticka:
## Description

o-b looks a lot better in skylab if the u and v wind components are reversed first. to test whether this is true before applying hofx, we need to adjust the converter, re-convert the data, and re-run JEDI for the 2023 sept period. 

After talking to Windborne, their labeling was backwards and `speed_x` is now changed to `speed_v` and `speed_y` to `speed_u`. the converter is updated to handle both cases (speed_u or speed_y). 

The variable names from their previous data were compared to their updated obs website, and they match, aside from the addition of speed_u and _v. (march 21st 2024). There is expected to be a specific humidity variable to come in the future. 

**Their converter script:** https://github.com/windborne/to-prepbufr/blob/main/wb_to_prepbufr.py
^in which, they convert RH to Q and give error estimates:
```
# Right now, we simply hardcode these values to reasonable averages
        # However, in the future we may calculate these ourselves
        wind_x_error = 1.18
        wind_y_error = 1.18
        relative_humidity_error = 8.23
        temperature_error = 0.83
```

## Issue(s) addressed

Resolves #1489

### DIFF
--- a/src/conventional/balloon_sonde_json2ioda.py
+++ b/src/conventional/balloon_sonde_json2ioda.py
@@ -153,10 +153,19 @@ def main(args):
         latitude = [file['observations'][ii]['latitude'] for ii in range(len(file['observations']))]
         longitude = [file['observations'][ii]['longitude'] for ii in range(len(file['observations']))]
         stationIdentification = [file['observations'][ii]['mission_name'] for ii in range(len(file['observations']))]
-        pressure = [file['observations'][ii]['pressure'] for ii in range(len(file['observations']))]
-        # NOTE: THESE 2 ARE TEMPORARILY REVERSED UNTIL DATA LABELING CONFIRMED
-        windEastward = [file['observations'][ii]['speed_y'] for ii in range(len(file['observations']))]
-        windNorthward = [file['observations'][ii]['speed_x'] for ii in range(len(file['observations']))]
+        pressure = [file['observations'][ii]['pressure'] for ii in range(len(file['observations']))] 
+        # Check when the file was created to handle wind components accordingly
+        dat_created = datetime.strptime(file['CreateDateTime'],"%Y-%m-%dT%H:%M:%S.%f")
+        # UPDATE THIS DATE WHEN Windborne CHANGES THEIR OBS RECORDS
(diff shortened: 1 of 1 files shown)


fabiolrdiniz:
@smaticka, thanks for creating it. Coding norms are failing with:
```
./balloon_sonde_json2ioda.py:157:9: E266 too many leading '#' for block comment
```

fcvdb (approved):
Thanls Sam!

src/conventional/balloon_sonde_json2ioda.py
```diff
@@ -154,8 +154,10 @@ def main(args):
         longitude = [file['observations'][ii]['longitude'] for ii in range(len(file['observations']))]
         stationIdentification = [file['observations'][ii]['mission_name'] for ii in range(len(file['observations']))]
         pressure = [file['observations'][ii]['pressure'] for ii in range(len(file['observations']))]
-        windEastward = [file['observations'][ii]['speed_x'] for ii in range(len(file['observations']))]
-        windNorthward = [file['observations'][ii]['speed_y'] for ii in range(len(file['observations']))]
+        #### NOTE: THESE 2 ARE TEMPORARILY REVERSED UNTIL DATA LABELING CONFIRMED
```
BenjaminRuston: we should really add the check on file creation date, or software version to trigger this guy

thanks for acting so quickly @smaticka

smaticka: ok, I can set it so it checks to be before a certain date, but that date should correspond to when Windborne changes their files (if they change them). Do you have plans to talk with them?

BenjaminRuston (approved):
thanks @smaticka this should be it with this for now

PR #1488, opened 2024-03-25
update handing of fractional seconds

### LINKED ISSUE #1487: fractional seconds in ocean surface wind vary in precision
## Description

the fractional seconds in the ocean surface wind product when provided can vary in precision

this can be problematic with the way the time is now stripped from the file, modify this to be more flexible and also do a floor on the seconds by stripping off the fractional seconds as this should not impact use for this data

BenjaminRuston:
## Description

jedi-ci-build-cache=skip

update handing of fractional seconds in ocean surface wind converter,

there are different lengths of the fractional seconds which with the string extraction method didn't work correctly, this fixes that

also functionally do a floor of the seconds to handle cases where too many digits are provided in the fractional seconds and this also cased issues with datetime ptime function

## Issue(s) addressed

Resolves #1487 

## Checklist

### DIFF
--- a/src/hdf5/osw_2ioda.py
+++ b/src/hdf5/osw_2ioda.py
@@ -204,7 +204,7 @@ def get_reference_time(afile, osw_source):
         dat_ref = afile['time'].attrs['units'][-19:].decode('UTF-8')
         dat_ref = datetime.strptime(dat_ref, '%Y-%m-%d %H:%M:%S').replace(tzinfo=timezone.utc).timestamp()
     elif osw_source in ('Spire'):
-        # the precision of the seconds appears troublesome when too many digits 
+        # the precision of the seconds appears troublesome when too many digits
         # take the floor of the seconds by stripping of fractional seconds
         dat_ref = afile['sample_time'].attrs['units'].decode('UTF-8').split()[-1]
         dat_ref = dat_ref.split('.')[0]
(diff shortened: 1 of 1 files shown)


src/hdf5/osw_2ioda.py
```diff
@@ -204,8 +204,11 @@ def get_reference_time(afile, osw_source):
         dat_ref = afile['time'].attrs['units'][-19:].decode('UTF-8')
         dat_ref = datetime.strptime(dat_ref, '%Y-%m-%d %H:%M:%S').replace(tzinfo=timezone.utc).timestamp()
     elif osw_source in ('Spire'):
-        dat_ref = afile['sample_time'].attrs['units'][-29:-3].decode('UTF-8')
```
BenjaminRuston, line 207: @smaticka this was what was causing the crash on certain files... the fractional seconds can vary in precision

so just changed to a generic split and then dropped the fractional seconds which for this data type is not critical

PatNichols (approved):
Looks good.

smaticka (approved):
thanks @BenjaminRuston it worked when I ran it on a jan 2024 file!

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

fcvdb (approved):
Thanks Ben!

PR #1496, opened 2024-04-10
update gnss-r osw converter

### LINKED ISSUE #1495: update gnss-r ocean surface wind converter
## Description

update the ocean surface wind converter to gather some additional metaData in the files this may be specific per vendor

particularly examining the `wind_confidence`, the `wind_std` in the file itself and `ice_flag` 

## Requirements

python interface to IODA 

a IODA formatted observation file is produced for osw data

BenjaminRuston:
## Description

Fix a few small issue with the osw converter and update with new metaData allowing more strict quality control

this uses wind standard deviation, presence of sea-ice and a wind confidence from vendor

## Issue(s) addressed

Resolves #1495 

## Checklist

### DIFF
--- a/src/hdf5/osw_2ioda.py
+++ b/src/hdf5/osw_2ioda.py
@@ -25,8 +25,12 @@ from pyiodaconv.orddicts import DefaultOrderedDict
 from pyiodaconv.def_jedi_utils import int_missing_value, long_missing_value, float_missing_value
 from collections import defaultdict
 
+# Globals
 metaDataName = iconv.MetaDataName()
 obsValName = iconv.OvalName()
+obsErrName = iconv.OerrName()
+obsQcName = iconv.OqcName()
+
 
 GlobalAttrs = {
     "converter": os.path.basename(__file__),
@@ -48,21 +52,14 @@ MetaDataKeyList = [
 meta_keys = [m_item[0] for m_item in MetaDataKeyList]
 
 # The outgoing IODA variables (ObsValues), their units, and assigned constant ObsError.
-obsvars = ['windSpeed']
-obsvars_units = ['m s-1']
-# obserrlist = [1.2]
-obsvars_dtype = ['float']
+obsvars = ['windSpeed', 'windSpeed'+obsErrName, 'windSpeed'+obsQcName]
+obsvars_units = 'm s-1'
+obsvars_dtype = 'float'
 
 # Assign dimensions to the obs values
 VarDims = {
     'windSpeed': ['Location'],
 }
-
-# creating data types
-metaDataName = iconv.MetaDataName()
-obsValName = iconv.OvalName()
-obsErrName = iconv.OerrName()
-qcName = iconv.OqcName()
 
 # Assign missing value details for the variables
 string_missing_value = '_'
@@ -133,8 +130,8 @@ def main(args):
     # replace missing values
     for MetaDataKey in MetaDataKeyList:
         obs_data[MetaDataKey[0]].fillna(missing_vals[MetaDataKey[1]], inplace=True)
-    for n, obsvar in enumerate(obsvars):
-        obs_data[obsvar].fillna(missing_vals[obsvars_dtype[n]], inplace=True)
+    iodavar = obsvars[0]
+    obs_data[iodavar].fillna(missing_vals[obsvars_dtype], inplace=True)
 
     # sort by instrument and then time
     if args.sort:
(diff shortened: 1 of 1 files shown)


BenjaminRuston:
just getting started more to do here but thought best to get the ball rolling

fcvdb (approved):
Thanks Ben

BenjaminRuston:
@smaticka think this is ready, worked through the other two `osw` datasets and now all should pass something for preQC and error estimate

src/hdf5/osw_2ioda.py
```diff
@@ -94,6 +91,7 @@ def main(args):
             print(f'Input (-i option) file: {file_name} does not exist')
             sys.exit()
         logging.debug(f"Reading input file: {file_name}")
+        print(f"Reading input file: {file_name}")
```
fabiolrdiniz:
```suggestion
```

src/hdf5/osw_2ioda.py
```diff
@@ -48,22 +52,15 @@
 meta_keys = [m_item[0] for m_item in MetaDataKeyList]
 
 # The outgoing IODA variables (ObsValues), their units, and assigned constant ObsError.
-obsvars = ['windSpeed']
-obsvars_units = ['m s-1']
-# obserrlist = [1.2]
-obsvars_dtype = ['float']
+obsvars = ['windSpeed', 'windSpeed'+obsErrName, 'windSpeed'+obsQcName]
```
fabiolrdiniz, line 55: Why the other variables are needed?
```suggestion
obsvars = ['windSpeed']
```

BenjaminRuston: yes they are needed, this is the data quality indicator particularly for the Spire data where they recommend when the retrieval is valid.   This is the `'windSpeed+obsQcName`

the other is the error estimate `'windSpeed+obsErrName`, this does exist for all the different `osw` we have so far

BenjaminRuston:
@smaticka could you have a look at this, let me know if you see anything

src/hdf5/osw_2ioda.py
fabiolrdiniz: Sorry, @BenjaminRuston. I didn't see where they are used inside the converter. I see `obsvars[0]`, but nothing else. Am I missing something?

fabiolrdiniz: I see they could maybe used around lines 177-180 in your new version of the code, but they are not used there. This is what makes me think the modification in this line is not needed.

src/hdf5/osw_2ioda.py
BenjaminRuston: we should discuss this, but think this should always be printed so added it intentionally

fabiolrdiniz: Sure, but why keep two print lines? The line above can be changed to `logging.info` instead of debug.

src/hdf5/osw_2ioda.py
BenjaminRuston: yes it's confusing, and yes these are used but these are the names or column titles in the pandas dataFrame, and then when you pull them out to put them into IODA they are extracted again a bit more manually in these two lines:
https://github.com/JCSDA-internal/ioda-converters/blob/9f23fae01decdb7b847d4d596f2b099f056c92e7/src/hdf5/osw_2ioda.py#L179
https://github.com/JCSDA-internal/ioda-converters/blob/9f23fae01decdb7b847d4d596f2b099f056c92e7/src/hdf5/osw_2ioda.py#L180

fabiolrdiniz: I see it now, it's used only here:

https://github.com/JCSDA-internal/ioda-converters/blob/9f23fae01decdb7b847d4d596f2b099f056c92e7/src/hdf5/osw_2ioda.py#L108

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

src/hdf5/osw_2ioda.py
```diff
@@ -94,6 +91,7 @@ def main(args):
             print(f'Input (-i option) file: {file_name} does not exist')
             sys.exit()
         logging.debug(f"Reading input file: {file_name}")
```
BenjaminRuston:
```suggestion
        logging.info(f"Reading input file: {file_name}")
```

src/hdf5/osw_2ioda.py
```diff
@@ -93,7 +90,8 @@ def main(args):
             logging.debug(f'Input (-i option) file: {file_name} does not exist')
             print(f'Input (-i option) file: {file_name} does not exist')
             sys.exit()
-        logging.debug(f"Reading input file: {file_name}")
+        logging.info(f"Reading input file: {file_name}")
+        print(f"Reading input file: {file_name}")
```
BenjaminRuston:
```suggestion
```

PR #1501, opened 2024-04-19
add compatibility for two labeling of speeds

### LINKED ISSUE #1500: bugfix balloonsonde converter key for wind

this bugfix still accommodates data that has incorrect specification of wind (denoted by presence of `wind_x` and `wind_y`) this has been corrected and now `speed_u` and `speed_v` are used

## To Reproduce

try to convert any of the new corrected .json files with `speed_u` and `speed_v`

## Expected behavior

when using bugfix a IODA file is generated

BenjaminRuston:
## Description

there have been changes to the wind speed files being ingested from Windborne, this includes changing from wind speed in the `x` and `y` to `u` and `v`.  

The winds were switched in the API from mis-labeled `speed_x` and `speed_y` to `speed_u` and `speed_v`.  The data was backprocssed.

However on 20240322T161500Z the files began showing *both* `speed_u` and `speed_v` as well as the `speed_x` and `speed_y` 

changes have been made to accommodate these modifications

## Issue(s) addressed

Resolves #1500 

## Impact

correctly decoded balloonsondes, must confirm if the `speed_x` and `speed_y` are now corrected to be the longitudinal and latitudinal winds

## Checklist

### DIFF
--- a/src/conventional/balloon_sonde_json2ioda.py
+++ b/src/conventional/balloon_sonde_json2ioda.py
@@ -124,12 +124,20 @@ def main(args):
         file = json.load(open(file_name))
 
         if file_cnt == 0:
+            observation_keys = file['observations'][0].keys()
+            if 'speed_u' in observation_keys:
+                speed_eastward = 'speed_u'
+                speed_northward = 'speed_v'
+                observation_keys = [x for x in observation_keys if x not in ['speed_x', 'speed_y']]
+            else:
+                speed_eastward = 'speed_y'
+                speed_northward = 'speed_x'
             # Create data frame to store each file's Meta Dataa
             df_metaData_files = pd.DataFrame(columns=file.keys())
             df_metaData_files.drop(['observations'], axis=1, inplace=True)
 
             # Create data frame to store all observation data
-            obs_data = pd.DataFrame(columns=file['observations'][0].keys())
+            obs_data = pd.DataFrame(columns=observation_keys)
             obs_data['releaseTime'] = None
             obs_data['stationElevation'] = None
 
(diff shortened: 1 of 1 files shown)


src/conventional/balloon_sonde_json2ioda.py
```diff
@@ -124,21 +124,29 @@ def main(args):
         file = json.load(open(file_name))
 
         if file_cnt == 0:
+            observation_keys = file['observations'][0].keys()
```
BenjaminRuston, line 127: we may want to create a fixed list of keys if they continue to modify the files without notification

the issue arises if we are trying to combine multiple 15 min files and these keys change between files.  This recently happened when decoding data from Mar22, 2024 and all the files forward now have multiple "speeds"

BenjaminRuston:
@smaticka this does warrant a thorough test most concerned with this line that requires the arrays to be in the order of the `obs_data.keys()`.  Since these keys were renamed think we should be ok, but again should be certain.

https://github.com/JCSDA-internal/ioda-converters/blob/bd74b4809339fadacbfca63d900991fbda784580/src/conventional/balloon_sonde_json2ioda.py#L184

smaticka (approved):
Thanks @BenjaminRuston. 

this worked for me when I ran the converter together on 2 files - 1 with the old obs key list (i.e. only `speed_x` and `speed_y`) and the 2nd with the new (i.e. both `speed_x` and `speed_y` and `speed_v` and `speed_u`). 

I agree that it is dependent on their obs keys staying the same.

smaticka:
> @smaticka this does warrant a thorough test most concerned with this line that requires the arrays to be in the order of the `obs_data.keys()`. Since these keys were renamed think we should be ok, but again should be certain.
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/bd74b4809339fadacbfca63d900991fbda784580/src/conventional/balloon_sonde_json2ioda.py#L184

This shouldn't be a problem since we define the order of the columns here: https://github.com/JCSDA-internal/ioda-converters/pull/1501/files#diff-dbe4cc8b28ed4b9fdc60d3e49d64bd91b7547c9ff88c4de04e7ec35364139bdbL140-L146

So the order won't change if they change their data again. Other things will break though if they change :) (like the variables we search for and store as our own, and as you found, the compatibility when converted 2 versions at the same time)

fcvdb (approved):
Thanks Ben!

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

PR #1509, opened 2024-04-30
add sample_flags for osw

### LINKED ISSUE #1508: sample flags for osw
## Description
the NOAA processed OSW product has a sample_flags that should be used in the PreQC 

## Requirements

pass the `poor_overall_quality_flag` into the `PreQC` for the data

the quality flag is passed into the IODA output

BenjaminRuston:
## Description

The NOAA CYGNSS ocean surface winds provide a sample_flags that should be used to flag `poor_overall_quality_flag`

A change in the list generation is also added this speeds up the converter by 10x or so

## Issue(s) addressed

Resolves #1508 

## Impact

A preQC value is passed from data to IODA file

## Checklist


fcvdb (approved):
Thanks Ben!

smaticka (commented):
Thanks @BenjaminRuston ! This works for me and is much faster with the comprehension adjustments.

for the bit labeling. they note in the headers that these flags also exist (below) do you think we should add those into the check as well?

And yes, I agree, the `snr` could be looked at as a possible additional PreQC. 

```128 = poor_quality_wind_sample: set when an unrealistic wind speed sample is detected.
64 = poor_quality_wind_nst: low confidence in the reported wind speed WHILE the star tracker flag (i.e. nst_att_status) is set```

**Comment by BenjaminRuston** [comment]
adding a `--qualitycontrol` that then drops data that have preQC flag set

**Comment by BenjaminRuston** [comment]
@smaticka think this is all good, 

> they note in the headers that these flags also exist (below) do you think we should add those into the check as well?

the bit1 flag includes those later flags, the comment on the flag notes this in `f['sample_flags'].attrs['comment']`:
```
1 = poor_overall_quality: logical OR of poor_quality_wind_nst, poor_quality_wind_sample, 
```

> And yes, I agree, the `snr` could be looked at as a possible additional PreQC.

I've consulted with the CYGNSS team at NOAA and they recommend just using the bit1 from the `sample_flags`

**Comment by BenjaminRuston** [comment]
thanks @smaticka this `drop` is now working 
https://github.com/JCSDA-internal/ioda-converters/blob/9085b0b9a78d3af9d9450e1aed374fd9e491be8a/src/hdf5/osw_2ioda.py#L309

the is now used with the `--qualitycontrol` option strict qc is enforced and the resulting file removes the points that do not pass `PreQC`:

File: src/hdf5/osw_2ioda.py [hunk 1]
```diff
@@ -318,19 +338,23 @@ def quality_control(obs_data):
                           help='base date for the center of the window')
     optional.add_argument('--debug', action='store_true',
                           help='enable debug messages')
-    optional.add_argument('--verbose', action='store_true',
-                          help='enable verbose debug messages')
+    optional.add_argument('-q', '--qualitycontrol', dest='qc_strict',
+                          default=False, action='store_true', required=False,
+                          help='turn on quality control georeality checks')
```
smaticka: Not sure what georeality check is, Is this more of a PreQC? It's a little confusing given there there is `disableQC` in Simobs after the simulation.

BenjaminRuston: this is similar to other data converter... 

when this option is present in the converter is applies strict QC to the data dropping the *bad* data points

the `disableQC` in the SIMOBS is hardcoded to look for `EffectiveQC` (not `PreQC`) which is what is assigned through Skylab (and generally assimilation)

BenjaminRuston:
```suggestion
                          help='turn on strict quality control checks dropping these points on output')
```

smaticka (approved):
thanks @BenjaminRuston !

this worked for me using:
`python3 $JEDI_SRC/iodaconv/src/hdf5/osw_2ioda.py -i ./cyg.ddmi.s20201215-000001-e20201215-235959.l2.wind_trackgridsize25km_NOAAv1.2_L1a21.d21.nc -o test.nc --qualitycontrol`

src/hdf5/osw_2ioda.py
```diff
@@ -229,25 +232,28 @@ def get_data_from_file(afile, col_names, osw_source, file_name):
...
-        windSpeedPreQC = [1 - afile['wind_confidence'][ii] for ii in range(len(afile['wind_confidence']))]
-        windSpeedObsError = [afile['wind_std'][ii] for ii in range(len(afile['wind_std']))]
+        latitude = [v for v in afile['sp_lat']]
+        longitude = [v for v in afile['sp_lon']]
+        dateTime = [int(v) for v in afile['sample_time']]  # datetime with different ref time
+        windSpeed = [v for v in afile['wind']]
+        windSpeedPreQC = [1 - v for v in afile['wind_confidence']]
+        # Data with 0 can be assumed to be ice free
+        qualIceFlag = [int(v > 0) for v in afile['quality_ice_flag']]
+        windSpeedPreQC += qualIceFlag
```
BenjaminRuston: this may be superfluous and included in the `wind_confidence` flag

PatNichols (approved):
Looks good

PatNichols:
@BenjaminRuston Is there a particular epic for this one?

src/hdf5/osw_2ioda.py
BenjaminRuston: it is not included this has been verified so it has been added for `strict_qc`

PR #1514, opened 2024-05-03
merge smap9km modifications into original converter

### LINKED ISSUE #1513: combine smap converters into single routine
## Description

maintenance of two nearly identical converters is not advantageous combine into a single routine for the SMAP surface soil moisture data 

## Requirements

SMAP surface soil moisture input files

successfully reproduces the output from two input data streams

BenjaminRuston:
build-group=https://github.com/JCSDA-internal/ioda/pull/1270

## Description

merge the changes from the 9km smap surface soil moisture converter into the original

## Issue(s) addressed

Resolves #1513 

## Impact

single converter for two flavors of SMAP surface soil moisture files

## Dependencies

- [x] https://github.com/JCSDA-internal/ioda/pull/1270

## Checklist


src/land/smap_ssm2ioda.py
```diff
@@ -149,6 +158,42 @@ def _read(self):
         DimDict['Location'] = len(self.outdata[('dateTime', 'MetaData')])
 
 
+def get_observation_time(filename, ncd, vals):
+    # get observation time from file if present fallback to extraction from filename
+    times = np.empty_like(vals, dtype=np.int64)
+    if 'tb_time_seconds' in ncd.variables.keys():
+        refsec = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()
+        for i in range(len(vals)):
+            dt = j2000_base_date + timedelta(seconds=int(refsec[i]))
```
BenjaminRuston:
```suggestion
        for isec in refsec:
            dt = j2000_base_date + timedelta(seconds=isec))
```

src/land/smap_ssm2ioda.py
```diff
@@ -149,6 +158,42 @@ def _read(self):
         DimDict['Location'] = len(self.outdata[('dateTime', 'MetaData')])
 
 
+def get_observation_time(filename, ncd, vals):
+    # get observation time from file if present fallback to extraction from filename
+    times = np.empty_like(vals, dtype=np.int64)
+    if 'tb_time_seconds' in ncd.variables.keys():
+        refsec = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()
+        for isec in refsec:
```
BenjaminRuston:
```suggestion
        for i, isec in enumerate(refsec):
```

src/land/smap_ssm2ioda.py
```diff
@@ -149,6 +158,42 @@ def _read(self):
...
 
 
+def get_observation_time(filename, ncd, vals):
+    # get observation time from file if present fallback to extraction from filename
+    times = np.empty_like(vals, dtype=np.int64)
+    if 'tb_time_seconds' in ncd.variables.keys():
+        refsec = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()
+        for i, isec in enumerate(refsec):
+            dt = j2000_base_date + timedelta(seconds=isec))
+            times[i] = round((dt - epoch).total_seconds())
```
BenjaminRuston:
```suggestion
        times = [round((j2000_base_date + timedelta(seconds=isec) - epoch).total_seconds()) for isec in refsec:

```

will not try this just yet but would like to simplify this if possible

src/land/smap_ssm2ioda.py
```diff
@@ -149,6 +158,42 @@ def _read(self):
         DimDict['Location'] = len(self.outdata[('dateTime', 'MetaData')])
 
 
+def get_observation_time(filename, ncd, vals):
+    # get observation time from file if present fallback to extraction from filename
+    times = np.empty_like(vals, dtype=np.int64)
+    if 'tb_time_seconds' in ncd.variables.keys():
+        refsec = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()
+        for i, isec in enumerate(refsec):
+            dt = j2000_base_date + timedelta(seconds=isec))
```
BenjaminRuston:
```suggestion
            dt = j2000_base_date + timedelta(seconds=isec)
```

src/land/smap_ssm2ioda.py
```diff
@@ -131,15 +135,19 @@ def _read(self):
...
         self.outdata[('dateTime', 'MetaData')] = times
         self.varAttrs[('dateTime', 'MetaData')]['units'] = iso8601_string
         self.outdata[('latitude', 'MetaData')] = lats
         self.outdata[('longitude', 'MetaData')] = lons
         self.outdata[('depthBelowSoilSurface', 'MetaData')] = deps
         self.varAttrs[('depthBelowSoilSurface', 'MetaData')]['units'] = 'm'
+        if sflg_present:
+            self.outdata[('surfaceQualifier', 'MetaData')] = sflg
+            self.varAttrs[('surfaceQualifier', 'MetaData')]['units'] = 'unitless'
+            self.outdata[('vegetationOpacity', 'MetaData')] = vegop
```
YoulongXia-NOAA, line 143: Remove 'unitless' sentence as this is old format. The new format will be default without unitless. right?

YoulongXia-NOAA:
@BenjaminRuston, Thank you for your effort. In particular for using ncks to reduce h5 file size into a small size file. I checked NRT smap soil moisture file and the format and variable names for both 9km SMAP and NRT SMAP files are the same. It makes sense to have a converter. As soil moisture converter is tested and used in land DA workflow only (by Darper's previous staff), it still need to tested within land DA workflow in the future.

src/land/smap_ssm2ioda.py
```diff
@@ -149,6 +161,42 @@ def _read(self):
...
+    # get observation time from file if present fallback to extraction from filename
+    times = np.empty_like(vals, dtype=np.int64)
+    if 'tb_time_seconds' in ncd.variables.keys():
+        refsec = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()
+        for i, isec in enumerate(refsec):
+            dt = j2000_base_date + timedelta(seconds=isec)
+            times[i] = round((dt - epoch).total_seconds())
+    else:
+        # get datetime from filename
+        file_refTime = re.search(r"\d{8}T\d{6}", filename).group()
```
BenjaminRuston: this should also be last occurrence of this match

BenjaminRuston:
```suggestion
        file_refTime = re.search(r"(\d{8}T\d{6})(?!.*\d{8}T\d{6})", filename).group()
```

BenjaminRuston:
you nailed it @ClaraDraper-NOAA the EASE parameters are in the *old* file as well:
DIFFER : VARIABLE : easeColumnIndex : DOES NOT EXIST IN "../../test/testoutput/smap_ssm.nc"
DIFFER : VARIABLE : easeRowIndex : DOES NOT EXIST IN "../../test/testoutput/smap_ssm.nc"
DIFFER : VARIABLE : surfaceQualifier : DOES NOT EXIST IN "../../test/testoutput/smap_ssm.nc"
DIFFER : VARIABLE : vegetationOpacity : DOES NOT EXIST IN "../../test/testoutput/smap_ssm.nc"

I will update the `testoutput` file and this should not break anything for downstream users as they are requesting things by name as they need them extra bits are just ignored

YoulongXia-NOAA:
@BenjaminRuston, that is great. As these two ioda converters are old not updated foe a long time. This is a great opportunity to get it updated. Thank you.

src/land/smap_ssm2ioda.py
```diff
@@ -149,6 +161,42 @@ def _read(self):
...
 
 
+def get_observation_time(filename, ncd, vals):
+    # get observation time from file if present fallback to extraction from filename
+    times = np.empty_like(vals, dtype=np.int64)
+    if 'tb_time_seconds' in ncd.variables.keys():
+        refsec = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()
+        for i, isec in enumerate(refsec):
+            dt = j2000_base_date + timedelta(seconds=isec)
+            times[i] = round((dt - epoch).total_seconds())
```
BenjaminRuston:
```suggestion
            times = [round(((j2000_base_date + timedelta(seconds=isec)) - epoch).total_seconds()) for isec in ncd.groups['Soil_Moisture_Retrieval_Data'].variables['tb_time_seconds'][:].ravel()]
```

untested but any advantage?

BenjaminRuston:
I'm about done reconciling the differences... Here is what I've found and it relates entirely to the `smap9km_ssm2ioda.py`

the *new* output file from the decoder looks correct and has these features. The right side entries.

```
DIFFER : VARIABLE : vegetationOpacity : TYPE : INT <> FLOAT
DIFFER : TYPES : ATTRIBUTE : _FillValue : VARIABLE : vegetationOpacity : INT <> FLOAT
```

the vegetationOpacity should be a `FLOAT` this is correct correspondingly the values should be updated. Again the *new* files is the right entry:

```
vegetationOpacity      /MetaData
DIFFER : VARIABLE : vegetationOpacity : POSITION : [0] : VALUES : 0 <> 0.0423718 : PERCENT : 100
DIFFER : VARIABLE : vegetationOpacity : POSITION : [1] : VALUES : 0 <> 0.277997 : PERCENT : 100
DIFFER : VARIABLE : vegetationOpacity : POSITION : [26335] : VALUES : 0 <> 0.201079 : PERCENT : 100
```

so these will correct

BenjaminRuston:
@ClaraDraper-NOAA where I have a question is regarding these values what should be used here.  

`smap_ssm2ioda.py` ==> NRT file (`testinput/SMAP_L2_SM_P_NRT_24342_A_20190822T224858_N16023_002.h5`)
`smap9km_ssm2ioda.py` ==> other file (`testinput/SMAP_L2_SM_P_E_36365_D_20211122T013945_R18240_001.h5`)

The NRT file uses 0.04 * soil_moisture
The other file from `smap9km_ssm9ioda.py` had a constant of 0.04. 

This update makes them consistent and they are now both `0.04 * soil_moisture` is this correct?

```
soilMoistureVolumetric /ObsError
DIFFER : VARIABLE : soilMoistureVolumetric : POSITION : [0] : VALUES : 0.04 <> 0.00646782 : PERCENT : 83.8304
...
DIFFER : VARIABLE : soilMoistureVolumetric : POSITION : [26335] : VALUES : 0.04 <> 0.0016456 : PERCENT : 95.886
```

BenjaminRuston:
then the error flags are altered between the two data types.   The `NRT` file does not modify the `PreQC` values which come from the `ncd.groups['Soil_Moisture_Retrieval_Data'].variables['retrieval_qual_flag']`

while the `smap9km_ssm2ioda.py` in the same loop they modify the `ObsError/soilMoistureVolumetric` also modifies the quality flags setting those *greater than 5* to `0`:

```
        for i in range(len(lons)):
            if vals[i] > 0.0:
                # assumed 4% SM rather than -999.0
                errs[i] = 0.04*vals[i]
                if qflg[i] > 5:
                    qflg[i] = 0
                else:
                    qflg[i] = 1
            else:
                qflg[i] = 1
```


is this alteration of the quality flag also acceptable? is the `PreQC` even used?

BenjaminRuston:
@srherbener and @gthompsnJCSDA this converter produced a file with units for `latitude` of `degrees_north` while the old file was `degree_north` 

the current ObsSpace.yaml convention has `degree_north`

first off where is this converter getting this from???? It's not being explicitly set for it is coming from the writer?

gthompsnJCSDA:
@BenjaminRuston  Singular and plural versions are interchangeable.  If the units are not explicitly set in the converter, then I am guessing the ioda (engines perhaps) is doing the assignment of those units by default.  I believe it is better to ensure the converter sets units for all appropriate variables rather than letting another program do it mysteriously.

BenjaminRuston:
@gthompsnJCSDA ... duh!!!

just wrote the units in the converter:
```
        self.varAttrs[('latitude', 'MetaData')]['units'] = 'degree_north'
        self.varAttrs[('longitude', 'MetaData')]['units'] = 'degree_east'
```

so that fixed that and now match convention... thanks!

@srherbener why does the `writer.BuildIoda` set `degrees_north` do you know where to change that?  We should have it match convention

test/CMakeLists.txt
```diff
@@ -2297,7 +2297,7 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_smap9km_ssm
                     COMMAND bash
                     ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
                             netcdf
-                            "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/smap9km_ssm2ioda.py
+                            "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/smap_ssm2ioda.py
                             -i testinput/SMAP_L2_SM_P_E_36365_D_20211122T013945_R18240_001.h5
                             -o testrun/smap9km_ssm.nc
```
BenjaminRuston, line 2302: this is also our chance to change these names... I still don't follow what the difference between the `9km` test which does NOT use the near real time `NRT` file

should we rename the `smap9km` to something else?  

or is it the other way around should we name these files and this test just `smap_ssm` and the other test using the `NRT` file the smap_NRT_ssm ?

I can do that if you like

src/land/smap_ssm2ioda.py
YoulongXia-NOAA: This should work

test/CMakeLists.txt
YoulongXia-NOAA: @BenjaminRuston, based on my understandings, SMAP 9km data lag from real-time is larger than NRT SMAP product. But NRT SMAP soil moisture product has coarse resolution (32 km) and low quality when compared 9km product. As NRT product is quickly check for data quality. However, for operational work, time is important as large lag product cannot be used for ops but it is good for research. This is my two cents.

YoulongXia-NOAA: @BenjaminRuston, as this converter can be used for both 9km and NRT SMAP soil moisture, just keep one input netcdf file and ioda output file using only smap. The users can decide either they use smap 9km data or NRT data. What do you think?

BenjaminRuston:
From @ClaraDraper-NOAA :

> The error should be constant at 0.04 m3/m3 (or 4%), and *not* 4% of the obs.

awesome thanks

YoulongXia-NOAA:
> then the error flags are altered between the two data types. The `NRT` file does not modify the `PreQC` values which come from the `ncd.groups['Soil_Moisture_Retrieval_Data'].variables['retrieval_qual_flag']`
> 
> while the `smap9km_ssm2ioda.py` in the same loop they modify the `ObsError/soilMoistureVolumetric` also modifies the quality flags setting those _greater than 5_ to `0`:
> 
> ```
>         for i in range(len(lons)):
>             if vals[i] > 0.0:
>                 # assumed 4% SM rather than -999.0
>                 errs[i] = 0.04*vals[i]
>                 if qflg[i] > 5:
>                     qflg[i] = 0
>                 else:
>                     qflg[i] = 1
>             else:
>                 qflg[i] = 1
> ```
> 
> is this alteration of the quality flag also acceptable? is the `PreQC` even used?

@BenjaminRuston, use smap9km_ssm2ioda.py as a right converter without change PreQC as did in smap9km_ssm2ioda.py as it is correct one. The smap_ssm2ioda.py is a very old version. Thank you.

src/land/smap_ssm2ioda.py
YoulongXia-NOAA: If this suggested change works, that is a good effort to simplify the code. Do your test.

BenjaminRuston:
thank you @YoulongXia-NOAA for the help sorting this out.  This should be all set to go now

test/CMakeLists.txt
BenjaminRuston: in a new issue and PR will take on removal of defunct routine `smap9km_ssm2ioda.py` and the change names of test files to be either plan `smap_ssm` or `smap_NRT_ssm` to distinguish

BenjaminRuston:
@YoulongXia-NOAA please have a look this is ready

src/land/smap_ssm2ioda.py
```diff
@@ -89,8 +93,9 @@ def _read(self):
         errs = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['soil_moisture_error'][:].ravel()
         qflg = ncd.groups['Soil_Moisture_Retrieval_Data'].variables['retrieval_qual_flag'][:].ravel()
 
+        times = get_observation_time(self.filename, ncd, vals)
+        sflg_present, sflg, vegop, erowi, ecoli = get_ease_surface_param(ncd)
```
YoulongXia-NOAA, line 97: Could you please check if vals is used? I am feeling that in this get_observation_time does not used vals. Am I right?

YoulongXia-NOAA:
> @YoulongXia-NOAA please have a look this is ready

@BenjaminRuston, please remove the sentence 

 self.varAttrs[('surfaceQualifier', 'MetaData')]['units'] = 'unitless'

and check if vals is used in 
 get_observation_time(self.filename, ncd, vals)

The others are good. If you fix them, I will approve this PR. Thank you.

BenjaminRuston:
@YoulongXia-NOAA yes `vals` is used in the `get_observation_time` function for the case when the `tb_time_seconds` is not in the variable keys

BenjaminRuston:
regarding the removal of the attribute for `surfaceQualifier` if this line is removed:
https://github.com/JCSDA-internal/ioda-converters/blob/7341b0c27de96127b2391b2f386f1104cd300961/src/land/smap_ssm2ioda.py#L143

the error is then generated:
```
DIFFER : VARIABLE "surfaceQualifier" IS MISSING ATTRIBUTE WITH NAME "units" IN FILE 
```

@gthompsnJCSDA what is the unit to use for this?  It is a unitless integer ?

YoulongXia-NOAA:
@BenjaminRuston @gthompsnJCSDA, I checked ObsSpace.yaml, its definition is below:
 - Variable: [ "surfaceQualifier", "surface_type" ]  # 0=land, 1=sea, 2=seaice
    MetaData: true
    Type: Enum
It is the integer but without unit.  It should be a variable without unit attribute, right?

BenjaminRuston:
> @BenjaminRuston @gthompsnJCSDA, I checked ObsSpace.yaml, its definition is below:
> 
> * Variable: [ "surfaceQualifier", "surface_type" ]  # 0=land, 1=sea, 2=seaice
>   MetaData: true
>   Type: Enum
>   It is the integer but without unit.  It should be a variable without unit attribute, right?

@YoulongXia-NOAA think we may want to change the `ObsSpace.yaml` 

@gthompsnJCSDA for this `surfaceQualifier` do we want to change its `ObsSpace.yaml` entry to this:
```
    Force Units: false
    Type: UInt32
    MetaData: true
```

YoulongXia-NOAA:
> > @BenjaminRuston @gthompsnJCSDA, I checked ObsSpace.yaml, its definition is below:
> > 
> > * Variable: [ "surfaceQualifier", "surface_type" ]  # 0=land, 1=sea, 2=seaice
> >   MetaData: true
> >   Type: Enum
> >   It is the integer but without unit.  It should be a variable without unit attribute, right?
> 
> @YoulongXia-NOAA think we may want to change the `ObsSpace.yaml`
> 
> @gthompsnJCSDA for this `surfaceQualifier` do we want to change its `ObsSpace.yaml` entry to this:
> 
> ```
>     Force Units: false
>     Type: UInt32
>     MetaData: true
> ```

@BenjaminRuston, we should not change ObsSpace.yaml just removed that units sentence. However, you told me that when you remove that sentence, this brings the error.  I assumed that remove this sentence should not bring error. we need to solve why it produce the error, right?

YoulongXia-NOAA:
@BenjaminRuston, I may be wrong. If the change you suggested for obsSpace.yaml can solve this issue, it is great. Thank you.

YoulongXia-NOAA:
@BenjaminRuston, @gthompsnJCSDA, I checked ObsSpace.yaml file, for flag and type alike variable, they use
 MetaData: true
Type: Enum 

But for Index alike variables, they use
Type: UInt32
MetaData: true

I am not sure what differences between the two. However, both of them has no units definition as they has no unit.

BenjaminRuston:
thanks @YoulongXia-NOAA ..... now .... we've removed `unitless` and updated the `ObsSpace.yaml` please also review the IODA PR:
https://github.com/JCSDA-internal/ioda/pull/1270

and this is ready ;0)

BenjaminRuston:
btw, talked to @gthompsnJCSDA and he advised to leave this as 
```
Type: Enum
```

and just add the:

```
 Force Units: false
 ```
 
 so that is what I did

**Comment by YoulongXia-NOAA** [comment]
> btw, talked to @gthompsnJCSDA and he advised to leave this as
> 
> ```
> Type: Enum
> ```
> 
> and just add the:
> 
> ```
>  Force Units: false
> ```
> 
> so that is what I did

Okay, I approved this PR now.

**Review by YoulongXia-NOAA — APPROVED** [review]
look good now. I approve this PR.

**Comment by BenjaminRuston** [comment]
@haydenlj or @gthompsnJCSDA could you have a look over this PR

File: src/land/smap_ssm2ioda.py (hunk 9 above)
Comment (gthompsnJCSDA) [reply src/land/smap_ssm2ioda.py:97]: @YoulongXia-NOAA   It looks to me that the variable `vals` is passed to the routine only to provide the "template" (effectively the same vector size) as the output `times` variable.  So it really is only used by the numpy `full_like` method.

File: src/land/smap_ssm2ioda.py [hunk 10]
```diff
@@ -102,44 +107,42 @@ def _read(self):
...
-        # my_datetime = datetime.strptime(str_datetime, "%Y-%m-%d")
-        # get datetime from filename
-        file_refTime = re.search(r"\d{8}T\d{6}", self.filename).group()
-        file_refTime = datetime.strptime(file_refTime, "%Y%m%dT%H%M%S")
-        time_offset = round((file_refTime - epoch).total_seconds())
         vals = vals.astype('float32')
         lats = lats.astype('float32')
         lons = lons.astype('float32')
         deps = deps.astype('float32')
+        # reassign to contant (could be pass in)
```
gthompsnJCSDA: spelling typo: contant should be `constant`

gthompsnJCSDA (approved):
spelling typo to fix, but otherwise I approve.

src/land/smap_ssm2ioda.py
PatNichols:
```suggestion
        # reassign to constant (could be pass in)
```

BenjaminRuston: suggestion committed

src/land/smap_ssm2ioda.py
YoulongXia-NOAA: Thank you for your explanation @gthompsnJCSDA :)

YoulongXia-NOAA:
Thank you, @BenjaminRuston. You made it done.

PR #1519, opened 2024-05-23
Update converter for last TEMPO public release

jeromebarre:
## Description

Update converter for last TEMPO public release. Add the `-v3` option for V3 NRT NO2 product.

## Issue(s) addressed

N/A

## Dependencies

- [ ] ...

## Impact

## Checklist

- [] I have made corresponding changes to the documentation


src/compo/tempo_nc2ioda.py
```diff
@@ -41,12 +41,13 @@
...
 
 class tempo(object):
-    def __init__(self, filenames, varname, columnType, qa_flg, thin, obsVar):
+    def __init__(self, filenames, varname, columnType, qa_flg, thin, nrt, obsVar):
         self.filenames = filenames
         self.varname = varname
         self.columnType = columnType
         self.qa_flg = qa_flg
         self.thin = thin
+        self.nrt = nrt
```
ashley314: @jeromebarre I tested this out but had to update all instances of `nrt` to `v3`. Is the intent for this PR to use `v3`?

jeromebarre: Oh sorry about that. I learned that V3 will be reprocessed too and won't only be 
NRT. So that's why I changed the option to` -v3`

ashley314: Okay great thanks! Yeah using `-v3` makes sense. Do you want me to update this PR or will you?

jeromebarre:
Updated the branch and testing this with:
https://github.com/JCSDA-internal/ewok/tree/feature/tempov3 I don't really think we need to work out a whole logic for v3 vs v2 as I was made aware that v3 will be the main version with reprocessed dates in the future. (not sure when...)

jeromebarre:
ran now fine in the ingest suite

ashley314 (approved):
Thanks, this worked for the v3 data from yesterday using https://github.com/JCSDA-internal/ewok/pull/989 for the EWOK ingest suite. 

image

mer-a-o (approved):

src/compo/tempo_nc2ioda.py
```diff
@@ -338,10 +359,14 @@ def main():
         "0 normal, 1 suspicious, 2 bad",
         type=int, default=0)
     optional.add_argument(
-        '-n', '--thin',
+        '-t', '--thin',
```
BenjaminRuston, line 362: just make a note but we have been removing the shorthand flags in our converters making the explicit `--thin` required to make them more obvious

BenjaminRuston (approved):
thank you @jeromebarre this is nice to have the backward compatibility kept for now as well

PR #1522, opened 2024-06-11
Update VIIRS L1b Albedo converter

weiwilliam:
## Description
This PR adds the needed sensorViewAngle and sensorScanPosition in VIIRS L1b IODA file for UFO CRTM passive radiance simulation.

We may need to revisit whether the stored reflectance need to convert to true reflectance with divide by cosine of solar zenith angle at Line 188.

## Dependencies
No dependencies

## Impact
No impacts on other repos

## Checklist


src/compo/viirs_l1bnc2ioda.py
```diff
@@ -170,7 +183,9 @@ def _read(self):
                 err = obsgrp.variables[errname]
                 err.set_auto_scale(False)
 
-                vals[:, ichan] = obs[:].data.ravel()
+                # NASA VIIRS stored reflectance need to divide by
+                # cosine of solar zenith angle to get true reflectance
+                vals[:, ichan] = obs[:].data.ravel() / np.cos(solar_za * deg2rad)
```
gthompsnJCSDA: For today, I think you should place this line within a logical/boolean IF-test and set a true/false variable near the top of the code to be `False` for right now.  The CRTM-UFO code for visible reflectance is currently NOT dividing by `cosine(solarZenithAngle)`.  At a later time, I believe we need to include a YAML option for turning on or off this division in UFO.

gthompsnJCSDA (commented):
See my comment about `cosine(solarZenithAngle)` - which could be temporary.

src/compo/viirs_l1bnc2ioda.py
weiwilliam: @gthompsnJCSDA Thank you, I will update it.

weiwilliam:
@gthompsnJCSDA I have added an input argument to decide whether the obsvals should multiply the secant of the solar zenith angle.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -158,6 +172,12 @@ def _read(self):
                 orbit_ad[:] = 1
             self.varAttrs['satelliteAscendingFlag', metaDataName]['description'] = '0=descending, 1=ascending'
 
+            # NASA VIIRS stored reflectance need to divide by
+            # cosine of solar zenith angle to get true reflectance
+            sec_term = 1.
+            if self.divide_cos_solarza:
+                sec_term = np.cos(solar_za * deg2rad) ** -1
```
gthompsnJCSDA: I am pretty sure that `1.0/a` operates on computers faster than `a**-1` so I would suggest making it a division not power.  Otherwise, I am OK with the changes.

weiwilliam: It used to be 1. / np.cos(...), but it triggered a Runtime Warning of invalid value in true_divide (something like that).
I printed the solar_za, but there was no NaN.

Interestingly, I can't reproduce it now (on s4).
I'll bring it back.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -146,6 +157,9 @@ def _read(self):
             solar_aa = geo_ncd.groups['geolocation_data'].variables['solar_azimuth'][:].data.ravel()
             sensor_za = geo_ncd.groups['geolocation_data'].variables['sensor_zenith'][:].data.ravel()
             sensor_aa = geo_ncd.groups['geolocation_data'].variables['sensor_azimuth'][:].data.ravel()
+            sin_va = np.sin(sensor_za*deg2rad) * earth_radius / (earth_radius + orbit_height)
+            sensor_va = np.arcsin(sin_va) * rad2deg
```
BenjaminRuston: ideally this should be a function as it is generic for all satellite observations that need to derive this

let me get back to you shortly on this @weiwilliam

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -158,6 +172,12 @@ def _read(self):
                 orbit_ad[:] = 1
             self.varAttrs['satelliteAscendingFlag', metaDataName]['description'] = '0=descending, 1=ascending'
 
+            # NASA VIIRS stored reflectance need to divide by
+            # cosine of solar zenith angle to get true reflectance
+            sec_term = 1.
+            if self.divide_cos_solarza:
+                sec_term = 1. / np.cos(solar_za * deg2rad)
```
BenjaminRuston: this is a main change correct as this will change some of the output values for the reflectance

it seems you updated the `testOutput` however it is not matching what is produced, does this need to be updated again?

image




  also the `sensorViewAngle` is now computed and added correct

BenjaminRuston: that CI output is the Intel CDASH result

weiwilliam: I noticed that, and I just realized that I have to update the CMakeFile for test.
I didn't add the -s argument in there.
Let me update it and see how it works.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -36,8 +38,12 @@
 # VIIRS M-band 11 reflective channels central wavelength. Do not change list below.
 wavelength = [0.412, 0.445, 0.488, 0.555, 0.672, 0.746,
               0.865, 1.24, 1.378, 1.61, 2.25]
+orbit_height = 829.       # in km
+earth_radius = 6378.137   # in km, mean earth radius
 speed_light = 2.99792458E8
 frequency = speed_light*1.0E6/np.array(wavelength)
+deg2rad = np.pi/180.
```
jeromebarre: Couldn't you just use pi in a few places you need this?

src/compo/viirs_l1bnc2ioda.py
jeromebarre: yes would be good to have something like a util.py in tool or something like that,

jeromebarre:
Looks good to me I haven't tested it.

src/compo/viirs_l1bnc2ioda.py
weiwilliam: Yes, I can update it.

weiwilliam:
Here is a sample image processed for 2021 August 5 04-05Z.
Used ncks to pick every 70 points.
image

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -273,13 +301,21 @@ def main():
         help="percentage of random thinning fro 0.0 to 1.0. Zero indicates"
         " no thinning is performed. (default: %(default)s)",
         type=float, default=0.0)
+    parser.add_argument(
+        '-s', '--secterm',
```
BenjaminRuston: recommend removing the `-s` option shortcut has been our standard practice as they can be too vague
```suggestion
        '--secterm',
```

weiwilliam: Sure, thanks. I have an impression that every argument is going to be in long name rather than the shortcut. @BenjaminRuston do you want me to update other arguments?

BenjaminRuston: yes please do so ...

see there is are two input files.... would say the `-o` is about the only option shortcut that could remain, and personally `obsinfo` and `geoinfo` would use `inputObs` and `inputGeo` as the long names but this is semantics and that change should be done separately from this PR 

so in short you could leave the output on `-o` but for the other remove the shortcut option name
also please see if my change works,, you can then merge into this branch... did you have a couple input files so I could run this too

BenjaminRuston: and I'm continuing to use your test files and that should be fine..

but did you notice in this program the channel number is not being recorded correctly:
```
>>> f.keys()
<KeysViewHDF5 ['Channel', 'Location', 'MetaData', 'ObsError', 'ObsValue', 'PreQC']>
>>> f['Channel'][:]
array([ 1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11], dtype=int32)
```

BenjaminRuston: wait this is fixed ;0)

test/CMakeLists.txt
```diff
@@ -1367,6 +1367,7 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_viirs_j1_l1b_albedo
                           "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/viirs_l1bnc2ioda.py
                           -i testinput/viirs_j1_l1b_albedo_obsval_202108050600.nc
                           -g testinput/viirs_j1_l1b_albedo_geoloc_202108050600.nc
+                          --secterm Y
```
BenjaminRuston: we can do these in a single change now can remove this `Y`
```suggestion
                          --secterm
```

weiwilliam: Yes, I'll remove it.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -273,13 +296,17 @@ def main():
         help="percentage of random thinning fro 0.0 to 1.0. Zero indicates"
         " no thinning is performed. (default: %(default)s)",
         type=float, default=0.0)
+    parser.add_argument(
+        '--secterm',
+        help="multiply sec of solar zenith angle to get true reflectance (Y/N)",
```
BenjaminRuston: and should change this comment
```suggestion
        help="presence of option will multiply secant of solar zenith angle to get true reflectance",
```

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -158,6 +166,12 @@ def _read(self):
                 orbit_ad[:] = 1
             self.varAttrs['satelliteAscendingFlag', metaDataName]['description'] = '0=descending, 1=ascending'
 
+            # NASA VIIRS stored reflectance need to divide by
+            # cosine of solar zenith angle to get true reflectance
+            sec_term = 1.
+            if self.divide_cos_solarza:
+                sec_term = 1. / np.cos(solar_za * np.pi / 180.)
```
BenjaminRuston:
```suggestion
            sec_term = 1. / np.cos(solar_za * np.pi / 180.) if self.divide_cos_solarza else 1.
```

weiwilliam: This change didn't work, so I updated the code to the traditional if-else structure.
Maybe that's the thing you suggested? @BenjaminRuston

jeromebarre (approved):
Looks good now. Thanks!

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -158,6 +166,13 @@ def _read(self):
                 orbit_ad[:] = 1
             self.varAttrs['satelliteAscendingFlag', metaDataName]['description'] = '0=descending, 1=ascending'
 
+            # NASA VIIRS stored reflectance need to divide by
+            # cosine of solar zenith angle to get true reflectance
+            if self.divide_cos_solarza:
```
gthompsnJCSDA: The `sec_term` variable could produce a floating point exception if solar_za is equal to 90 degrees exactly.  You may need to include a check for that value in the IF-statement for safety sake.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -223,19 +245,20 @@ def _read(self):
         # Write other MetaData
         self.varAttrs[('dateTime', metaDataName)]['units'] = iso8601_string
         output_chidx = np.array(channels, dtype=np.int32) - 1
-        self.outdata[('sensorCentralFrequency', metaDataName)] = np.array(frequency[output_chidx], dtype=np.float64)
+        self.outdata[('sensorCentralFrequency', metaDataName)] = np.array(frequency[output_chidx], dtype=np.float32)
```
gthompsnJCSDA, line 246: I am under the impression that `sensorCentralFrequecy` is a double-precision variable by the data conventions and should be kept double precision.  Why was this changed?  Can you double-check?

gthompsnJCSDA (dismissed):

src/compo/viirs_l1bnc2ioda.py
weiwilliam: Yeah, I got the error message below when I set it to double.
```
        Reason: Variable 'MetaData/sensorCentralFrequency' is not of any supported type
        source_column:  0
        source_filename:        /data/users/swei/Git/skylab/jedi-bundle/ioda/src/engines/ioda/include/ioda/Variables/VarUtils.h
        source_function:        auto ioda::VarUtils::forAnySupportedVariableType(const ioda::Variable &, const Action &, const ErrorHandler &) [with Action = lambda [](auto)->auto; ErrorHandler = ioda::VarUtils::ThrowIfVariableIsOfUnsupportedType]
        source_line:    128
        stacktrace:
```
Then I checked other test files in ufo-data and I found it should be in single precision.

src/compo/viirs_l1bnc2ioda.py
weiwilliam: Got it. 
I tried to run the calculation based on an array from 60 to 94 degrees and it doesn't trigger the floating point exception.
How about masking those points where solar_za is over 85?

src/compo/viirs_l1bnc2ioda.py
gthompsnJCSDA: OK, this is surprising to me, but let's move forward with single-precision. I will hit resolve now.

src/compo/viirs_l1bnc2ioda.py
gthompsnJCSDA: You could easily just say: `if self.divide_cos_solarza and solar_zenith_angle != 90.0:`
I believe that is easiest and safest.

src/compo/viirs_l1bnc2ioda.py
BenjaminRuston: this is the new standard syntax,,, there are AI tools to debug it ;0)

let me run it again and get this correct

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -158,6 +166,11 @@ def _read(self):
                 orbit_ad[:] = 1
             self.varAttrs['satelliteAscendingFlag', metaDataName]['description'] = '0=descending, 1=ascending'
 
+            # NASA VIIRS apply secant or divide reflectance by
+            # cosine of solar zenith angle to get true reflectance
+            sec_term = 1. / np.cos(self.solar_za * np.pi / 180.) if (
+                self.apply_secterm and self.solar_za != 90.) else 1.
```
BenjaminRuston:
```suggestion
            sec_term = 1. / np.cos(self.solar_za * np.pi / 180.) if (self.solar_za != 90.) else 1.
```

actually the cleanest way readibility

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -170,7 +183,7 @@ def _read(self):
                 err = obsgrp.variables[errname]
                 err.set_auto_scale(False)
 
-                vals[:, ichan] = obs[:].data.ravel()
+                vals[:, ichan] = obs[:].data.ravel() * sec_term
```
BenjaminRuston:
```suggestion
                vals[:, ichan] = obs[:].data.ravel() * sec_term if self.apply_secterm else obs[:].data.ravel()
```

BenjaminRuston: so break the conditionals into two sections one to compute secant exclusively

and one to apply it exclusively

there will likely be no performance hit as optimizers within the language would handle such a straightforward logic.  and it does make the readibility ever so slightly better

src/compo/viirs_l1bnc2ioda.py
weiwilliam: Okay, let the UFO filter screen out those points.
However, it does not trigger the error of division by zero because the value inside np.cos will not exactly equal 0.5pi.
6.123233995736766e-17 is the cosine value when solar_za is 90.
So, it should be fine.

src/compo/viirs_l1bnc2ioda.py
weiwilliam: It won't work because solar_za is an array, not a single value.

weiwilliam:
@BenjaminRuston @gthompsnJCSDA 
I would like to leave the sec_term calculation as it was because it will never be the case to get zero from np.cos with using np.pi inside.
I understand that the cosine of 0.5 pi is zero mathematically, but np.pi won't be exactly the mathematical pi.
So the np.cos(0.5 * np.pi) is very close to zero but not zero.
But I will take Ben's suggestion to update the code in the channel loop.

src/compo/viirs_l1bnc2ioda.py
gthompsnJCSDA: Shih-Wei.  Unfortunately, you cannot count on the value being E-17 or similar depending on architecture and compilers, etc.  I wouldn't trust what you found.  I urge the use of additional conditional to be safe.

src/compo/viirs_l1bnc2ioda.py
BenjaminRuston: yes ... so that is more reason to break it down

please see the update this to me is the most readable and clear in intent

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -170,7 +181,8 @@ def _read(self):
                 err = obsgrp.variables[errname]
                 err.set_auto_scale(False)
 
-                vals[:, ichan] = obs[:].data.ravel()
+                # NASA VIIRS apply secant if option enabled
+                vals[:, ichan] = obs[:].data.ravel() * sec_term if self.apply_secterm else obs[:].data.ravel()
```
BenjaminRuston, line 186: @weiwilliam separating the logic will help for clarity,, but now whoops need to add the location check for @gthompsnJCSDA

src/compo/viirs_l1bnc2ioda.py
weiwilliam: That is true if the compiler automatically zeros the small value, but I doubt it.
I can add a mask array rather than `solar_za != 90` because it does not work with an array.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -158,6 +166,9 @@ def _read(self):
                 orbit_ad[:] = 1
             self.varAttrs['satelliteAscendingFlag', metaDataName]['description'] = '0=descending, 1=ascending'
 
+            # secant of solar zenith angle to get true reflectance (apply_secterm option)
+            sec_term = np.where(solar_za != 90, 1. / np.cos(solar_za * np.pi / 180.), 1.)
```
BenjaminRuston: and done,, this is the way to handle the case for an array `solar_za`

BenjaminRuston: @weiwilliam please let me know if this doesn't work for you

I'm getting it to run with the `secterm` applied and matching the output file of the branch

weiwilliam: Thank you @BenjaminRuston, I'm also working on it.

BenjaminRuston: please have a look it makes more sense do a line with computes secant for all points,, set to 1 if solar zenith is 90-degrees

src/compo/viirs_l1bnc2ioda.py
BenjaminRuston: then only apply secant if requested by the user

this is clean and reproduces what you had before

src/compo/viirs_l1bnc2ioda.py
weiwilliam: @BenjaminRuston It works well and the output passed the nccmp.

gthompsnJCSDA:
@BenjaminRuston @weiwilliam  This conversation about possible divide by zero is going on way too long.  Remember, you could have an angle times PI that results in the division by zero.  It's not that PI will never be exact, but what about the angle being just the magic number needed so that the COS function gives the zero.  So perhaps the true safe way to work is testing for 90 +- epsilon to avoid any chance for floating point exception.

The one-line IF-tests are really, really hard to read compared to the original idea of `sec_term = 1.0` then reassign new value if using the option. That truly is far more readable if you ask me.

Otherwise we are wasting a lot of time just discussing such a minor issue.

weiwilliam:
@gthompsnJCSDA Agree, now I make it calculate sectant solar za when cosine za is not zero.
Then it should be able to avoid any magic numbers make cosine zero.

And I quickly gave it a try with 200,000 values between 89 to 91 degrees.
There is no zero value in np.cos results.

BenjaminRuston (approved):
this is the clearest version in many ways, minor but I think worth the effort thanks @weiwilliam

PR #1525, opened 2024-06-25
update routine to use globals from def_jedi_utils

BenjaminRuston:
## Description

Recommended changes to have some constants and functions come from centralized location


BenjaminRuston:
@weiwilliam can you try these changes?

the one I'm not sure about is the deg2rad getting in correctly, but if all works correctly should give you the same answer you're already getting... a lateral move

note too you can merge this into your branch

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -157,8 +153,7 @@ def _read(self):
             solar_aa = geo_ncd.groups['geolocation_data'].variables['solar_azimuth'][:].data.ravel()
             sensor_za = geo_ncd.groups['geolocation_data'].variables['sensor_zenith'][:].data.ravel()
             sensor_aa = geo_ncd.groups['geolocation_data'].variables['sensor_azimuth'][:].data.ravel()
-            sin_va = np.sin(sensor_za*deg2rad) * earth_radius / (earth_radius + orbit_height)
-            sensor_va = np.arcsin(sin_va) * rad2deg
+            sensor_va = compute_scan_angle(sensor_za, orbit_height, sensor_za)
```
BenjaminRuston: this function needs a cleanup btw,,, the first argument is not used... I will remove in a separate issue and PR and correct here, so it would look like `compute_scan_angle(orbit_height, sensor_za)`

BenjaminRuston: this is likely to occur after this PR is merged

weiwilliam (approved):

PR #1528, opened 2024-07-11
GMAO Converter Update

### LINKED ISSUE #1218: Gmao IODA_v3 converter is slow
**gmao_obs2ioda.py**(marine) takes exponentially longer to execute for files with increasing number of observation numbers. Also, there doesn't seem to be a ctest for the GMAO IODA conversion except for a validation test.

It starts happening after the updates from IODA_v2 to IODA_v3. IODA_v2 conversions are instant for the same files.

I conducted tests with multiple file types and two different versions of IODA_v3:

#### gmao-obs-2021071500-jason3.nc
Found a total number of observations: 23817
Elapsed time: 276.86 seconds

#### gmao-obs-2021070100-saral.nc
Found a total number of observations: 10972
Elapsed time: 58.64 seconds

#### gmao-obs-20180415.nc (this is in the test/testinput)
Found a total number of observations: 7701
Elapsed time: 22.15 seconds

Dooruk:
Previous GMAO converter code was ineffective and slow, this update resolves that issue.

## Issue(s) addressed

Resolves #1218 

## Checklist


BenjaminRuston (approved):
thank you for the cleanup @Dooruk looks like one of the main improvements besides readability is the replacement of an explicit loop with use of numpy `where`, any other nugget you'd like to pass on that gave significant performance improvement

Dooruk:
> the main improvements besides readability is the replacement of an explicit loop with use of numpy `where`, any other nugget you'd like to pass on that gave significant performance improvement

Yes, preallocating full arrays with the `_FillValue` helped a lot on the performance. I tested it with a SMAP file containing upwards of 100k observations and conversion was instant. 

I added @srherbener as a reviewer but i wasn't able to add @PatNichols for some reason, not sure who would be more appropriate for the approval.

src/marine/gmao_obs2ioda.py
```diff
@@ -146,91 +173,71 @@ def __init__(self, files_input, filename, date, obsList):
...
+            data[(key, metaDataName)] = np.full(totalObs, fill_value, dtype=dtypes[dtypestr])
+
+        print(f"Present variable(s) and their attribute(s) in this input: {presentVarDict}")
+
+        # Prefill data with missing values
+        # QC (preQC) value is zero for now (assume all data is good!)
+        for key, (_, variable, _, _, _) in presentVarDict.items():
+            data[(variable, obsValName)] = np.full(totalObs, float_missing_value, dtype=dtypes['float'])
+            data[(variable, obsErrName)] = np.full(totalObs, float_missing_value, dtype=dtypes['float'])
+            data[(variable, qcName)] = np.full(totalObs, 0, dtype=dtypes['integer'])
```
srherbener, line 206: This seems wrong to me, but I may be misunderstanding what is going on. It looks like on line 180 the _FillValue for PreQC is set to `int_missing_value` which I think is good. Looks like on line 64 `int_missing_value` is set to the default integer value which should be the maximum negative 32-bit integer value:  https://github.com/JCSDA-internal/ioda-converters/blob/3c437f464572ce8d5ab076b4ad713f7f7f24dd04/src/pyiodaconv/ioda_conv_engines.py#L51-L52

Then here you set all values in PreQC to zero. Since zero does not match the _FillValue (-2147483647) the ioda reader will not replace the zeros with the JEDI missing values and the UFO operators will be mislead into thinking the data provider for the obs data has marked all values with a zero quality value. Note that in lots of cases a zero quality value means the data is good. 

I think the missing values (`int_missing_value`) should remain in PreQC since PreQC is not given anything extracted from the data provider (ie, the input file to this converter).

My recommendation is to either remove this line (which will leave the missing values, -2147483647, in PreQC), or to replace `0` with `int_missing_value` in this line.

srherbener (dismissed):
Thanks for fixing the performance issues! I've got just the one comment about the PreQC variable.

srherbener:
Note that Pat Nichols has left JCSDA (which is why you can't add him to the reviewer list).

BenjaminRuston:
I'll add @travissluka as another set of eyes

BenjaminRuston:
@srherbener didn't this get cleaned up, can you review your change request and remove if now obsolete

Dooruk:
Following the [comments](https://github.com/JCSDA-internal/ioda-converters/issues/1218) on the QC flags, I will make some additional changes to this. Also I will add some fixes to the time dimension after we made some changes in the ODAS outputs. Sorry, didn't want to bother folks with more notifications until I made those changes and can't seem to change this to draft.

srherbener:
@Dooruk under the Reviewers list (right sidebar) there is a link you can click on to convert this to a draft PR. The link reads "Still in progress? Convert to draft". Or is the issue that you tried the link and you don't have permission to change the PR status? If it's the latter let me know and I can change this PR to a draft. Thanks!

Dooruk:
> @Dooruk under the Reviewers list (right sidebar) there is a link you can click on to convert this to a draft PR. The link reads "Still in progress? Convert to draft". Or is the issue that you tried the link and you don't have permission to change the PR status? If it's the latter let me know and I can change this PR to a draft. Thanks!

Didn't see that my bad, now it's merely a draft.

srherbener:
No worries, I always have to search around to find that link. They should make it more obvious.

src/marine/gmao_obs2ioda.py
Dooruk: So it turns out output from our ODAS processing does not contain QC flags. They are all assumed "good", hence the 0 flag. There is a basic boundary check in this converter script and I added the UFO QC flag "12" if that happens to be triggered. 

In that sense I feel like the code is working as intended, first creating the `_FillValue` attribute, and then assigning `0` for the full `nobs` dimension. But are you saying if I switch from 0 to `int_missing_value` would all of the obs still be used?

Dooruk:
This is ready for review again. One change I introduced is that now it uses the `datetime` information from the input data, which wasn't the case before.

Our GMAO/ODAS pre-processing is not as sophisticated to apply @gthompsnJCSDA's suggestions but maybe in the future we could once we improve upon it:) 

https://github.com/JCSDA-internal/ioda-converters/issues/1218#issuecomment-2229239684

huishao-r:
@srherbener would you please take another look given you commented before?

src/marine/gmao_obs2ioda.py
srherbener: I agree that the code is doing what you are intending, but I am questioning the use of the PreQC group for this purpose.

The issue I'm bringing up is based on my understanding that any variable in the PreQC group represents QC marks that came from the provider of the data (ie, whoever it was that provided the input file to this converter). If my understanding is correct, then to me this code is speaking on behalf of the provider and inserting good/bad QC without the provider's input. Perhaps another group besides PreQC could be used so it is clear that we have deemed the good/bad quality of the variables and it is clear that the provider hasn't done that.

@BenjaminRuston and @gthompsnJCSDA could you help with this? Please check my understanding and provide guidance. If I'm wrong that's fine but I just want to make sure that we are doing the right thing. Thanks!

gthompsnJCSDA: Based on past usage of quality markers at NCEP-EMC (table 7), I would suggest simply changing the default starting value of PreQC to "2" instead of "0".  That link to quality markers at NCEP indicates that 0 value means it is judged as "keep" and always assimilate, versus 1 for "good" and 2 is used for default or not really checked.

What that means to me is zero is something that may need no additional checks at all (except perhaps `Background Check`) where 2 is saying that UFO should do whatever QC steps are needed before saying the obs is ready for assimilation.

Ultimately, if you believe the incoming data are just always good, then you can pick either zero or one, but I still suggest two is the proper assignment in the converter.

gthompsnJCSDA (dismissed):
Just suggesting an easy one-line change.

BenjaminRuston:
really @gthompsnJCSDA `2` as default that very counter intuitive to me
all observations even those that are `0` undergo QC anything non-zero is consider a fail imho, the use of `2` is non-generic and non-standard and refers to a NOAA-EMC table? never heard of this table and it sounds again very specific and do not wish to put in agency specifics 

@Dooruk do you know any more on this.  is my thinking off, am I missing something?

gthompsnJCSDA:
> really @gthompsnJCSDA `2` as default that very counter intuitive to me all observations even those that are `0` undergo QC anything non-zero is consider a fail imho, the use of `2` is non-generic and non-standard and refers to a NOAA-EMC table? never heard of this table and it sounds again very specific and do not wish to put in agency specifics
> 
> @Dooruk do you know any more on this. is my thinking off, am I missing something?

In the "big picture" view, it really doesn't matter what anyone picks for the starting value, but very typical use in files converted from both EMC and NASA have used the value 2 since they are anchored mostly in prepBUFR use.  I think I explained it pretty well.  No one has to agree, but typical usage/convention of these two groups would seem better than switching things arbitrarily even if we don't agree on the rather ancient Table 7.  Nothing makes JEDI's `QCflags.h` anything special either.

BenjaminRuston:
i've re-read the comments and agree with @Dooruk on this one.

a PreQC should not be limited to those from the data provider (often in the form of quality bits) but can also be physical reality checks such as those Dooruk mentioned.  

further the PreQC with a value of `0` should NOT be assumed `good` and subsequent checks in the UFO should always be performed.

Dooruk:
If 0 still has to go through UFO machinery that is good enough for me, at least for now. Appreciate the discussion!

Thanks again @gthompsnJCSDA. I shared this discussion and the NCEP table with our obs groups to benefit people like me who weren't aware of this standard.

srherbener (approved):
Thanks everyone for the discussion about the PreQC group. It seems that my understanding was incorrect, so I'm good approving now. Thanks for your patience!

PR #1533, opened 2024-07-22
ensure epoch uses UTC timezone

### LINKED ISSUE #1532: epoch time object in def_jedi_utils should use utc timezone
## Description

the epoch time as defined in `def_jedi_utils` should use the UTC timezone

## Requirements

datetime library of python

all test files that are impacted are updated using new `epoch` definition

BenjaminRuston:
## Description

modify the definition of epoch to use UTC timezone

## Issue(s) addressed

Resolves #1532 

## Checklist


fabiolrdiniz (approved):
Thanks, @BenjaminRuston! It's nice to see enforcing UTC times.

mer-a-o (approved):

srherbener (approved):
Thanks for fixing this!

BenjaminRuston:
using this `UTC` epoch the correct approach we believe and will need to update some of the `testouput` for these ctests:

```
test_iodaconv_icartt..................................................................failed
test_iodaconv_viirs_j1_l1b_albedo.....................................................failed
test_iodaconv_atms_nc4................................................................failed
test_iodaconv_atms_BGRemap_nc4........................................................failed
test_iodaconv_ghcn_snod...............................................................failed
```

BenjaminRuston:
this appears to be a do nothing PR, it does make it explicit should be aware to use UTC times but it only adds the UTC timezone attribute to the python datetime objects.  no `testoutput` files needed updating as they already observed UTC times.

also just to be sure added this to one of the converters:
`os.environ["TZ"] = "UTC"`

and also no changes in the output

PR #1499, opened 2024-04-18
initial commit of cloudsat decoder

BenjaminRuston:
## Description

Initial commit of a cloudSat decoder for CPR L2 data

## Issue(s) addressed

Resolves #1498 

## Impact

Expected impact on downstream repositories:

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have run the unit tests before creating the PR

PR #1534, opened 2024-07-23
IODA converter for DPR GPM

imoradi:
## Description

Included the converter for DPR GPM and also made modifications in the main code so that can process the data using the date range for each window.

## Issue(s) addressed

Resolves #<issue_number>

## Dependencies

- [x] #1499 

## Impact

## Checklist


imoradi:
@BenjaminRuston I have added and tested the DPR convertor. It works fine but still issues with the second channel. I made some changes in the main code so instead of providing the filenames, we can just provide a date-range like from 2022021521 to 2022021603 and the code will list all the files within the window and process them so we don't have to deal with finding and merging the files afterwards. It is much easier to process the data this way than a single file, though the code still works if you provide the filenames. I still have the cloudsat related variables in the main code that needs to be revised (like cpr_obs which now should say something like radar_obs). The main code is modified to work with both cloudsat and GPM DPR.

src/hdf5/read_dpr_gpm.py
```diff
@@ -0,0 +1,162 @@
...
+        for idt in dt_range:
+            # 2A.GPM.DPR.V9-20211125.20170907-S075404-E092639.020033.V07A.HDF5
+            file_pattern = '*.%s*.HDF5' % idt.strftime('%Y%m%d-S%H')
+            for dir, _, _ in os.walk(dpr_dir):
+                dpr_fnames.extend(glob(os.path.join(dir, file_pattern)))
+
+    if not dpr_fnames:
+        return []
+
+    # limit data so that dt_start <= date <= dt_end
```
BenjaminRuston: we generally do not limit times here as this is easily accomplished within the UFO

it does mean files may have some overlapping data in the IODA files if say even 6-hourly files are created

src/hdf5/read_dpr_gpm.py
```diff
@@ -0,0 +1,162 @@
...
+    dprdata['centerWN'] = xr.DataArray(np.array([0.453, 1.183]), dims={"nfreq": 2})  # 1/cm
+
+    # rename dimensions
+    dprdata = dprdata.rename({'nbin': 'elevation', 'nfreq': 'channel', 'nray': 'fov', 'nscan': 'scanline'})
+    dprdata['fov1'] = dprdata.fov.copy()
+
+    return dprdata
+
+
+def read_dpr_gpm(dpr_fnames,
```
BenjaminRuston: this function is only used by this decoder and scans for all files

generally like to handle this outside the decoder, pass in wildcards or a list of files that one would like to process

BenjaminRuston:
```
pycodestyle --config=../../.pycodestyle read_dpr_gpm.py 
```

this is how you run for a single file

test/CMakeLists.txt
```diff
@@ -961,6 +963,18 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_mrms
...
                           mrms_reflectivity_Z1500m.nc ${IODA_CONV_COMP_TOL_ZERO})
 
+ecbuild_add_test( TARGET  test_${PROJECT_NAME}_dpr_gpm
+                  TYPE    SCRIPT
+                  ENVIRONMENT "PYTHONPATH=${IODACONV_PYTHONPATH}"
+                  COMMAND bash
+                  ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                          netcdf
+                          "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/cloudsat_netcdf_to_ioda.py
+                          -i testinput/gpm_dpr_20240401-S105223-E122536.sample.hdf5
```
fcvdb:
```suggestion
                          -f testinput/gpm_dpr_20240401-S105223-E122536.sample.hdf5
```

src/hdf5/cloudsat_netcdf_to_ioda.py
```diff
@@ -95,55 +137,58 @@ def main(args):
...
+    obs_data[('longitude', metaDataName)] = file_obs_data.lon.values.astype(np.float32)
     obs_data[('satelliteIdentifier', metaDataName)] = np.full((nobs), WMO_sat_ID, dtype='int32')
-    obs_data[('sensorCentralFrequency', metaDataName)] = np.array([94.05]).astype(np.float32)    # cpr_sim.Frequency.values.astype(np.float32)
-    obs_data[('sensorCentralWavenumber', metaDataName)] = np.array([3.1371]).astype(np.float32)  # cpr_sim.Wavenumber.values.astype(np.float32)
-    obs_data[('sensorPolarizationDirection', metaDataName)] = np.array([9]).astype(np.int32)
-    obs_data[('sensorScanPosition', metaDataName)] = cpr_obs.fov1.values.astype(np.int32)
-    obs_data[('sensorChannelNumber', metaDataName)] = cpr_obs.channel.values.astype(np.int32)
+    obs_data[('sensorCentralFrequency', metaDataName)] = file_obs_data.centerFreq.values.astype(np.float32)
+    obs_data[('sensorCentralWavenumber', metaDataName)] = file_obs_data.centerWN.values.astype(np.float32)
+    obs_data[('sensorPolarizationDirection', metaDataName)] = np.full((nobs), 9, dtype='int32')
```
BenjaminRuston: @imoradi what is this value?  Is it in the provided files, it appears it's set to a hardcoded value of `9` for both cloudsat and GPM DPR?

@BenjaminTJohnson is this needed?

imoradi:
That is the channel polarization - it is not in the files anywhere. It is not actually even used. I added it for other purposes.

BenjaminRuston:
> That is the channel polarization - it is not in the files anywhere. It is not actually even used. I added it for other purposes.

is it correct to have it hardcoded to a single value for both cloudsat and the gpm_dpr?  should we just remove it?

You can add values via the yaml file too,,, so if you do want to add `sensorPolarizationDirection` to data for the UFO to use it can be done there without modifying the converter

imoradi:
you can just remove it

BenjaminRuston (approved):
thanks for getting this going @imoradi 

will do some small clean-up just renaming files but this is a great new addition many thanks

fabiolrdiniz (approved):
Thanks, @imoradi!

PR #1536, opened 2024-07-26
remove ims scf ctest no longer used

### LINKED ISSUE #1515: tidy up smap converters
## Description

now that the smap converters have been consolodated, tidy up by removing files and renaming testinputs and testoutputs to simplify

simply this is to remove this file: `ims_scf2ioda.py`
which is no longer used in any ctest the converter: `imsfv3_scf2ioda.py` handles both file types ascii and grid (netCDF4) correctly

it does look like a ctest used this converter:
`-i testinput/imssnow_24km.grib2`

and it uses grib files

lateral move this should go in transparently

## Dependencies
ctests passing

BenjaminRuston:
## Description

There is a ctest that is no longer used, let us remove this and the testinput and output

## Issue(s) addressed

Resolves #1515 

## Impact

Reduce entropy

## Checklist


BenjaminRuston:
@YoulongXia-NOAA can we also remove the `afwa_snod` and associated source and files too?

this would also let us remove a conditional in the CMakeLists.txt for `iodaconv_pygrib_ENABLED`

do we know anyone still using these. they will be saved for posterity in the repository if ever needed

YoulongXia-NOAA:
> @YoulongXia-NOAA can we also remove the `afwa_snod` and associated source and files too?
> 
> this would also let us remove a conditional in the CMakeLists.txt for `iodaconv_pygrib_ENABLED`
> 
> do we know anyone still using these. they will be saved for posterity in the repository if ever needed

Yes, please remove it. Thank you.

ClaraDraper-NOAA (approved):

srherbener (approved):

YoulongXia-NOAA (approved):

CMakeLists.txt
```diff
@@ -62,14 +62,6 @@ else()
   ecbuild_find_package( eckit VERSION 1.13.0 QUIET )
   ecbuild_find_package( oops QUIET )
 endif()
-# Optional: pygrib for python for GRIB1/GRIB2 input files
-if ( USE_PYGRIB )
```
BenjaminRuston, line 66: @srherbener tried to be thorough and did a:
```
find . -name CMakeLists.txt -exec grep -H "iodaconv_pygrib_ENABLED" {} +
```
down the test directory and there are no more use cases for the pygrib the only other converter using `.grib2` specifically is the MRMS one and it uses `eccodes` rather than the `pygrib` library:
https://github.com/JCSDA-internal/ioda-converters/blob/8ae8581f72d9a50cd9f589e1a858b0c562e2da52/src/mrms/mrms_grib2ioda.py#L17

please let me know if you think there are any pitfalls with this removal of the `USE_PYGRIB`

test/CMakeLists.txt
```diff
@@ -2220,32 +2216,6 @@ endif()
...
 
-if( iodaconv_pygrib_ENABLED )
-  ecbuild_add_test( TARGET  test_${PROJECT_NAME}_ims_scf
-                    TYPE    SCRIPT
-                    ENVIRONMENT "PYTHONPATH=${IODACONV_PYTHONPATH}"
-                    COMMAND bash
-                    ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
-                            netcdf
-                            "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/ims_scf2ioda.py
-                            -i testinput/imssnow_24km.grib2
```
BenjaminRuston, line 2231: @srherbener this input file wasn't in the `testinput` directory does it only get loaded if the `USE_PYGRIB` is set?  does it have to be removed from somewhere

basically all the `testinput` and `testoutput` files are not in the repository so as you can see they are not removed in this PR

CMakeLists.txt
srherbener: I think it is safe to remove these converters and tests. My recollection is that they were temporary, and to be usable one had to install pygrib in their python environment (which was the purpose of having the USE_PYGRIB control). 

I think this is unlikely, but if something breaks after removing all of this, then we can address it then.

test/CMakeLists.txt
srherbener: That seems another indication that these converters are not being used.

srherbener (approved):

PR #1538, opened 2024-08-04
[bugfix] Fix the timestamp for proper 4 digits (ignore seconds)

gthompsnJCSDA:
## Description

The time stamp (hours/minutes) was incorrect since it could be single digit or two digits.  So force the integer retrieved for the time to be 4 digits before parsing for the datetime object.

## Issue(s) addressed

## Dependencies

None

## Impact

There should be no impacts on downstream repositories

## Checklist


BenjaminRuston (approved):
thanks @gthompsnJCSDA and for maintaining these

fabiolrdiniz (approved):
Thanks, @gthompsnJCSDA!

PR #1545, opened 2024-08-22
Assigning error to DerivedObsValues. OSW

smaticka:
## Description

Ocean surface winds, osw, are data only containing wind spd. the components are calculated using an ObsFunction, which store the components in DerivedObsValues group. We aren't able to assign ObsError to the DerivedObsValue group. This PR creates a workaround for that. 

ObsValue/windEastward and ObsValue/windNorthward are initiated in the converter as 0's, which makes them `observed variables` in addition to the `simulated variables` and allows for error assignment. Missing values are also assigned in colocations as missing wind speed values. 

skylab yamls need to be updated to match what worked in the ctest. 

to test it, I ran the `fv3-jedi` ctest: `fv3jedi_test_tier1_hyb-3dvar` with an obsSpace for CYGNSS and the needed variables. It is able to process the CYGNSS obs and creates an hofx of the wind components and has the expected ObsErrors we assign. 

**NOTE**:
The current handling of PreQC is to reject all windSpeed values that have 'bad' flags. The desired functionality is to assign the flags to both the speed and components, and let the user choose the flags to use. Currently we are not able to pass the PreQC flags to the wind components, so this needs to be addressed (perhaps another PR). 

**NOTE**:
**Any new CYGNSS and Spire obs need to use this converter and the updated skylab yamls instead of UFO branch with the FinalCheck.cc hack.** 

## Dependencies

skylab PR:  https://github.com/JCSDA-internal/skylab/pull/561

## Checklist

- [x] a converted obs file of CYGNSS runs with hofx 
- [x] the correct ObsError is assigned
- [x] the skylab yamls are updated

### DIFF
--- a/src/hdf5/osw_2ioda.py
+++ b/src/hdf5/osw_2ioda.py
@@ -189,10 +189,9 @@ def main(args):
 
     # Transfer from the 1-D data vectors and ensure output data (ioda_data) types using numpy.
     for iodavar in obsvars_ttl:
-        #iodavar = obsvars[0]
         ioda_data[(iodavar, obsValName)] = np.array(obs_data[iodavar], dtype=np.float32)
         ioda_data[(iodavar, obsQcName)] = np.array(obs_data[iodavar+obsQcName], dtype=np.int32)
-        if iodavar=='windSpeed':
+        if iodavar == 'windSpeed':
             ioda_data[(iodavar, obsErrName)] = np.array(obs_data[iodavar+obsErrName], dtype=np.float32)
 
     # setup the IODA writer
--- a/src/hdf5/osw_2ioda.py
+++ b/src/hdf5/osw_2ioda.py
@@ -56,6 +56,8 @@ meta_keys = [m_item[0] for m_item in MetaDataKeyList]
 obsvars = ['windSpeed', 'windSpeed'+obsErrName, 'windSpeed'+obsQcName]
 # Add wind components to be able to assign obsError with dummy values
 obsvars_dummy = ['windEastward', 'windNorthward']
+# List of the observables
+obsvars_list = [obsvars[0]] + obsvars_dummy
 obsvars_units = 'm s-1'
 obsvars_dtype = 'float'
 
@@ -133,10 +135,9 @@ def main(args):
         file_cnt += 1
 
 #   a value of 0. is set. Using missing value triggers UFO rejection
-    obs_data[obsvars_dummy[0]] = 0.
-    obs_data[obsvars_dummy[1]] = 0.
-    obs_data[obsvars_dummy[0]+obsQcName] = 0
-    obs_data[obsvars_dummy[1]+obsQcName] = 0
+    for iodavar in obsvars_dummy:
+        obs_data[iodavar] = 0.
+        obs_data[iodavar+obsQcName] = 0
 
     # replace missing values
     for MetaDataKey in MetaDataKeyList:
@@ -146,8 +147,8 @@ def main(args):
 
     # find where windSpeed is missing and set the components to missing as well
     mask = obs_data['windSpeed'] == missing_vals[obsvars_dtype]
-    obs_data[obsvars_dummy[0]].mask(mask.values, missing_vals[obsvars_dtype], inplace=True)
-    obs_data[obsvars_dummy[1]].mask(mask.values, missing_vals[obsvars_dtype], inplace=True)
+    for iodavar in obsvars_dummy:
+        obs_data[iodavar].mask(mask.values, missing_vals[obsvars_dtype], inplace=True)
 
     # sort by instrument and then time
     if args.sort:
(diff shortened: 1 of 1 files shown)


fabiolrdiniz (changes requested):
Thanks, @smaticka! Please, notice this file uses 4-space indentation, so I suggest you to adjust your changes to that too.

smaticka:
@fabiolrdiniz, thanks. the coding changes have been made and it should be good to go after the skylab PR is submitted. I believe the only test that is failing is not related to this PR.. 
image

smaticka:
also, noting that the current handling of PreQC is to reject all windSpeed values that have 'bad' flags. The desired functionality is to assign the flags to both the speed and components, and let the user choose the flags to use. Currently we are not able to pass the PreQC flags to the wind components, so this needs to be addressed (perhaps another PR).

fabiolrdiniz (approved):

src/hdf5/osw_2ioda.py
```diff
@@ -151,17 +165,19 @@ def main(args):
     varDict = defaultdict(lambda: DefaultOrderedDict(dict))
     varAttrs = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
 
+    obsvars_0s.insert(0, obsvars[0])
+    obsvars_ttl = obsvars_0s
```
fabiolrdiniz: It is not clear to me why you need this variable.

src/hdf5/osw_2ioda.py
```diff
@@ -125,12 +127,24 @@ def main(args):
         # count files
         file_cnt += 1
 
+    # Add 0s for wind components to be able to assign obsError
+    obsvars_0s = ['windEastward', 'windNorthward']
+    obs_data[obsvars_0s[0]] = 0.
+    obs_data[obsvars_0s[1]] = 0.
+    obs_data[obsvars_0s[0]+obsQcName] = 0.
+    obs_data[obsvars_0s[1]+obsQcName] = 0.
```
fabiolrdiniz: QC flags are integers.

smaticka:
should be good now

src/hdf5/osw_2ioda.py
```diff
@@ -151,17 +165,18 @@ def main(args):
...
-    # set the obs space attributes
-    varDict[iodavar]['valKey'] = iodavar, obsValName
-    varDict[iodavar]['errKey'] = iodavar, obsErrName
-    varDict[iodavar]['qcKey'] = iodavar, obsQcName
-    varAttrs[iodavar, obsValName]['coordinates'] = 'longitude latitude'
-    varAttrs[iodavar, obsErrName]['coordinates'] = 'longitude latitude'
-    varAttrs[iodavar, obsQcName]['coordinates'] = 'longitude latitude'
-    varAttrs[iodavar, obsValName]['units'] = obsvars_units
-    varAttrs[iodavar, obsErrName]['units'] = obsvars_units
+    for iodavar in obsvars_0s:
```
BenjaminRuston: this is incorrect and will not set the variable attributes for the 'Windspeed' variable

src/hdf5/osw_2ioda.py
```diff
@@ -172,10 +187,11 @@ def main(args):
...
 
     # Transfer from the 1-D data vectors and ensure output data (ioda_data) types using numpy.
-    iodavar = obsvars[0]
-    ioda_data[(iodavar, obsValName)] = np.array(obs_data[iodavar], dtype=np.float32)
-    ioda_data[(iodavar, obsErrName)] = np.array(obs_data[iodavar+obsErrName], dtype=np.float32)
-    ioda_data[(iodavar, obsQcName)] = np.array(obs_data[iodavar+obsQcName], dtype=np.int32)
+    for iodavar in obsvars_0s:
+        ioda_data[(iodavar, obsValName)] = np.array(obs_data[iodavar], dtype=np.float32)
+        ioda_data[(iodavar, obsQcName)] = np.array(obs_data[iodavar+obsQcName], dtype=np.int32)
+        if iodavar == 'windSpeed':
```
BenjaminRuston, line 197: `iodavar` will never be 'windSpeed'

src/hdf5/osw_2ioda.py
```diff
@@ -151,17 +165,18 @@ def main(args):
     varDict = defaultdict(lambda: DefaultOrderedDict(dict))
     varAttrs = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
 
+    obsvars_0s.insert(0, obsvars[0])
```
BenjaminRuston: ahhh I see this works but is confusing

src/hdf5/osw_2ioda.py
```diff
@@ -151,17 +169,18 @@ def main(args):
     varDict = defaultdict(lambda: DefaultOrderedDict(dict))
     varAttrs = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
 
+    obsvars_dummy.insert(0, obsvars[0])
```
BenjaminRuston: this really should be changed as it's adding the *true* observable to the `dummy` array which is just confusing for human readibility

BenjaminRuston (approved):
thanks @smaticka a couple more small changes this now completed

fcvdb (approved):
Thanks!

PR #1548, opened 2024-09-02
fix the crash of bufr2ioda.x when it is copied outside the build directory

### LINKED ISSUE #1547: [Bug] bufr2ioda.x crashes when it is copied outside the build directory
When bufr2ioda.x is copied outside the build directory, it crashes with the following error message:
```
./bufr2ioda.x: error while loading shared libraries: libingester.so: cannot open shared object file: No such file or directory
```

## To Reproduce

> What computer are you running on?   

Hera, Jet

> What compilers/modules are you using?

spack-stack 1.6.0

> Steps to reproduce the behavior

1.  copy bufr2ioda.x from build/bin to a location outside the build directory
2.  run bufr2ioda.x using any valid yaml file.

## Expected behavior
bufr2ioda.x should run correctly even it is copied outside of the build directory

guoqing-noaa:
## Description

This PR fixes the crash of bufr2ioda.x when it is copied outside the build directory

## Issue(s) addressed

Resolves #1547 

## Dependencies

none

## Impact

none

## Checklist

### DIFF
--- a/src/bufr/CMakeLists.txt
+++ b/src/bufr/CMakeLists.txt
@@ -125,6 +125,8 @@ if ( iodaconv_bufr_query_ENABLED )
 
   add_library( ${PROJECT_NAME}::ingester ALIAS ingester)
 
+  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:${CMAKE_BINARY_DIR}/lib64")
   ecbuild_add_executable( TARGET  bufr2ioda.x
                           SOURCES bufr2ioda.cpp
                           LIBS    ingester )


ShunLiu-NOAA (approved):
Looks good to me.

PraveenKumar-NOAA (approved):
Looks good to me.

fcvdb (approved):
Thanks Guoqing!

src/bufr/CMakeLists.txt
```diff
@@ -125,6 +125,8 @@ if ( iodaconv_bufr_query_ENABLED )
 
   add_library( ${PROJECT_NAME}::ingester ALIAS ingester)
 
+  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:${CMAKE_BINARY_DIR}/lib64")
```
BenjaminRuston, line 129: looks pretty save this is relative path I assume `CMAKE_INSTALL_RPATH`, does this essentially add the current local directory to allow it to run "stand-alone"?

guoqing-noaa: Yes, this is to encode the relative path ($bundle_build/lib64) into the executable and hence even bufr2ioda.x is moved out of the build/bin directory,  it can still find the correct libraries.

rmclaren (approved):
Thanks for this!

rmclaren:
Just FYI: This code (bufr reading code in ioda-converters) is **deprecated** at this point, and will be removed at some point in the fairly near future. 

This code now has its own project repository (https://github.com/NOAA-EMC/bufr-query). There are many enhancements to the code there, including the ability to read files via MPI and use of YAML "mapping" files in your python scripts (assuming you are using YAML). There are small changes in the YAML file format (please see https://bufr-query.readthedocs.io/en/latest/index.html).

guoqing-noaa:
@rmclaren Thanks a lot for the information!
Is "bufr-query" ready to be used right now?  Do we plan to replace the current bufr2ioda.x in GDASApp and RDASapp soon?  Thanks!

rmclaren:
Yes and yes. I think the plan is to do a code sprint to update these other projects at some point.

guoqing-noaa:
That's great! Thanks, @rmclaren !

src/bufr/CMakeLists.txt
rhoneyager-tomorrow: The current approach may break the bundles in hard-to-debug places. These are top-level CMake settings, and they will override the defaults for all new targets created afterwards. For users of [jedi-bundle](https://github.com/JCSDA/jedi-bundle/blob/develop/CMakeLists.txt), this would change the default behavior in fv3-jedi, fv3, soca, mpas-jedi, etc.

As you've set the `CMAKE_INSTALL_RPATH` to a _fixed_ path in `CMAKE_BINARY_DIR` (i.e. the hardcoded path to your current build), you've now made that build directory unremovable. Also, `lib64` vs `lib` is system-specific, so this won't work in general.

If you want to be able to copy your `bufr2ioda.x` executable elsewhere, you can set `CMAKE_BUILD_RPATH_USE_ORIGIN` to FALSE in your _local_ build, and this will hardcode the path.

rhoneyager-tomorrow (changes requested):
See comment. Thanks @srherbener for letting me review.

guoqing-noaa:
@rhoneyager-tomorrow Thanks for the review and suggestion. I will test this and then make corresponding changes.

src/bufr/CMakeLists.txt
srherbener: I've tried both solutions (the PR approach with CMAKE_INSTALL_RPATH, and @rhoneyager-tomorrow approach with CMAKE_BUILD_RPATH_USE_ORIGIN) on my Mac and neither fix this issue. I'm thinking this is because in the CMAKE_INSTALL_RPATH solution lib64 is not used on the Mac, and in the CMAKE_BUILD_RPATH_USE_ORIGIN case the Mac is not one of the platforms that uses "ORIGIN" so this setting has no effect (which might be an issue on other platforms). 

Perhaps this is a "don't care" for Mac?

I did try a third solution that worked on my Mac (and should work on any platform) which is to make a symbolic link instead of a copy of bufr2ioda.x. This works because the link gets followed before running the executable making the working directory the correct path for the relative RPATH to work. 

Would the symbolic link approach be acceptable?

guoqing-noaa: @srherbener Thanks for the testing and providing the perspective of a MAC machine.
A symbolic link will not be accepted in NOAA operation, that's why we created this PR.

For your testing, if `lib64` is not used, does it use `lib` on MAC? 
On HPCs, `lib64` is a link to `lib`. So if MAC uses `lib`, we can change lib64 in this PR to lib to solve the problem.

rhoneyager-tomorrow: @guoqing-noaa can you clarify what you mean by moving bufr2ioda.x out of the build/bin directory? Are you just copying the program, or are you *installing* it to another path via `make install`?

If you are using the first approach, then CMake has no knowledge of the directory structure of where the `bufr2ioda.x` executable exists relative to its libraries. I'm assuming that the purpose of your setting `CMAKE_BUILD_WITH_INSTALL_RPATH` is so that you have some control over the search path, but setting this for all users, including those not using the NOAA HPCs, is not a great option. If you want to go down this route, take a look at tools like [`patchelf`](https://github.com/NixOS/patchelf) and [`chrpath`](https://linux.die.net/man/1/chrpath). Both let you set the RPATH manually.

With the second approach, you can tell CMake explicitly where you want to place your binaries and libraries as part of an installation phase that is run after the binaries are built. Here's the [relevant tutorial page](https://cmake.org/cmake/help/latest/guide/tutorial/Installing%20and%20Testing.html). When I test using this approach, everything just works, and there's no need to muddle with the RPATH manually.

@srherbener re: your comment about "$ORIGIN": you're correct that macs don't use this. Instead they use `@loader_path` and `@executable_path`. These are similar but not quite equivalent, and the CMake RPATH docs mention that [`BUILD_RPATH_USE_ORIGIN`](https://cmake.org/cmake/help/latest/prop_tgt/BUILD_RPATH_USE_ORIGIN.html#prop_tgt:BUILD_RPATH_USE_ORIGIN) has no effect on non-supporting systems.

srherbener: @rhoneyager-tomorrow thanks for the notes. I agree with your suggestions:
- If copying bufr2ioda.x to an arbitrary location, use patchelf or chrpath
- If running `make install` to place bufr2ioda.x then

Using either of these approaches will work for the Mac. The Mac has its own tool (install_name_tool) that can be used to modify the RPATH entries for the first approach, and CMake configuration should handle the second approach.

My vote is for  the second approach (`make install`) since this will copy the necessary libraries (libingester.so) and the relative path in the RPATH for finding libingester.so will work without modification. I tested this on my Mac and it works. Also, using `make install` will create a self-contained installation that has no dependence on the build area making the build area removable.

BenjaminRuston:
@guoqing-noaa and @danholdaway I believe this is now moot, can you please fill us in on your progress also please loop in @srherbener ;0)

danholdaway:
We're following this since it's needed for NCO requirements. We (Cory) are testing out some other ways of building JEDI as a whole with moveable executables.

danholdaway:
Close?

guoqing-noaa:
Thank all for lots of inputs and good discussions.

We will live with the links at the moment and revisit the issue when we get closer to the operational implementation in a few years (if we still use bufr2ioda.x). It looks like we will transition to `bufr-query` soon.

PR #1539, opened 2024-08-06
add ctest for ground-based gnss ipw product

BenjaminRuston:
add a ctest for the ground-based gnss integrated precipitable water (IPW) product

## Description

add a ctest for an NCEP BUFR file for IPW from ground-based GNSS

## Issue(s) addressed

Resolves #<issue_number>

## Impact

Potential to add IPW product to assimilation

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have run the unit tests before creating the PR

PR #1550, opened 2024-09-11
Feature/viirs aod smoke dust

BlakeJAllen-NOAA:
## Description

The VIIRS AOD Converter has been modified to add the ability to identify smoke- and dust-affected AOD observations and output them in separate IODA files. This processing uses information contained in Aerosol Data Product (ADP) files that are distributed as a companion dataset for each VIIRS AOD granule. The smoke and dust processing feature is initially needed for use with the RRFS Smoke/Dust model.

The list and description of changes to enable the new functionality is somewhat long, but I will provide it if it would be helpful for the code review. 

## Issue(s) addressed

N/A

## Dependencies

None

## Impact

None. Converter behavior should not change if new smoke and dust processing capability is not activated using --adp_mask CLI argument

## Checklist

- [ X] I have performed a self-review of my own code
- [ X] I have made corresponding changes to the documentation (assuming the current documentation for the converters is only that in the README.md file)
- [ X] I have run the unit tests before creating the PR


src/compo/viirs_aod2ioda.py
```diff
@@ -13,10 +13,57 @@
...
+    # input array contains decimal representations of the binary values
+    # 10 (2 in base10) is medium confidence and 00 (0 in base10) is high confidence
+    # 01 (1 in base10) is low confidence and 11 (3 in base10) is missing/bad data
+
+    if conf_lvl == 'med' or conf_lvl == 'medhigh':
+        conf_lvl = 'MedHigh'
+    if conf_lvl == 'low':
+        conf_lvl = 'Low'
+    if conf_lvl == 'high':
+        conf_lvl = 'High'
```
CoryMartin-NOAA:
```suggestion
    else:
        conf_lvl = conf_lvl.title()
```
This should force a capital letter and reduces the if statements

src/compo/viirs_aod2ioda.py
```diff
@@ -13,10 +13,57 @@
 import numpy as np
 import os
 
+import glob
```
CoryMartin-NOAA: this isn't used?

ytangnoaa: this imported module is not used

src/compo/viirs_aod2ioda.py
```diff
@@ -117,65 +260,153 @@ def _read(self):
...
+                    lons_dust = lons_dust[mask_thin_dust]
+                    lats_dust = lats_dust[mask_thin_dust]
+                    errs_dust = errs_dust[mask_thin_dust]
+                    qcpath_dust = qcpath_dust[mask_thin_dust]
+                    qcall_dust = qcall_dust[mask_thin_dust]
+                    obs_time_dust = obs_time_dust[mask_thin_dust]
+
+            # calculate and store ob errors
             # defined surface type and uncertainty
             if self.method == "nesdis":
```
ytangnoaa, line 310: This scripts looks only support the method of "nesdis". If it is not "nesdis", the variable "errs" become undefined. If so, please update the document mentioning that this script only support nesdis product. The zero observation error could cause some issues in assimilation.

src/compo/viirs_aod2ioda.py
BlakeJAllen-NOAA: Thanks for catching that import, it was left over from an earlier way of creating the list of ADP files. I have removed it in the newest commit 5161a8b6beb8dbedf2f8edc301435c7e763df8c1.

BlakeJAllen-NOAA: Thanks for catching that import, it was left over from an earlier way of creating the list of ADP files. I have removed it in the newest commit 5161a8b6beb8dbedf2f8edc301435c7e763df8c1.

jeromebarre (changes requested):
Please add a test for this new capability.

src/compo/viirs_aod2ioda.py
BlakeJAllen-NOAA: Good point @ytangnoaa. Based on the documentation, it looks like when that method selection conditional was added there was supposed to be an else block that set the errors to zero if the nesdis method wasn't specified with -m/--method. So the current documentation is incorrect regarding the default behavior. 

I could add in the missing else block to set the errors to zero, but given that both undefined and zero error values could cause problems during DA, I would propose changing the default behavior to use the nesdis method and then updating the documentation accordingly. That way the method option will remain available at the command line for existing scripts that call the code using that option, but the ability to use a method other than nesdis will be gone unless additional methods are added in later. If you think anyone may for some reason be using the converter with the expectation of it following the current default of leaving the error values undefined, though, then I will just update the documentation. 

For clarity, the options I see are:
1.) Leave the code as is, update the documentation to note that the default behavior leaves the error values undefined and is not recommended.
2.) Add in the else block to set the error values to zero for the default case to match the behavior stated in the current documentation, but also update the documentation to note that the default zero error values may cause problems.
3.) Change the default value for the -m/--method argument from none to nesdis and update the documentation to note that the nesdis method is the only method currently available, but the argument remains available for ease of possible future expansion (and for current backward compatibility).

ytangnoaa: Thank you for this information. Since this script is only tested with the nesdis product, the option 3) looks more reasonable

src/compo/viirs_aod2ioda.py
```diff
@@ -107,7 +184,70 @@ def _read(self):
...
+                # Using negative indexing makes the paths irrelevant for the comparison
+                if self.adp_mask[n][-53:-20] != f[-53:-20]:
+                    print(f'ADP data file {self.adp_mask[n]} start and end times do not match those of AOD file {f}')
+                    print(f'Continuing to next AOD and ADP file (or ending)')
+                    continue
+
+                print('Processing the following AOD and ADP files:')
+                print(f)
+                print(self.adp_mask[n])
+                print()
```
jeromebarre: make sure you remove unnecessary print statements

src/compo/viirs_aod2ioda.py
jeromebarre: We should have a default option that do make sense and should avoid creating strictly 0 errors. This WILL cause problems in data assimilation. The data assimilaiton code will crash.  So we should modify the code (the if statement in question here) to avoid creating zeros as errors and make sure that we can only use the nesdis option or whatever allowed options we could have in the future. 
Ok with solution 3. And update the the help argument to something that is not confusing: for example: `obs error calculation method available: nesdis, user must specify it or the converter won't work`.
Thanks

BenjaminRuston:
@BlakeJAllen-NOAA are you familiar with how to add a ctest?

here is an example of a new converter (it just added a yaml file to use the `bufr2ioda.x` executable):
https://github.com/JCSDA-internal/ioda-converters/pull/1539

but you add a small sample input file to `test/testinput` and a corresponding output to `test/testoutput` along with the changes to CMakeLists.txt calling the new test and referring to these new files

src/compo/viirs_aod2ioda.py
BlakeJAllen-NOAA: Thanks @jeromebarre , I was wondering if I would need to create a test for this. Also, sorry for the slow reply, I was out on vacation last week.

Regarding the test, would it be preferred for me to expand the current test that exists for this IODA converter, or should I add a new, separate test to cover the new functionality? 

Regarding the print statements, I had added those just before opening the PR as I thought it may be good for users to have a log of what files were processed, but I've got no problem removing those if that is unneccessary or unwanted. 

(edited to add a user tag, as I'm unsure if reply notifications are sent if that isn't done)

BlakeJAllen-NOAA:
> @BlakeJAllen-NOAA are you familiar with how to add a ctest?
> 
> here is an example of a new converter (it just added a yaml file to use the `bufr2ioda.x` executable): #1539
> 
> but you add a small sample input file to `test/testinput` and a corresponding output to `test/testoutput` along with the changes to CMakeLists.txt calling the new test and referring to these new files

Thanks, Benjamin, it's good to have an example of adding a ctest. I haven't added any before, but it looks like the process is fairly well documented and plus there is the earlier test of this converter's functionality that I can follow in addition to the info that is in the PR you linked. I'll be busy with other things the next couple of days, but hopefully I will have an chance to work on this later this week.

BenjaminRuston:
@BlakeJAllen-NOAA shall we work on this PR.. this week?

if you have a sample data file, can show the next steps.... Just send an email a short meeting will likely wrap this up.

jeromebarre:
Thanks @BenjaminRuston for the support on this. If need be you can add me on the meeting just as optional, not sure I'll be able to make it.

test/CMakeLists.txt
```diff
@@ -1231,6 +1233,20 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_viirs_aod
                           -n 0.0"
                           viirs_aod.nc ${IODA_CONV_COMP_TOL})
 
+ecbuild_add_test( TARGET  test_${PROJECT_NAME}_viirs_aod
```
BenjaminRuston:
```suggestion
ecbuild_add_test( TARGET  test_${PROJECT_NAME}_viirs_smoke_dust
```

this test should have a unique name

BenjaminRuston:
@BlakeJAllen-NOAA I'm going to put this into draft mode as well as then you can do many updates and pushes to the repository and not kick off a bunch of regression tests

once we think it's ready, then will remove from draft mode, and run the CI

README.md
```diff
@@ -108,6 +108,12 @@ Usage: <converter.py> -i INPUT_FILE(S) -o OUTPUT_FILE -m nesdis -k maskout -t 0.
 ```
 For method option (-m) of bias and uncertainty calculation (default/nesdis), deafult means to set bias and uncertainty as 0.0 and nesdis means to use NESDIS bias and uncertainty calculation method. For maskout option (-k) default/maskout, default means to keep all missing values and maskout means to not write out missing values. For thinning option, the value should be within 0.0 and 1.0 depending how much data will be thinned, and 0.0 means without any thining.
 
+The AOD converter normally produces a single IODA output file, but it can also produce two additional output files, one containing only smoke-affected obs and one containing only dust-affected obs. This functionality uses flagging in the Aerosol Data Product (ADP) files created as companion datasets for the native AOD550 observation files. To activate the smoke and dust processing, use the --adp_mask flag followed by the ADP file(s) that correspond to the AOD file(s) specified with -i.
+
+```
+Usage (with smoke and dust processing): <converter.py> -i AOD_INPUT_FILE(S) --adp_mask ADP_INPUT_FILE(S) --adp_conf_lvl 'medhigh' -o OUTPUT_FILE -m nesdis -k maskout -t 0.0
```
BenjaminRuston:
```suggestion
Usage (with smoke and dust processing): <converter.py> -i AOD_INPUT_FILE(S) --adp_mask ADP_INPUT_FILE(S) --adp_conf_lvl 'medhigh' -o OUTPUT_FILE -m nesdis -k maskout -n 0.0
```

think the `-t` is a typo and it should be `-n` for thinning?  I've only used `-n`... and as we work on this will change these flags to be more descriptive and should be able to condense

the dual capability can likely be made clearer in the help message from the parser too, will look into that

BlakeJAllen-NOAA: Yes, your change is correct. I like the approach @jeromebarre suggested in our meeting earlier this week of using more long form arguments (e.g. --method instead of -m). The -i and -o flags are in common enough use for input and ouput files that those are easy to understand as they are, but it would definitely be more straightforward for all of the other arguments to be long form. If nothing else it will keep me from forgetting that -t isn't the correct flag for thinning.

test/CMakeLists.txt
```diff
@@ -1251,6 +1254,22 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_viirs_aod
...
+                  ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
+                          netcdf
+                          "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/viirs_aod2ioda.py
+                          -i testinput/JRR-AOD_v3r2_j01_s20240110T202044Z_e20240110T202044Z.sample.nc
+                          --adp_mask testinput/JRR-ADP_v3r2_j01_s20240110T202044Z_e20240110T202044Z.sample.nc
+                          --adp_conf_lvl 'medhigh'
+                          -o testrun/obs.20240110T202044Z_PT1M_aod.nc
+                          -m nesdis
+                          -k maskout
+                          -n 0.0"
```
BenjaminRuston, line 1270: I would advise modifying the code and simplify these, making a few more transparent `-n` should not have a shorthand and be --thinning_distance (option) and should have a default of 0.0.... the `-k` should be a logical switch, the `adp_conf_level` again spelled out a bit more `adp_confidence_level` and should default to `'medhigh'` and then does not need to be present in the call.  I can't think of a more verbose name for the `--adp_mask` but think it too should be more explicit.  And the help message should spell out the ADP acronym and have two section describing separately how to run for just AOD, and then how to run for ADP.

At this point OK doing 2 ctests one checking smoke and the other dust; however, this could be done via optional flag requiring two call to the converter

```suggestion
                          --adp_mask testinput/JRR-ADP_v3r2_j01_s20240110T202044Z_e20240110T202044Z.sample.nc
                          -o testrun/obs.20240110T202044Z_PT1M_aod.nc
                          --apply_mask"
```

BenjaminRuston:
@CoryMartin-NOAA are you using this ?

@BlakeJAllen-NOAA any plans to continue to pursue this?

CoryMartin-NOAA:
@BenjaminRuston this is for the 'simplified' smoke/dust system developed mainly by GSL, so I'm aware but EMC is not actively using it

jeromebarre:
@BlakeJAllen-NOAA, Could you please prioritize finishing this PR? It's important to avoid leaving relatively straightforward PRs like this one unresolved for an extended period. We appreciate your contributions and your responsibility in seeing this through. However, if it's not completed soon, we may need to close it without merging into the develop branch. Thanks!

jeromebarre:
@BlakeJAllen-NOAA nothing has been done so far I am about to close this, I give you until the end of the week. I appreciate your understanding.

BlakeJAllen-NOAA:
> @BlakeJAllen-NOAA nothing has been done so far I am about to close this, I give you until the end of the week. I appreciate your understanding.

Hi @jeromebarre, apologies for the time it has taken me to get back to this, but I finally have time to get to it now. I'm
 working on the ctest this afternoon.

jeromebarre:
Just be aware of the other PR on the VIIRS converter. Once the other PR merged you'll most likely have conflict. This is the exact reason why we don't want to have PRs hanging for that long.

jeromebarre:
https://github.com/JCSDA-internal/ioda-converters/pull/1604

BlakeJAllen-NOAA:
Thanks for the reminder about that PR @jeromebarre. I will look through their changes to see how complicated reconciling the changes would be if that is needed.

BenjaminRuston:
@BlakeJAllen-NOAA I had a look through this,, 
and thinking the easiest pathway for this is actually a routine that uses `viirs_aod2ioda.py` but is independent `viirs_aod_adp2ioda.py`

this requires the `adp` files think this may be the better approach to reconcile the differences... took a stab at it but not quite there

it also abstracts the specific code for handling the `adp` files

BenjaminRuston:
@BlakeJAllen-NOAA not sure where you're at but likely going to close this

again happy to have a talk with you regarding where I went with it and what I understand about how to best integrate it and add unit tests, but at this point do think it is maybe adding a new routine specifically for the `adp` files and it utilizes the current `aop` code rather than creating a all-in-one

BenjaminRuston:
due to inactivity going to close for now, please re-open when work picks up

this can be added as an extension to existing capability, recommend trying to separate and make explicit which will highlight that this is unique from AOD and narrowing on constituents

PR #1573, opened 2024-10-31
Update pandora converter with site classification information

mer-a-o:
## Description

This PR removes Vertice information from this converter, as it is a total column observation, and Vertice information is unnecessary. 

Site classification information based on `pct_urb` from gaoOniel file (CESM input) is added to the converter and the obs file. This can information can be used to separate urban vs rural sites. pct_urban for low, medium, and high density grids are saved separately under `pctUrbL`, `pctUrbM`, `pctUrbH`. These variables may be useful If we decide to spend more time improving this site classification.
`airQualityClassification` is 0 for UNKNOWN, 1 for rural, 2 for suburban, and 3 for urban sites. 

## Issue(s) addressed

Related to https://github.com/JCSDA-internal/skylab/issues/608

## Dependencies

non2

## Impact

## Checklist


src/compo/pandora_2ioda.py
```diff
@@ -97,11 +100,44 @@ def read(self):
...
+                print('site classification file is available')
+                pandora_sites = pd.read_csv(self.site_classification)
+                pandora_lats = pandora_sites['lat']
+                pandora_lons = pandora_sites['lon']
+
+                lat_tol = 0.01
+                lon_tol = 0.01
+                # Filter rows within the tolerance range
+                close_rows = pandora_sites[
+                    (pandora_lats >= lat - lat_tol) & (pandora_lats <= lat + lat_tol)
```
jeromebarre, line 115: I think you could use that instead:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html#:~:text=Interpolate%20and%20extrapolate,%3DNone)

in case we have sharp gradients in the population density this would be a bit cleaner.

mer-a-o: I don't clearly see how to use interpolation here. Can you give a little more info please?

weiwilliam:
@mer-a-o Why do you want to check lat/lon tolerance? Can we use the file name or [Pandora???s1] to locate the information in the classification CSV file?

mer-a-o:
@weiwilliam, yes, we could use site names, but I thought there's always a chance that site names could change. So checking lat/lon within a tolerance may be safer.

src/compo/pandora_2ioda.py
```diff
@@ -97,11 +100,44 @@ def read(self):
...
+            aq_class = np.zeros(nlocs, dtype=np.int32)
+
+            if self.site_classification:
+                print('site classification file is available')
+                pandora_sites = pd.read_csv(self.site_classification)
+                pandora_lats = pandora_sites['lat']
+                pandora_lons = pandora_sites['lon']
+
+                lat_tol = 0.01
+                lon_tol = 0.01
```
BenjaminRuston, line 112: @mer-a-o not as concerned about the interpolator, but did you want to make the tolerance an optional input parameter?

mer-a-o: No, not really. I picked that value based on the two closest sites, and it doesn't need the user's input.

src/compo/pandora_2ioda.py
jeromebarre:
```
print('site classification file is available')
pandora_sites = pd.read_csv(self.site_classification)

# Extract the required columns
pandora_lats = pandora_sites['lat'].values
pandora_lons = pandora_sites['lon'].values
urb_classes = pandora_sites['urb_class'].values
pct_urb_values = pandora_sites['pct_urb'].apply(lambda x: list(map(float, x.strip("[]").split()))).tolist()

# Create interpolators
urb_class_interpolator = RegularGridInterpolator((pandora_lats, pandora_lons), urb_classes, bounds_error=False, fill_value=0)
pct_urb_L_interpolator = RegularGridInterpolator((pandora_lats, pandora_lons), [v[0] for v in pct_urb_values], bounds_error=False, fill_value=0)
pct_urb_M_interpolator = RegularGridInterpolator((pandora_lats, pandora_lons), [v[1] for v in pct_urb_values], bounds_error=False, fill_value=0)
pct_urb_H_interpolator = RegularGridInterpolator((pandora_lats, pandora_lons), [v[2] for v in pct_urb_values], bounds_error=False, fill_value=0)

# Query point
query_point = np.array([lat, lon])

# Perform interpolation
urb_class = urb_class_interpolator(query_point)
pct_urb_L = pct_urb_L_interpolator(query_point)
pct_urb_M = pct_urb_M_interpolator(query_point)
pct_urb_H = pct_urb_H_interpolator(query_point)

# Handle cases where no valid data is found
if np.isnan(urb_class):
    print("No nearby location found within the tolerance. assign UNKNOWN")
    urb_class = 0
    pct_urb_L, pct_urb_M, pct_urb_H = 0, 0, 0

# Assign values to arrays
aq_class = np.full(nlocs, urb_class, dtype=np.float32)
pct_urb_L = np.full(nlocs, pct_urb_L, dtype=np.float32)
pct_urb_M = np.full(nlocs, pct_urb_M, dtype=np.float32)
pct_urb_H = np.full(nlocs, pct_urb_H, dtype=np.float32)
```

src/compo/pandora_2ioda.py
weiwilliam: One thing I'm not sure if it happens.
If two stations are very close and fall within the tolerance, the code will use the first row, but it doesn't have to be the closest.

mer-a-o: I doubt that will happen. The tolerance is already pretty small. If a new site is added, the list needs to be updated anyway so the user can then adjust to tolerances if needed.

jeromebarre:
Approving this for now until we find a generalized solution for classification capability in ioda/ioda-convs

jeromebarre (approved):

weiwilliam (approved):

BenjaminRuston (approved):
i've run the converter and matched the testoutput file.. thanks @mer-a-o

PR #1576, opened 2024-11-08
converter for ground based tec from tenet network

### LINKED ISSUE #1583: create an initial ground-based total electron content decoder
## Description

ground based total electron content is an observation consistently in use by ionospheric modelers

the TENET format is a deprecating standard and will be soon replaced. The TENET text-based format will have large archives for a prolonged period so having a capability to decode these data is appropriate.

## Requirements

Ingest a ground-based TEC file and create a IODA output.

working decoder and ctest 

## Dependencies
python bindings to IODA

BenjaminRuston:
## Description

This adds a decoder for observations in TENET format which are ground based total electron content

## Issue(s) addressed

Resolves #1583 

## Impact

Adds a ground based TEC decoder

## Checklist


BenjaminRuston:
@haydenlj this adds a couple new IODA definitions... one key one is that for `IPP` and forgot to ask @davidkuhl 

it's maybe `Ionospheric Pierce Point` ?

src/gnssro/gnss_tec_ground_based.py
```diff
@@ -0,0 +1,469 @@
...
+# These modules need the path to lib-python modules
+import pyiodaconv.ioda_conv_engines as iconv
+from pyiodaconv.orddicts import DefaultOrderedDict
+from pyiodaconv.def_jedi_utils import iso8601_string, epoch
+
+os.environ["TZ"] = "UTC"
+
+# these are the unique values in the raw input file
+varDict = {'totalElectronContentVertical': ['totalElectronContentVertical', "integer", 'TECU'],
+           'totalElectronContentSlant': ['totalElectronContentSlant', "integer", "TECU"],
```
BenjaminRuston: @davidkuhl do we ignore the `Vertical` TEC, we only care about the `Slant` TEC correct?

src/gnssro/gnss_tec_ground_based.py
```diff
@@ -0,0 +1,469 @@
...
+}
+
+# these are the MetaData common to each input
+locationKeyList = [
+    ('latitude', 'float', 'degrees_north'),
+    ('longitude', 'float', 'degrees_east'),
+    ('pseudoRandomNoiseCode', 'integer', 'GNSS transmitter pseudoRandomNoise PRN code'),
```
BenjaminRuston: this is potentially another new one @haydenlj , need to look what we have for GNSS already the PRN codes are for each GNSS constellation

src/gnssro/gnss_tec_ground_based.py
davidkuhl: @BenjaminRuston Yes, we are only interested in assimilating the Slant TEC and verifying against the Slant TEC.

src/gnssro/gnss_tec_ground_based.py
haydenlj: @BenjaminRuston looks like we don't have anything for PRN specifically, what we have is satellite identifier https://github.com/JCSDA-internal/ioda/blob/develop/share/ioda/yaml/validation/ObsSpace.yaml#L887
Is this different?

haydenlj (approved):
Thanks @BenjaminRuston I was able to run

haydenlj:
> @haydenlj this adds a couple new IODA definitions... one key one is that for `IPP` and forgot to ask @davidkuhl
> 
> it's maybe `Ionospheric Pierce Point` ?

Ok I'll add these to the IODA definitions PR. Thanks!

src/gnssro/gnss_tec_ground_based.py
```diff
@@ -0,0 +1,468 @@
...
+    'totalElectronContent': ['totalElectronContent', "integer", 'TECU'],
+}
+
+# these are the MetaData common to each input
+locationKeyList = [
+    ('latitude', 'float', 'degrees_north'),
+    ('longitude', 'float', 'degrees_east'),
+    ('pseudoRandomNoiseCode', 'integer', 'GNSS transmitter pseudoRandomNoise PRN code'),
+    ('elevationAngleGNSS', 'float', 'GNSS transmitter satellite elevation angle in degrees'),
+    ('azimuthAngle', 'float', 'aziumuth angle viewing GNSS transmitter in degrees west'),
```
haydenlj: @BenjaminRuston regarding the new names, is this different from `beamAzimuthAngle` or `sensorAzimuthAngle` ?

BenjaminRuston: great catch @haydenlj we should pick one of those... can you help decide which is more appropriate?

it is a sensor, beam also works though....  would think the `beamAzimthAngle` is being used for RADAR, but may be best?

src/gnssro/gnss_tec_ground_based.py
BenjaminRuston: the PRN number is kinda like the identifier number or series number from each constellation.... so GPS has 1-42 PRN,,, and so does Galileo and GLONASS.

in the conventions we have the `satelliteTransmitterId` which I think is what I'm looking for:
https://github.com/JCSDA-internal/ioda/blob/24df6cc5938d655aed1274af5605b53795dd55bb/share/ioda/yaml/validation/ObsSpace.yaml#L890

and then this defined GPS, GLONASS, Galileo, ... :
https://github.com/JCSDA-internal/ioda/blob/24df6cc5938d655aed1274af5605b53795dd55bb/share/ioda/yaml/validation/ObsSpace.yaml#L896

does this make sense to you, think this is the correct one to use... we can ask others if it makes sense too

src/gnssro/gnss_tec_ground_based.py
BenjaminRuston: I guess beam infers the observation location as the point of reference and here the beam originates at the transmitter.. so maybe `sensor` should be selected in this case as it's looking towards a source

haydenlj: Yes that's the way I was thinking about it too, sounds good to me

src/gnssro/gnss_tec_ground_based.py
haydenlj: Yes that makes sense

BenjaminRuston: I just looked at a IODA file and yes this is what we want:

```
range(/MetaData/satelliteTransmitterId): [1, 32]
```

davidkuhl (approved):

huishao-r (commented):
@BenjaminRuston I checked the testoutput/obs.20201001T000000Z_PT1M_tec_groundbased.nc4, which only has one group, MetaData. There is no ObsValue group. Is it correct?

BenjaminRuston:
@huishao-r this is fixed and can now make a plot of the data... notice the units for this TEC for the groundbased is in different units then the TEC from the space-based GNSS receivers.... they use something called `TECU` not sure what units these groundbased TEC are in

fcvdb:
The test_iodaconv_tec_groundbased test passed, but the IODA validator prints some errors in the IODA conventions:

/work2/noaa/jcsda/fvandenb/skylab/jedi_2024121023_tec_ground_based_tenet_gnu_hercules/build/bin/ioda-validate.x "
--ignore-warn" "--ignore-error" "/work2/noaa/jcsda/fvandenb/skylab/jedi_2024121023_tec_ground_based_tenet_gnu_hercules/build/ioda/share/test/testinput//validation/ObsSpace.yaml" "/work2/noaa/jcsda/fvandenb/skylab/jedi_2024121023_tec_ground_based_tenet_gnu_hercules/jedi-bundle/iodaconv/test/testoutput/obs.20201001T000000Z_PT1M_tec_groundbased.nc4"

```
285:   Variable MetaData/stationIdentifierWMO
285:    ERROR: Variable 'MetaData/stationIdentifierWMO' is not listed in the YAML conventions file.
285:   Variable MetaData/xECEFPosition
285:    ERROR: File variable 'MetaData/xECEFPosition' has units of 'receiving station Earth Centered Earth Fixed X-coordinate in meters', which are invalid.
```

fcvdb (approved):

PR #1584, opened 2024-11-22
adjust bit checking of quality word

### LINKED ISSUE #1574: TROPICS converter is reading the wrong QC bits
https://github.com/JCSDA-internal/ioda-converters/blob/6f87a0f279e836fd604e5b313a25bd1e54bff80e/src/hdf5/tropics_2ioda.py#L315

I've been debugging why the tropics2ioda converter is not producing the same results as my internal TROPICS/TMS converter. `tropics_2ioda.py` has an off-by-one error. The TROPICS L1B describes the calQualityFlag bitfield with bit positions starting at 1, and as the code is written the bit shift goes one position too far.

Here's a comparison of what the converter is currently doing (top), versus what it should be doing (bottom).

satelliteAscendingFlag_20240917T00Z_noch
flagAscDesc_20240917T00Z_ch1

 The QC flag misread causes two issues:
- The day/night flag is stored as `MetaData/satelliteAscendingFlag`.
- The set_missing_value function (https://github.com/JCSDA-internal/ioda-converters/blob/develop/src/hdf5/tropics_2ioda.py#L298) in this converter is designed to skip QC-bad data and avoid writing it to a ioda file. It is misinterpreting the different flags and is filtering the wrong data. Ex: descending orbits are taken to be failures in the hot QC flag.

BenjaminRuston:
## Description

adjust the bit checks for the quality word to begin with 0 rather than 1

## Issue(s) addressed

Resolves #1574 

## Impact

Quality word interpretation corrected

## Checklist

### DIFF
--- a/src/hdf5/tropics_2ioda.py
+++ b/src/hdf5/tropics_2ioda.py
@@ -260,14 +260,14 @@ def get_data_deprecated(f, obs_data, skip=1):
     obs_data[(k, "ObsError")] = np.full((nlocs, nchans), 5.0, dtype='float32')
     obs_data[(k, "PreQC")] = np.full((nlocs, nchans), 0, dtype='int32')
 
-    # Bit 1: land/undefined
-    # Bit 2: Lunar/solar intrusion
-    # Bit 3: Active Maneuver
-    # Bit 4: Cold Cal. Consistency
-    # Bit 5: Hot Cal. Consistency
-    # Bit 6: Ascending/Descending
-    # Bit 7: Day/Night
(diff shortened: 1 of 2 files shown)


BenjaminRuston:
@smaticka you'll need to re-run the ingest suite for January, the current converter will get data into IODA but these flags should improve the coverage

smaticka (approved):
thanks @BenjaminRuston this worked when I used it during a TROPICS ingest.

fcvdb (approved):
The changes pass the test. Thanks!

PR #1587, opened 2024-11-25
The Tomorrow.io TMS converter

### LINKED ISSUE #1574: TROPICS converter is reading the wrong QC bits

### LINKED ISSUE #1585: update tropics converter for tomorrowio sounder data
## Description

update converter to ensure compatibility with L1-TB and L1-TBR files from tomorrow(dot)io

## Requirements

early release data files to test converter

files successfully convert, can deprecate proxy data decoding capability

rhoneyager-tomorrow:
## Description

This PR adds the ability to convert data from the [Tomorrow.io Microwave Sounder (TMS)](https://www.tomorrow.io/space/sounder/) L1B data product into IODA format. TMS is a 12-channel microwave sounder with heritage from the NASA TROPICS (Time-Resolved Observations of Precipitation structure and storm Intensity with a Constellation of Smallsats) instruments. The first two TMS satellites have launched, and many more are expected through 2025.

This code provides an example of how to use the IODA C++ API to write a converter.

While the code is designed for Tomorrow's instrument, it is also fully capable of converting the latest TROPICS L1B file format (V05-02).

## Issue addressed

This lets @BenjaminRuston run some experiments with TMS data. USAF asked if I could pass along this converter to enable downstream use.

This may just resolve #1585, and it also offers an alternate path to fix #1574.

## Dependencies

None

## Downstream impact

JEDI can run with TMS!

## Reviewing

I've got a set of diagnostic plots showing that the converter works as expected. Let me know where I should send this for review. The same goes for the L1B product documentation.

Ex. of using eva to plot a few granules for the W band channel:
brightnessTemperature_20241101T0000Z_ch1

The code has notes about recommended IODA updates. Perhaps @srherbener can take a look.

The [ioda validation YAML](https://github.com/JCSDA/ioda/blob/develop/share/ioda/yaml/validation/ObsSpace.yaml) will also need an update, as I have suggested several new MetaData variables. Thoughts on the naming scheme, or on any other part of this PR are quite welcome.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,6 +13,7 @@ add_subdirectory(ssec)
 add_subdirectory(wrfda_ncdiag)
 add_subdirectory(single_observation)
 add_subdirectory(mrms)
+add_subdirectory(tomorrow_io)
 
 # Optional components
 if(iodaconv_gnssro_ENABLED)
--- a/src/tomorrow_io/CMakeLists.txt
+++ b/src/tomorrow_io/CMakeLists.txt
@@ -1,0 +1,21 @@
+# (C) Copyright 2024 The Tomorrow Companies, Inc.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+add_executable(convert_tio.x
+	creation.cc
+	creation_tio_sat_inst_specs.cc
+	copy.h
+	copy_function.cc
+	copy_helpers.cc
+	copy_tms.cc
+	copy_tms_datetime.cc
+	detect_tms_type.cc
+	product.cc
+	product.h
+	main_tio.cc
+	)
+target_link_libraries(convert_tio.x PUBLIC ioda)
+set_target_properties(convert_tio.x PROPERTIES CXX_STANDARD 17)
+
--- a/src/tomorrow_io/README.md
+++ b/src/tomorrow_io/README.md
@@ -1,0 +1,12 @@
+# The Tomorrow Microwave Sounder (TMS) converter
+
+The source codes here provide the ability to convert the Tomorrow.io TMS L1B-TC product
+and the NASA TROPICS L1B product into IODA format.
+
+Usage: `convert_tio.x input_file_1 [input_file_2 ...] output_file`
+
+This converter is written in C++ and uses IODA to read and write the observation data files.
+It provides an example of how to convert data to ioda using the C++ interface.
+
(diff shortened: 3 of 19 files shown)


BenjaminRuston:
@rhoneyager-tomorrow this should be able to be closed, as I've transferred this to an [internal PR](https://github.com/JCSDA-internal/ioda-converters/pull/1590)

will this support the L1-TBR as well

rhoneyager-tomorrow:
This PR targets V00-07 of the L1B-TB product (last week's release). The resampled 1B product (L1B-TBR) is not covered by this PR, as it's under a month old. Will likely address this separately in December.

After every release we do reprocess older data to ensure a consistent product, and a few changes to the file structure were made in the past month based on feedback from older versions. Your TROPICS-based converter (#1588) isn't affected this time.

The main difference between the L1B-TB and -TBR files is that in -TBR the brightness temperatures are resampled to the ch. 11 geolocation (boresight center). The channel geolocations are very close and typically agree to within 1-3 km. This means that either product is usable in JEDI. Here's an example subset of a Tomorrow-S2 scan pattern over the UK:

geolocation

Note that neither product resamples to a common field of view. I don't recall if the FOV differences are handled in the JEDI-CRTM interface, but it would be nice.

rhoneyager-tomorrow:
Closing in favor of #1590. Would be nice if the CI supported this use case. https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

PR #1592, opened 2024-12-06
GMAO converter depth fix

Dooruk:
There was an issue with the `depth` dimension renamed wrong and not working with `SOCA`. We fixed it locally but this updates the repo version.

Hope I don't break the ctests again 🤞


BenjaminRuston:
@travissluka do we want to add `depth` to this:
https://github.com/JCSDA-internal/ioda/blob/6f0148147c712a3b35b17a8d08a926d398c96e42/share/ioda/yaml/validation/ObsSpace.yaml#L678

travissluka:
> @travissluka do we want to add `depth` to this: https://github.com/JCSDA-internal/ioda/blob/6f0148147c712a3b35b17a8d08a926d398c96e42/share/ioda/yaml/validation/ObsSpace.yaml#L678

well, from what I can tell, according to the conventions spreadsheet, `depthBelowWaterSurface` is what all ocean profiles *should* use. But, after searching through old emails, I see that in 2022 we complained that no one kept SOCA in the loop with what was being changed when these PRs originally went in, so, although `depthBelowWaterSurface` was decided on by whoever was involved, it seems everything in SOCA and all the prepared obs still use `depth`. @BenjaminRuston I leave it as a judgement call to you as to whether we want to take the time to correctly switch from `depth` to `depthBelowWaterSurface`.... I doubt anyone has the energy at this point in changing. Also, some of the operators are still hardcoded to use `depth`: `src/ufo/operators/marine/insitutemperature/ufo_insitutemperature_tlad_mod.F90`

so, I vote for adding `depth`, and we'll add this issue to the list of "technical debt". Maybe someday we'll refactor that obs operator to allow for a configurable depth variable so that we can transition to the correct name

or, don't add `depth` so that validate fails, to remind us to change this in the future?

travissluka (approved):
@BenjaminRuston and I decided that `depthBelowWaterSurface` is actually the correct name, but it's too late to enforce that now.
We'll merge this, and promise ourselves that whenever the marine insitu observation operator is touched next we'll modify it to allow for both `depth` and `depthBelowWaterSurface` and slowly transition over to the latter at a later date

BenjaminRuston (approved):
excellent thanks @Dooruk and @travissluka for your review

let us keep this in mind and in the tech debt column for now proceed -- thanks

PR #1595, opened 2024-12-10
Change the unit of electron density from m-3 to cm-3 following model naming convention

huishao-r:
## Description

electron density unit should be cm^-3, instead of m^-3, since the model geoval unit will be cm^-3. Here is the [Model geoval naming convention](https://github.com/JCSDA-internal/pyiri-jedi/issues/49)

## Issue(s) addressed

This change will ensure the same unit implemented for electron density observed values, geoval values and h(x) values.

## Impact

converted ionosonde obs files will be updated with new unit and values.

## Checklist


src/conventional/ionosonde_ascii2ioda.py
```diff
@@ -295,6 +295,10 @@ def populate_obsValue(line, local_data):
         height, freq, f_conf, density, density_conf = map(float, [height, freq, f_conf, density, density_conf])
 
         if all(x > 0 for x in [freq, f_conf, density, density_conf]):
+
+            # Scale the electron density by 1e6 (for m^-3 to cm^-3)
+            density = density * 1e-6
```
BenjaminRuston:
```suggestion
            density *= 1.e-6
```

src/conventional/ionosonde_ascii2ioda.py
```diff
@@ -18,7 +18,7 @@
 varDict = {'criticalFrequency': ['criticalFrequency', "float", 'm s-1'],
            'criticalFrequencyConfidence': ['criticalFrequencyConfidence', "float", 'fractional percent'],
            'height': ['height', "float", "m"],
-           'electronDensity': ['electronDensity', "float", 'number / m^3'],
+           'electronDensity': ['electronDensity', "float", 'number / cm^3'],
```
BenjaminRuston:
```suggestion
           'electronDensity': ['electronDensity', "float", 'number cm-3'],
```

belive this is the convention we use

BenjaminRuston:
@huishao-r a file got added with `^M` at the end of every line!

and seemed like it was maybe from git?  anyway fixed it and added conversion to `density_conf` as well

BenjaminRuston (approved):
this does the conversion to number per cubic centimeter, which isn't SI but apparently follows model conventions

BenjaminRuston:
@huishao-r this is ready correct.. should we get @haydenlj to review

haydenlj (approved):
Thanks @huishao-r !

PR #1600, opened 2025-01-18
fix cygnss quality flag

### LINKED ISSUE #1599: correct interpretation of quality flag in CYGNSS data
## Description

the quality flag in GNSS-R ocean surface wind converter for the NOAA L2 products should be examine and corrected if necessary

## Requirements

input data this should address at a minimum NOAA L2 CYGNSS data.

converter correctly flags and remove data with `bad_quality` marker

## Dependencies
input data and a working converter

BenjaminRuston:
## Description

This corrects the get_normalized_bit function in the CyGNSS ingest

## Issue(s) addressed

Resolves #1599 

## Impact

Correct interpretation of quality bits

## Checklist

### DIFF
--- a/src/hdf5/osw_2ioda.py
+++ b/src/hdf5/osw_2ioda.py
@@ -387,7 +387,7 @@ def quality_control(obs_data, qc_strict=False):
 
 
 def get_normalized_bit(value, bit_index):
-    return (int(value) >> bit_index) & 1 == 0
+    return (int(value) >> bit_index) & 1 == 1
 
 
 if __name__ == "__main__":
--- a/src/hdf5/osw_2ioda.py
+++ b/src/hdf5/osw_2ioda.py
@@ -387,7 +387,7 @@ def quality_control(obs_data, qc_strict=False):
 
 
 def get_normalized_bit(value, bit_index):
-    return (value >> bit_index) & 1
+    return (int(value) >> bit_index) & 1 == 0
 
 
 if __name__ == "__main__":


BenjaminRuston:
think this fixes it... should also check the ascending / descending bit

BenjaminRuston:
this will also apply to this, and now is a good time to put `get_normalized_bit` into a utility (https://github.com/JCSDA-internal/ioda-converters/blob/develop/src/pyiodaconv/def_jedi_utils.py):

https://github.com/JCSDA-internal/ioda-converters/blob/a2f538327ed37be6f98a21ee6a5acf3ef4dd96b8/src/hdf5/tropics_2ioda.py#L420

BenjaminRuston:
and in the tropics converter I tested the ascending descending flag....

believe this PR introduces the correct method and is what should be used... should confirm this method with the tropics converter too

BenjaminRuston:
here's the new figure with quality flag applied:
image


and it's subtle,,, but the counts, min/max and stdv all show improvements in the figure above here's without or no QC bit applied:
image

BenjaminRuston:
tested this change with tropics converter too. because it's an array can't wrap the passed in value with `int(value)` but using it as written and with the `== 1` at the end produced the correct land_flag behavior and asc/des behavior:
land flag (mislabeled):
image


and asc/dsc:
image

BenjaminRuston:
think this whole PR is a nothing burger.... just pulled the CyGNSS data from r2d2 for first 5 days in May and the high wind speeds are already screened here's the figures for data previously converted and newly converted and they are identical:
image

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

BenjaminRuston:
this is a do nothing change the same behavior is output with and without this change

PR #1608, opened 2025-01-28
use the forward flag in the data to determine negative viewing angles

### LINKED ISSUE #1607: tropics use flag to set negative viewing angle
## Description

The forward and aft flag is included in the tropics data and can be used to set the positive and negative viewing angles which are needed for bias correction

## Requirements

recent TROPICS data and the python bindings to IODA

a IODA file can be produced from TROPICS data

BenjaminRuston:
## Description

Use the flag in the TROPICS data to determine which viewing angles are assigned negative values

## Issue(s) addressed

Resolves #1607 

## Impact

The TROPICS viewing angle can be negative and which side of the scan this occurs is triggered by the foreward flag

## Checklist


ashley314 (approved):
This worked for me. Although I did not find a date where the view angle was or needed to be negative.

BenjaminRuston:
@fcvdb or @fabiolrdiniz this needs another review too is another step for tropics assimilation

BenjaminRuston:
this PR show how to use the bit8 (i_forward) which shows spins direction of sensor to assign negative viewing angles

BenjaminRuston:
from Vince Leslie MITLL:
image

BenjaminRuston:
here's the latest update from Vince Leslie (MITLL)

```
So new algorithm for scan angle for the TROPICS NRT NOAA BUFR:

Multiply the scan angles for spots 1 to 40 by -1.
For spot 41:  (new spot 40 scan angle + spot 42 scan angle)/2
```
Spot 42 is blue, Spot 41 is orange, and spot 40 is yellow.
image

fabiolrdiniz:
The unit test fails because of the expected 200% difference in the `sensorViewAngle` due to the flip to negative values. I believe we only need to update the reference file.

BenjaminRuston:
think this is working correctly @fabiolrdiniz 
image


however, I may need to change the ctest testinput file as the scan positions and view angles do not seem correct so think when it was reduced something odd happened

src/hdf5/tropics_2ioda.py
```diff
@@ -260,11 +260,34 @@ def get_data(f, obs_data, skip=1):
...
+#   end testing purpose
+    #  newly adopted strategy by MIT-LL
+    #  Multiply the scan angles for spots 1 to 40 by -1.
+    #  For spot 41:  (new spot 40 scan angle + spot 42 scan angle)/2
+    obs_data[('sensorViewAngle', metaDataName)][scan_sign_backward == 1] *= -1.
+    # ugh what to do here
+    obs_data[('sensorViewAngle', metaDataName)][scan_middle == 1] = (
+        obs_data[('sensorViewAngle', metaDataName)][scan_pre_middle == 1] +
+        + obs_data[('sensorViewAngle', metaDataName)][scan_post_middle == 1]) / 2.
+
```
smaticka, line 288: It looks like lines 264-269, 276-282 aren't used here. Is this PR ready for review @BenjaminRuston or are do you still need these for testing?

Overall, it looks like the codes is suppose to:
1 find when `sensorScanPosition` is less than 41 and then multiply the `sensorViewAngle` by -1, and 
2 find when the `sensorScanPosition` is 40, and set the `sensorViewAngle` equal to the average of `sensorViewAngle` when `sensorScanPosition` is 41 and of the `sensorViewAngle` when `sensorScanPosition` is 42. 

Is that what it's supposed to do? I'm not sure I followed # 2 above.

fabiolrdiniz: @smaticka, (2) is to set position 41 as the average value between 40 and 42.

smaticka: ah, thanks, I misread where the parentheses were. (edited above to reflect)

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

BenjaminRuston:
thanks @fabiolrdiniz I've updated the test reference as indeed it did not contain all the scanPosition

the new file contains all `81` positions and only is about 2x larger still ~100Mb ;0)

PR #1617, opened 2025-02-20
Add surface qualifier and update the uncertainty in VIIRS L1b converter

### LINKED ISSUE #1710: Add surfaceQualifier to VIIRS level 1 albedo converter
## Description

Add surfaceQualifier to VIIRS level 1 albedo converter for future use (filter/diagnose).

surfaceQualifier is added to MetaData group in the IODA file.

## Dependencies
N/A

weiwilliam:
## Description
This PR adds/updates following,
- `MetaData/surfaceQualifier`: include the surface type information from geolocation data.
- the uncertainty now is the absolute value rather than percentage.
- ~~a global attribute `storedReflecatance`: mark the converted data has been divided by cosine of solar zenith angle or the stored reflectance.~~
- date range option.

jedi-ci-test-select=gcc

## Issue(s) addressed

Resolves #1710 

## Dependencies

- [ ] ...

## Impact

## Checklist


src/compo/viirs_l1bnc2ioda.py
```diff
@@ -279,34 +293,44 @@ def main():
...
         help="name of ioda-v2 output file",
         type=str, required=True)
-    parser.add_argument(
+    required.add_argument(
         '-n', '--thin',
         help="percentage of random thinning fro 0.0 to 1.0. Zero indicates"
         " no thinning is performed. (default: %(default)s)",
         type=float, default=0.0)
-    parser.add_argument(
+    required.add_argument(
```
BenjaminRuston: it seems to me this should be an optional argument and the default it `True`.

I see the ctest uses this by default would we recommend this is always on?  If so.  Recommend making it optional with a default to `True` and remove argument from ctest.  Lead them to the recommended method.

Also a required switch that when added makes it `True` seems like an always True action by default anyway?

weiwilliam: @BenjaminRuston Yes, I always run with the secant term in my current case.
I will move it to optional arguments.

But I think the current setup is more intuitive because it reminds users that VIIRS L1b data stores reflectance as the product of the true reflectance and the cosine of the solar zenith angle.
If default this option to `true`, it means this option should be `--no_secterm` and the if-statement of this option in the converter should be inversed because it is weird when running the converter with `--secterm` means not applying the secant term.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -185,12 +190,15 @@ def _read(self):
                 err.set_auto_scale(False)
 
                 # NASA VIIRS apply secant if option enabled
-                vals[:, ichan] = obs[:].data.ravel() * sec_term if self.apply_secterm else obs[:].data.ravel()
+                vals[:, ichan] = obs[:].data.ravel() * sec_term if not self.stored_refl else obs[:].data.ravel()
```
jeromebarre: For clarity, can you put this with a more traditional if/else statement with and use 4 lines of code?

jeromebarre: In general would it be more forward looking to add this secant term in the in ioda file and have either this scaling done in UFO using a yaml key option and/or used as a predictor in VarBC?

jeromebarre: This would avoid creating multiple datasets with not knowing what it's in there after while.

src/compo/viirs_l1bnc2ioda.py
```diff
@@ -185,12 +190,15 @@ def _read(self):
...
                 qcfs[:, ichan] = qcf[:].data.ravel()
-                errs[:, ichan] = 1. + err.scale_factor * err[:].data.ravel() ** 2
+                # Convert Uncertainty Index (in percentage) into absolute error
+                errs[:, ichan] = (1. + err.scale_factor * err[:].data.ravel() ** 2) * 0.01 * vals[:, ichan]
```
jeromebarre: Is this formula from the VIIRS L1b ATBD? Or is it something you came up with? Either specify where does it come from or justify it in a comment above it birefly.

src/compo/viirs_l1bnc2ioda.py
weiwilliam: That is a good idea. I feel it can be confused so I added the global attribute to tell whether the data is true reflectance or stored reflectance. How do I add the scaling on observation? I'm working on hofx in UFO/CRTM.

jeromebarre: Load SZA in UFO, calculate the secant term and apply it.

src/compo/viirs_l1bnc2ioda.py
jeromebarre: Following our meeting can the error be scaled to account for the secant term instead of multiplying O in the converter and then B in UFO?
Also, would an obs function be more appropriate for applying the secant term either on obs error or O-B ?
@BenjaminRuston is there such a thing in obs functions?

weiwilliam: I will remove secant term from obs value and apply it to obs error.

jeromebarre (approved):
Looks good now.

BenjaminRuston (approved):
thanks @weiwilliam for making the changes this looks ready to go

PR #1625, opened 2025-03-11
Feature/gnssro bufr satsubid

steve-marshall-piq:
## Description

This modifies the python script gnssro_bufr2ioda.py, which is used to convert Radio Occultation data from BUFR to IODA NetCDF4 format.

It is a common problem that satelliteSubIdentifier is missing from standard WMO BUFR data. This variable was a later addition to the BUFR standard for Radio Occultation (RO) data, and at least some processing centers do not provide this variable. This variable provides the number of the LEO (receiver) satellite. It is needed for traceability in order to know which satellite produced a particular occultation observation, and is relevant  to quality control.  It is particularly important for commercial RO data where there are multiple receiver satellites associated with a single mission (satelliteIdentifier).

The satelliteSubIdentifier is missing from all BUFR files processed by UCAR on data.cosmic.ucar.edu. While UCAR intends to fix this, this has not occurred yet and the fix is not scheduled.  Also, historical reprocessing of old data may never be done. This fix bridges that gap, allowing satelliteSubIdentifer to be intelligently filled-in.  The filename parsing works with the DAAC filename conventions used on data.cosmic.ucar.edu and other similar filenaming conventions.

This checkin also includes a line command option to disable tangent point drift by using the average tangent point for all occultations in the profile.  This is useful for experiments testing the importance of handling tangent point drift in GNSSRO forward operators. This causes no change in behavior unless the --no-tp-drift option is invoked.

## Issue(s) addressed

This does not have an associated issue number. I am new to JEDI development and I have had trouble understanding how to gain access to the issue-tracking system. However, I am eager to learn and would like to follow these processes.

## Dependencies

- None

## Impact

None.  The only difference is that satelliteSubIdentifier may have non-missing values in cases where it was missing previously. This variable is not used by any existing GNSSRO forward operator; it is simply passed along for traceability.

## Checklist

- [X ] I have performed a self-review of my own code
- [X ] I have made corresponding changes to the documentation
- [ X] I have run the unit tests before creating the PR

### DIFF
--- a/src/gnssro/gnssro_bufr2ioda.py
+++ b/src/gnssro/gnssro_bufr2ioda.py
@@ -45,12 +45,12 @@ locationKeyList = [
 # The first parentheses in the pattern should match the satId
 # and the second parentheses should match the number (subidentifier).
 # Satellite subidentifiers are typically not needed for government
-# missions. For example, Cosmic1 (C[0-9]{3}) and Cosmic2 (C2E[0-9]) 
+# missions. For example, Cosmic1 (C[0-9]{3}) and Cosmic2 (C2E[0-9])
 # do not use subidentifiers.
 satIdPatterns = [
-    '^(GN)([0-9]{2})$',     # PlanetiQ
-    '^(S)([0-9]{3})$',      # Spire
-    '^(GO)([0-9]{2})$'      # GeoOptics
+    r'^(GN)([0-9]{2})$',     # PlanetiQ
+    r'^(S)([0-9]{3})$',      # Spire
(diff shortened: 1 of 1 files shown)


BenjaminRuston:
ok thanks @steve-marshall-piq 

I will need to checkout your branch and bring into internal making a PR on your behalf... first glance looks straightforward

BenjaminRuston:
thanks @steve-marshall-piq we have picked up your branch and put in [PR 1630](https://github.com/JCSDA-internal/ioda-converters/pull/1630)on your behalf

please reopen this if you feel there is more needed

PR #1654, opened 2025-05-12
Add IMS stationIdentification to output

### LINKED ISSUE #1642: Add location IDs to IMS IODA files
## Description

The IMS IODA files need a location ID for each observations, so that certain locations can be filtered out. 

## Requirements

A UFO filter can be used to exclude IMS obs from certain locations.

Completed a test in which the location IDs are added to the IODA file, then the UFO filters out a selected location.

## Dependencies

The IMS_processing code (external to JEDI) needs to be update first, to include a location ID with the grid cell information contained in it.

yuanxue2870:
## Description
1.  Revise errsd from 40 to 80 to be consistent with offline workflow. Confirmed IMS errsd value via email exchanges on 05/12/2025.
2.  Based on Youlong's recent PR in https://github.com/NOAA-EMC/land-SCF_proc/pull/1#, now we need to output IMS station IDs (i.e., stationIdentification as a nine-digit string) to the ioda formatted output to be used for follow-up QC.

## Issue(s) addressed

Resolves #1642 

## Dependencies

- https://github.com/NOAA-EMC/land-SCF_proc/pull/1#

## Impact

Expected impact on downstream repositories: No significant impacts are expected. More IMS QCs can be done based on this added IMS ID feature. The input and output to the ctest of "test_iodaconv_imsfv3grid_scf" needs to be changed, otherwise, the ctest will fail.

## Checklist

### DIFF
--- a/src/land/imsfv3_scf2ioda.py
+++ b/src/land/imsfv3_scf2ioda.py
@@ -84,14 +84,17 @@ class imsFV3(object):
         ncd = nc.Dataset(self.filename)
         lons = ncd.variables['lon'][:].ravel()
         lats = ncd.variables['lat'][:].ravel()
+        stid = ncd.variables['stid'][:].ravel()
         oros = ncd.variables['oro'][:].ravel()
         sncv = ncd.variables['IMSscf'][:].ravel()
         sncv[sncv == -999.] = float_missing_value
         sndv = ncd.variables['IMSsnd'][:].ravel()
         sndv[sndv == -999.] = float_missing_value
+        strstid = np.empty_like(lons, dtype=object)
 
         lons = lons.astype('float32')
         lats = lats.astype('float32')
+        stid = stid.astype('int64')
         oros = oros.astype('float32')
         sncv = sncv.astype('float32')
         sndv = sndv.astype('float32')
@@ -104,6 +107,9 @@ class imsFV3(object):
 
         times = get_observation_time(self.filename, sncv, ncd)
 
+        for i in range(len(lons)):
+            strstid[i] = str(stid[i])
+
         ncd.close()
 
         # add metadata variables
@@ -112,6 +118,7 @@ class imsFV3(object):
         self.varAttrs[('dateTime', metaDataName)]['_FillValue'] = long_missing_value
         self.outdata[('latitude', metaDataName)] = lats
         self.outdata[('longitude', metaDataName)] = lons
+        self.outdata[('stationIdentification', metaDataName)] = strstid
         self.outdata[('stationElevation', metaDataName)] = oros
         self.varAttrs[('stationElevation', metaDataName)]['units'] = 'm'
 
--- a/src/land/imsfv3_scf2ioda.py
+++ b/src/land/imsfv3_scf2ioda.py
@@ -100,7 +100,7 @@ class imsFV3(object):
         qdflg = 0*sndv.astype('int32')
         errsc = 0.0*sncv
         errsd = 0.0*sndv
-        errsd[:] = 40.
+        errsd[:] = 80.
 
         times = get_observation_time(self.filename, sncv, ncd)
 
--- a/test/testinput/ascii_input_IMSscf.20230501.oro_C48.nc
+++ b/test/testinput/ascii_input_IMSscf.20230501.oro_C48.nc
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:14a6de18ae11e9c965d8158c071dcd0715f76904c395b38696f6978da8e3f590
-size 120681
+oid sha256:460bc5e98d2f41bee907fbe81f149b31ba7e9210a578deba0cf35ddc85786efb
+size 132463
--- a/test/testinput/nc_input_IMSscf.20230501.oro_C48.nc
+++ b/test/testinput/nc_input_IMSscf.20230501.oro_C48.nc
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:e158ec9dccf195cd60e8fa9b01555ed0f0a53a5507e667b10aee36484c8848b0
-size 120783
+oid sha256:d1e665ac789c9b770108c046476dfa76bb2839dc7291b70755d9c0f1adeb1902
+size 132565
--- a/test/testoutput/imsfv3_scf.nc
+++ b/test/testoutput/imsfv3_scf.nc
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:93aa715c557f9308c767ecae2208f6c63d579bfe87a3834820042566cb6046a6
-size 66041
+oid sha256:4e78f36a96811e67a4909b3691eec46afb524fa1bf3fb782b100ee5a0822d293
+size 207133
--- a/test/testoutput/imsfv3grid_scf.nc
+++ b/test/testoutput/imsfv3grid_scf.nc
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:087cc14684641450e26886fdd26f26bcc6a73d831c2433a435fc880f2d926255
-size 66041
+oid sha256:4146d780b62df4c53c5b239e55c1a0eb1fdc9bb3e9ea1a8989db9e5a884dc390
+size 207133


yuanxue2870:
Please review: @BenjaminRuston, @YoulongXia-NOAA, @CoryMartin-NOAA, Thank you!

CoryMartin-NOAA:
@yuanxue2870 please update the test files as part of this PR, if you need help with that, let me know

YoulongXia-NOAA:
@yuanxue2870, test files include both input and output files. please keep consistency with ioda converters input and output filenames. Otherwise, you also need to modify .txt file to do the ioda-converter test.

yuanxue2870:
> @yuanxue2870 please update the test files as part of this PR, if you need help with that, let me know

Thanks Youlong and Cory for your comments. I have already created the new test input and output files on Hera, also included their paths on this PR. The administrator can grab these new files and replace with the old ones directly (i.e., file names are consistent). Unless there is a public place that I can drop off these two files?

CoryMartin-NOAA:
The IODA converters repository uses "git-lfs" so they should be committed to this PR and replace the existing files

yuanxue2870:
> The IODA converters repository uses "git-lfs" so they should be committed to this PR and replace the existing files

Oh, I see. That is why it is different from other repos... Thanks for your clarification! I will do that. Convert to draft while uploading new test files...

yuanxue2870:
Please re-review: @BenjaminRuston, @YoulongXia-NOAA, @CoryMartin-NOAA, Thank you!

CoryMartin-NOAA (approved):

jiaruidong2017 (approved):

src/land/imsfv3_scf2ioda.py
```diff
@@ -84,14 +84,17 @@ def _read(self):
...
         oros = ncd.variables['oro'][:].ravel()
         sncv = ncd.variables['IMSscf'][:].ravel()
         sncv[sncv == -999.] = float_missing_value
         sndv = ncd.variables['IMSsnd'][:].ravel()
         sndv[sndv == -999.] = float_missing_value
+        strstid = np.empty_like(lons, dtype=object)
 
         lons = lons.astype('float32')
         lats = lats.astype('float32')
+        stid = stid.astype('int64')
```
YoulongXia-NOAA, line 97: stid = stid.astype('int32')

YoulongXia-NOAA: @yuanxue2870 I recalled that only datetime use the double precision. For station ID, single precision is sufficient.

yuanxue2870: Hi Youlong, thanks for your comments. For now, I do not think int32 or int64 would make a difference here as the "stid" is a nine-digit integer, and not being output to the IODA formatted as well. However, int32 will not be enough if we goes to higher resolution (e.g., C1152), where we may need a 11 to 12-digit integer, which will exceed int32 storage capacity. Hence, I would like to keep int64 here. - Thanks,Yuan

YoulongXia-NOAA: Okay, keep it as suggested, @yuanxue2870. I am fine with it but this is the first time I found that people uses int64 for station ID, a bit weird.

YoulongXia-NOAA (approved):
I approved this PR.

src/land/imsfv3_scf2ioda.py
```diff
@@ -112,6 +118,7 @@ def _read(self):
         self.varAttrs[('dateTime', metaDataName)]['_FillValue'] = long_missing_value
         self.outdata[('latitude', metaDataName)] = lats
         self.outdata[('longitude', metaDataName)] = lons
+        self.outdata[('stationIdentification', metaDataName)] = strstid
```
srherbener, line 121: Am I understanding correctly that stid itself, as an integer, is not written to the output file. Rather it is converted to a string and the string value is written out in the stationIdentification value? If so, I think it's okay for stid to be a 64-bit integer.

Pardon my ignorance of this obs type, but I am curious why ~4 billion values is not enough unique station id values. If this obs type is going to have very large numbers of locations, we might want to consider using a numeric station id, or be very careful to use a character array instead of a variable length string, since the variable length string storage and access in the file is going to be very slow. Not for this PR but something to think about for the future.

srherbener (commented):
Thanks for this update!

src/land/imsfv3_scf2ioda.py
YoulongXia-NOAA: @srherbener, as @ClaraDraper-NOAA would like to filter out some processed snow depth values at some IMS grids, and she asked me to add a staion ID as txxxxyyy, t is tile number for FV3 grid, 1-6, xxxx is the number for the longitude, and yyyy is number of latitude, e.g., C1152,  4608x2406  grids, xxxx from 0001 to 4608, yyyy 0001 to 2406, therefore, it is a 9-digit integer. In general, stid use string for ioda-converter. Are you thinking to use an integer instead?

yuanxue2870: @srherbener Yes, you are correct. Only strid is written out.
Thanks for your suggestion. IMS is gridded obs, which is different from conventional "in-situ" (i.e., point-scale) obs. We want to give each grid an ID because we want to do some further QCs afterwards. The way we set up each ID is by concatenating its tile number, x-coordinate number, and y-coordinate number -- so we have super big values for stid. For now, int32 is sufficient. By using int64, I am trying to leave some wiggle room for higher resolution (in the long run) which is likely with an increased concatenated ID based on the current code logic. 
I am happy to adjust if defining 'stid' as an int64 takes too much memory. Thanks!

YoulongXia-NOAA: @srherbener, In future. you suggest to use an integer. but for now, If it is 9 digit string, do you know how slow it will be for a vector 6x4608x2406?

srherbener: @YoulongXia-NOAA the vector you mention is ~66 million entries. If we are reading that many locations into a single ObsSpace, then a variable length string data type on a variable (eg stationIdentification) in the file could be problematic. However, if it's the case where we need the 66 million values to cover all the possible grid cells, but we only read a subset of those grids (say 10's or 100's of thousands of locations) into a single ObsSpace then we should be fine.

CoryMartin-NOAA: For the immediate future the vector will only be 6*1152*1152 or ~8 million. Plus we do thin the obs with a filter at the early stage of the generation of the obs space

srherbener: Does that mean that you would have 8 million locations (one for each grid cell) in a single obs file? If so, that could be problematic if even a handful of variables are variable length string type. This is the reason we moved away from strings for the datetime variable.

CoryMartin-NOAA: hmm, okay, so you're suggesting we should use integer, not string. Is it okay that it would be 'stationIdentification' which should be string according to the IODA convention? or would we need to use a new variable name

yuanxue2870: Please allow me to chime in. I do not think integer vs. string will cause a memory issue here. I might be wrong --please feel free to correct me. We are not outputting millions of location IDs. We are only giving each land point an IMS ID. For example, for C48, we only output a total of 2820 locations (i.e., we are not outputting 6 * 48 * 48 location IDs), each location ID itself is a nine-digit string, converted from an integer stid (which itself is a big number, i.e., a nine-digit integer). For C96, we only output a total of 12252 location ID strings, etc.

CoryMartin-NOAA: oh yeah, it's only land values. So it should be not an issue, even at C1152

YoulongXia-NOAA: @yuanxue2870, you can use your converter to run a case for C1152 for 1 January of a given year, and then count how many IMS grid points the file (snow depth >=0.0) has, and then will know as I thought 1 January can approximately represent maximum number. This will give us an estimation.

CoryMartin-NOAA: I don't think that's necessary, we can approximate it with the fact that C48 is ~4x the grid points of C96, and the ratio of 12252/2820 is ~4.3. So to go from C96 to C1152 is 144x the grid points, which means we should expect around 2 million values

YoulongXia-NOAA: We have total land points for C1152, it is  2381853, if latitude is only 0-90oN, it has 1190926 land points ~ 1 million grid points for land.

YoulongXia-NOAA: If there 10% data is missing, it is nearly ~ 1 million.

yuanxue2870: For C96, we output a total of 12252 location ID strings on December 01. The total number of land points in C96 is 18320, meaning ~67% of the total land points are being output in the IMS-IODA output. For C1152, we have a total number of land points of 2381853, similarly, say when 67% are being output on December 01 (for example), we will output 1595841 ID strings.

srherbener: The issue is not necessarily memory, rather it is runtime. When an hdf5/netcdf file has a large array (vector) of strings, and those strings are the variable length string datatype (which is what stationIdentification is), it can sometimes take a long time to read those strings into memory. When in memory the strings are stored as a vector<string> in C++ and processing (compare, sort, conversion to DateTime objects, etc) those can also be slow, but not as bad as the file I/O.

I think we are in the gray zone for runtime performance with a string vector that contains a million elements. It may still be okay but if so it's probably getting close to where the runtime starts blowing up. It might make sense to:
1. Leave stationIdentification as is for now and see how it performs
2. If the runtime is too slow, then consider using a new integer variable, say "gridIdentification"

If this sounds like a good approach with everyone, you could merge this PR as is (using stationIdentification) and then do some profiling to decide if a numeric id is warranted (and add that in a subsequent PR).

CoryMartin-NOAA: What you propose sounds good to me, thank you @srherbener for the suggestions and insight! (as always!)

yuanxue2870: Sounds good to me. Thanks for your help!

srherbener (approved):

src/land/imsfv3_scf2ioda.py
srherbener: Sounds great. Happy to approve now.

src/land/imsfv3_scf2ioda.py
YoulongXia-NOAA: That is an optimal suggestion, @srherbener, thank you for your deep thoughts. sounds good with me.

yuanxue2870:
Dear admin team, this PR is required for GFSv17 for the correct use of IMS observation errors. It had been approved by four reviewers so far. Please consider merging the PR when you get a chance. Feel free to let me know if there is anything needed on my end. Thank you! cc: @CoryMartin-NOAA @ClaraDraper-NOAA

ClaraDraper-NOAA (approved):

BenjaminRuston:
@srherbener you still have commit privilege on this repository as well correct?

I will watch and get this merged pending CI and local tests thanks

BenjaminRuston:
@srherbener, @yuanxue2870 , @CoryMartin-NOAA and @ClaraDraper-NOAA 

I guess the only remaining issue is this is very bespoke and there is no IODA convention for such a thing as `tile` and the `coordinates`

@ClaraDraper-NOAA in particular, in the bigger picture the goal is to IODA-ify the pre-processing and then JEDI would implicitly know these values already. Am I correct in this assumption?

CoryMartin-NOAA:
Bigger picture is that we have a preprocessor that provides the input to the ioda-converter here, but we need to refactor it so that it uses IODA directly to write out a compliant file.

BenjaminRuston:
the CI also is not going to pass unless this is an internal branch

at this point should I just make a parallel PR for this?

yuanxue2870:
> the CI also is not going to pass unless this is an internal branch
> 
> at this point should I just make a parallel PR for this?

Sure. Please feel free to make a parallel PR. Thank you for your help!

PR #1672, opened 2025-07-09
obs2ioda alignment 

### LINKED ISSUE #1670: obs2ioda alignment with ncar
## Description
the obs2ioda converter is still in use by at a minimum JCSDA and NCAR MMM though the user base is likely wider 

there have been bugfixes going into both the JCSDA-internal and NCAR version:  https://github.com/NCAR/obs2ioda

which is inefficient, putting a single aligned repository for this is in the best interest.  having it in JCSDA-internal makes sense as can leverage the unit test structure as well as the IODA convention file:
https://github.com/JCSDA-internal/ioda/blob/develop/share/ioda/yaml/validation/ObsSpace.yaml

a new obs2ioda.x executable this will deprecate:
https://github.com/JCSDA-internal/ioda-converters/tree/develop/src/ncar-bufr2nc-fortran

BenjaminRuston:
## Description

this brings in updated obs2ioda which can now be aligned and maintained as a unified repository incorporating bug fixes from all developers

## Issue(s) addressed

Resolves #1670 

## Impact

single unified obs2ioda repository


BenjaminRuston:
@amstokely there was a file it complained about as the git-lfs tracking wasn't correct:
```
Encountered 1 file that should have been a pointer, but wasn't:
	src/ncar/obs2ioda/test/validation/data/goes_abi/write_ioda_v3/reference/write_goes_abi_ioda_v3.nc
```

for now I set it to ignore git-lfs by altering the `.gitattributes` but think we can fix this with cleanup of the branch now that it's been moved to internal

ibanos90:
Hi @BenjaminRuston, thank you for your efforts on this! 

As we work toward the unification of the code under ioda-converters, I wanted to highlight that, historically, obs2ioda has been very easy to compile with minimal dependencies. It would be ideal to preserve this simplicity. 

My question is, would it be possible to add an option to compile obs2ioda as a standalone component? Given that obs2ioda do not depend on any other ioda/oops code, could we add a CMake argument to enable standalone compilation?

BenjaminRuston:
@ibanos90 thank you and I am not opposed, there is another issue open where we're trying to help someone build the ioda-converters from the public jedi-bundle.  Here we recommended to comment out a block of lines specifically these lines 60-73:
https://github.com/JCSDA/jedi-bundle/blob/47c9177d337cd2a8426eb09e1158679b7379cead/CMakeLists.txt#L60-L73

would that be acceptable to put into a conditional driven by a CMake argument

ibanos90:
> @ibanos90 thank you and I am not opposed, there is another issue open where we're trying to help someone build the ioda-converters from the public jedi-bundle. Here we recommended to comment out a block of lines specifically these lines 60-73: https://github.com/JCSDA/jedi-bundle/blob/47c9177d337cd2a8426eb09e1158679b7379cead/CMakeLists.txt#L60-L73
> 
> would that be acceptable to put into a conditional driven by a CMake argument

@BenjaminRuston I believe this will still build the whole ioda-converters, correct?
Would it be possible to build just obs2ioda, without having to compile the rest of the ioda-converters code? I guess with current capabilities it may not be possible, right?

BenjaminRuston:
> Would it be possible to build just obs2ioda, without having to compile the rest of the ioda-converters code?

hmmm, I'm not sure about that one, and yes @ibanos90, you're correct something that would skip L60-L73 in the CMakeLists.txt example above would still make all of ioda-converters

when you make your bundle sounds like you typically don't make ioda-converters then... let me have a think on this and see if have any other insights

amstokely:
> > Would it be possible to build just obs2ioda, without having to compile the rest of the ioda-converters code?
> 
> hmmm, I'm not sure about that one, and yes @ibanos90, you're correct something that would skip L60-L73 in the CMakeLists.txt example above would still make all of ioda-converters
> 
> when you make your bundle sounds like you typically don't make ioda-converters then... let me have a think on this and see if have any other insights

@ibanos90 Ideally, it would be nice to retain the ability to build only obs2ioda. However, I don’t think that’s feasible anymore since obs2ioda now relies on ioda, particularly for access to the ObsSpace.yaml. I previously kept a copy of that file within obs2ioda, but now that it's part of a library with a direct dependency on ioda, maintaining a separate copy doesn't make much sense.

ibanos90:
> > > Would it be possible to build just obs2ioda, without having to compile the rest of the ioda-converters code?
> > 
> > 
> > hmmm, I'm not sure about that one, and yes @ibanos90, you're correct something that would skip L60-L73 in the CMakeLists.txt example above would still make all of ioda-converters
> > when you make your bundle sounds like you typically don't make ioda-converters then... let me have a think on this and see if have any other insights
> 
> @ibanos90 Ideally, it would be nice to retain the ability to build only obs2ioda. However, I don’t think that’s feasible anymore since obs2ioda now relies on ioda, particularly for access to the ObsSpace.yaml. I previously kept a copy of that file within obs2ioda, but now that it's part of a library with a direct dependency on ioda, maintaining a separate copy doesn't make much sense.

I see, thanks @amstokely

fabiolrdiniz:
Related to @ibanos90's point, would it be a problem to have a makefile inside the folder? And/or the instructions inside the README file on how to build a standalone version of the converter? I'm more inclined towards the latter. Regardless of the chosen option, it would be helpful to have some information about these details included in the README file.

ibanos90:
> Related to @ibanos90's point, would it be a problem to have a makefile inside the folder? And/or the instructions inside the README file on how to build a standalone version of the converter? I'm more inclined towards the latter. Regardless of the chosen option, it would be helpful to have some information about these details included in the README file.

That would be great! Thanks for looking into this, @fabiolrdiniz!

amstokely:
@eap This PR introduces  `yaml-cpp` as a dependency, which is built via `FetchContent`. Since this does not require users to install an additional dependency, do you see this being a problem?

BenjaminRuston:
@amstokely needs reviews to merge!  but this appears ready to go (for this initial version)

src/ncar/obs2ioda/src/cxx/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+FetchContent_Declare(
+        yaml-cpp
+        GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
+        GIT_TAG 0.8.0
```
BenjaminRuston, line 4: @eap this is  a new dependency, as believe the eckit handle the yaml parsing

not this is a first implementation of this new `obs2ioda` and would be nice if can make this common.

BenjaminRuston:
@eap added you mostly if you could scan the CMake files, and make sure nothing may cause issues

this brings this converter inline with IODAv3 which is great, and we expect there will likely be some additional changes needed on top of this but any major roadblocks can be addressed now and other issues let's tackle in smaller chunks if you agree

amstokely (approved):

src/ncar/obs2ioda/src/cxx/CMakeLists.txt
amstokely: While I'm not familiar with eckit, it should be straightforward for someone who is to replace the YAML parsing code with an eckit-based interface equivalent to the current [yaml-cpp interface](https://github.com/JCSDA-internal/ioda-converters/blob/de6cb65060b755afb6646d9a61cece0cc183f1f1/src/ncar/obs2ioda/src/cxx/ioda_obs_schema.h#L20).

amstokely:
The new [C++](https://github.com/NCAR/obs2ioda/tree/feature/obs2ioda_alignment/test/cxx) and[ Fortran ](https://github.com/NCAR/obs2ioda/tree/feature/obs2ioda_alignment/test/fortran)wrapper I/O is very well tested. The tests are not included in this PR as they introduce more dependencies (`googletest` and `pFUnit`) that are outside the JEDI ecosystem.

src/ncar/obs2ioda/.gitignore
```diff
@@ -0,0 +1,7 @@
+
+.DS_Store
+.DS_Store
+**/__pycache__
+test/validation/test_utils/FilePathConfig.py
+.obs2ioda
+build
```
climbfuji, line 7:
```suggestion
build

```

src/ncar/obs2ioda/cmake/Obs2Ioda_CompilerFlags.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+# Set Debugging Fortran compiler flags for the Intel Compiler
+# -check uninit: Checks uninitialized variables
+# -ftrapuv: Enable trapping of uninitialized variables
+# -g: Enable production of debug information
+# -traceback: Give symbolic traceback on errors
+# -fpe0: Stop execution when a floating-point exception occurs
+set(FORTRAN_COMPILER_INTEL_DEBUG_FLAGS
+    $<$<COMPILE_LANGUAGE:Fortran>:-check uninit -ftrapuv -g -traceback -fpe0>
+)
```
climbfuji, line 36:
```suggestion
)

```

src/ncar/obs2ioda/src/cxx/CMakeLists.txt
climbfuji: I checked the license, yaml-cpp has an MIT license and it can be used in an Apache 2 licensed product like JEDI

climbfuji (commented):
What was the reasoning for adding this de-facto separate piece of software to ioda-converters with all the complications for the build system etc?

Why not make ncar obs2ioda its own package/repository?

snyderdad:
@climbfuji : That approach (own repo for obs2ioda) would be OK w/ NCAR.  Perhaps preferable if it eases or enables a stand-alone build.

climbfuji:
> @climbfuji : That approach (own repo for obs2ioda) would be OK w/ NCAR. Perhaps preferable if it eases or enables a stand-alone build.

Modularity certainly helps with the spack-ification of the JEDI packages, and looking at how complicated the build integration between ioda-converters (ecbuild) and ncar obs2ioda (standard cmake) is, I think it would be better altogether.

amstokely:
> > @climbfuji : That approach (own repo for obs2ioda) would be OK w/ NCAR. Perhaps preferable if it eases or enables a stand-alone build.
> 
> Modularity certainly helps with the spack-ification of the JEDI packages, and looking at how complicated the build integration between ioda-converters (ecbuild) and ncar obs2ioda (standard cmake) is, I think it would be better altogether.


@BenjaminRuston @climbfuji @snyderdad  If it’s important for JCSDA to maintain `obs2ioda` within `ioda-converters`, it should be relatively straightforward to convert its CMake build system to use `ecbuild`. While I don’t have time to do this myself, it would likely take only a few days of work for someone on the `ioda` team to make the necessary changes.

BenjaminRuston:
> @BenjaminRuston @climbfuji @snyderdad If it’s important for JCSDA to maintain obs2ioda within ioda-converters, it should be relatively straightforward to convert its CMake build system to use ecbuild. While I don’t have time to do this myself, it would likely take only a few days of work for someone on the ioda team to make the necessary changes.

My strategy will be to open an issue to do the conversion and since this would be a different person would propose that work be done in a separate issue and PR

@climbfuji or @eap is there any unforeseen build issues this would cause for other broader users if not this can continue from this initial contribution

climbfuji:
> > @BenjaminRuston @climbfuji @snyderdad If it’s important for JCSDA to maintain obs2ioda within ioda-converters, it should be relatively straightforward to convert its CMake build system to use ecbuild. While I don’t have time to do this myself, it would likely take only a few days of work for someone on the ioda team to make the necessary changes.
> 
> My strategy will be to open an issue to do the conversion and since this would be a different person would propose that work be done in a separate issue and PR
> 
> @climbfuji or @eap is there any unforeseen build issues this would cause for other broader users if not this can continue from this initial contribution

I think this is going into the wrong direction. Instead of converting standard cmake to ecbuild, we should be doing the opposite ... but that's just me. That being said, both ioda-converters and obs2ioda being ecbuild is still cleaner than a mix. Most important is that all source code for obs2ioda is included in the ioda-converters repository and there are no "hidden" downloads of code during a `cmake`, `ecbuild`, or `make` step.

amstokely:
> > > @BenjaminRuston @climbfuji @snyderdad If it’s important for JCSDA to maintain obs2ioda within ioda-converters, it should be relatively straightforward to convert its CMake build system to use ecbuild. While I don’t have time to do this myself, it would likely take only a few days of work for someone on the ioda team to make the necessary changes.
> > 
> > 
> > My strategy will be to open an issue to do the conversion and since this would be a different person would propose that work be done in a separate issue and PR
> > @climbfuji or @eap is there any unforeseen build issues this would cause for other broader users if not this can continue from this initial contribution
> 
> I think this is going into the wrong direction. Instead of converting standard cmake to ecbuild, we should be doing the opposite ... but that's just me. That being said, both ioda-converters and obs2ioda being ecbuild is still cleaner than a mix. Most important is that all source code for obs2ioda is included in the ioda-converters repository and there are no "hidden" downloads of code during a `cmake`, `ecbuild`, or `make` step.

@climbfuji I was suggesting converting the cmake build to ecbuild solely for consistency reasons, not because I like ecbuild :). Without using `FetchContent`, wouldn't that mean that all `ioda-converter` users would have to install `yaml-cpp` in their JEDI environment? `yaml-cpp` is very easy to install, but I'm always hesitant to require people to manually install a new library that is outside the JEDI spack-stack environment.

climbfuji:
> > > > @BenjaminRuston @climbfuji @snyderdad If it’s important for JCSDA to maintain obs2ioda within ioda-converters, it should be relatively straightforward to convert its CMake build system to use ecbuild. While I don’t have time to do this myself, it would likely take only a few days of work for someone on the ioda team to make the necessary changes.
> > > 
> > > 
> > > My strategy will be to open an issue to do the conversion and since this would be a different person would propose that work be done in a separate issue and PR
> > > @climbfuji or @eap is there any unforeseen build issues this would cause for other broader users if not this can continue from this initial contribution
> > 
> > 
> > I think this is going into the wrong direction. Instead of converting standard cmake to ecbuild, we should be doing the opposite ... but that's just me. That being said, both ioda-converters and obs2ioda being ecbuild is still cleaner than a mix. Most important is that all source code for obs2ioda is included in the ioda-converters repository and there are no "hidden" downloads of code during a `cmake`, `ecbuild`, or `make` step.
> 
> @climbfuji I was suggesting converting the cmake build to ecbuild solely for consistency reasons, not because I like ecbuild :). Without using `FetchContent`, wouldn't that mean that all `ioda-converter` users would have to install `yaml-cpp` in their JEDI environment? `yaml-cpp` is very easy to install, but I'm always hesitant to require people to manually install a new library that is outside the JEDI spack-stack environment.

That raises another question. Since JEDI uses `eckit` for parsing yaml files, does it make sense to introduce another dependency for doing the same task? I understand that while obs2ioda was developed in isolation outside of the JEDI system, using `yaml-cpp` made a lot of sense. But now that the `obs2ioda` code is integrated in the JEDI codebase, wouldn't it make more sense to use `eckit`?

BenjaminRuston:
believe we are largely in agreement here.  for me still happy for suggestions on order of things.

there is a bit of a handoff as it goes to the next stage, which is the `converting standard cmake to ecbuild`.  This then also brings the ability to remove `yaml-cpp`.

I am panhandling for 2cents @eap and @climbfuji if it is too messy to keep that dependency now and to execute the next stage subsequently? 

With the `FetchContent` to be deprecated once verified can use other pre-existing method (withholding extra discussion)

amstokely:
> believe we are largely in agreement here. for me still happy for suggestions on order of things.
> 
> there is a bit of a handoff as it goes to the next stage, which is the `converting standard cmake to ecbuild`. This then also brings the ability to remove `yaml-cpp`.
> 
> I am panhandling for 2cents @eap and @climbfuji if it is too messy to keep that dependency now and to execute the next stage subsequently?
> 
> With the `FetchContent` to be deprecated once verified can use other pre-existing method (withholding extra discussion)

@climbfuji @BenjaminRuston Updating the YAML parsing code to use eckit instead of yaml-cpp should be fairly straightforward for someone familiar with eckit (I’m not). The key step would be implementing an interface equivalent to the current yaml-cpp one used [here](https://github.com/JCSDA-internal/ioda-converters/blob/de6cb65060b755afb6646d9a61cece0cc183f1f1/src/ncar/obs2ioda/src/cxx/ioda_obs_schema.h#L20).

The NCAR repo includes a [Googletest-based test suite](https://github.com/NCAR/obs2ioda/blob/feature/obs2ioda_alignment/test/cxx/ioda_obs_schema.test.cc) for the YAML parsing code that could help validate the new implementation. If the test suite is to be used in production under eckit, it would need to be adapted accordingly.

srherbener:
Here's how I see this situation. The main issue that needs to be fixed is that we currently have (at least) 4 evolutionary copies of the "obs2ioda" converter, starting with the original Jamie Bresch converter along with various snapshots and ending with Andy's obs2ioda converter. The goal is to obsolete the older snapshots and move forward with only Andy's obs2ioda converter. The existence of 4 or more snapshots of the same converter is not supportable and maintainable.

I see two options:
1. Keep the converter in NCAR which means that NCAR is responsible for its maintenance and development
2. Move the converter to JEDI (ioda-converters repo) which means that JCSDA and partners are responsible for its maintenance and development

I think the deciding factor for picking one of the options, is how important it is to keep this converter stand alone. If that is the highest priority then the converter should stay in NCAR (option 1). You would be free to continue with cmake, and use whatever library dependencies you see fit.

We started this PR because we were assuming that the highest priority was to consolidate the effort to support and develop this converter by moving it to JEDI, with the hope of spreading out the responsibilities (among the several people that are supporting the various snapshots) because there is too much demand on Andy's time. If this is the correct priority, then we should continue with option 2. However, with option 2, we need to pull obs2ioda into the JEDI system meaning that we should not introduce new library dependencies for functionality we already have in spack-stack. This means that we should convert obs2ioda to a oops::Application, which will facilitate using eckit::Configuration, and we should use the eckit/fckit test framework, etc.

This also means that we shouldn't force it to be a standalone converter. You would be able to build it by cloning jedi-bundle, and commenting out the ecbuild_bundle commands for every package except oops, ioda, ioda-data and ioda-converters. This will still build quickly.

Also, ecbuild is simply a set of cmake macros, and it runs cmake under the hood. It may not be all that much work (if any) to build with ecbuild instead of cmake. This is primarily due to the fact that the top-level CMakeLists.txt file is already in ioda-converters and that is where the bulk of "ecbuidl" flavor configuration resides. 

If we agree that option 2 is the right choice, then I would advocate for biting the bullet and doing the conversion of obs2ioda into a JEDI (OOPS) style application, etc from the start. I think that might actually turn out to be less effort than trying to add new dependencies (and manage those dependencies) to spack-stack.

What do others think about this?

BenjaminRuston:
@srherbener we are going with option2,

think we've been using option1 for the last few years and have decided it's not in everyone's best interest.  This has been discussed with the MMM group and we have agreement.  

The decision now is move towards ecbuild from standard CMake, and this will also allow removal of yaml-cpp dependency

ability to allow for a stand-alone build could be possible most simply with the addition of a Makefile which would not be executed by majority of partners but could allow a build of the single component.  This is something that is not encouraged broadly only for special circumstances, we are deeming so in this case

BenjaminRuston:
thanks @rajichidamb , and yes let's proceed with current track which is to have this in parallel with the other version

for now the `FetchContent` is done so that all ioda-converter users do not have to install yaml-cpp independently in their JEDI environment

this work will be picked up by JCSDA to do the `ecbuild` conversion of the CMake and then we can proceed with a few leftover bits: 
- verifying this against the to be deprecated `bufr2nc_fortran.x`
- making sure one more time bug-fixes from MMM and JCSDA are incorporated
- method for stand-alone build,, currently plan is a largely to be ignored Makefile in the directory with appropriate modifications to documentation about why it's there and how to use it -- default ignore it

fcvdb:
I built the code on Hercules (GNU spack-stack 1.9) and ran the converters (`bufr2nc_fortran.x prepbufr.gdas.2021021600.nr`) on this prepbufr file (from NCAR RDA derecho: `/glade/campaign/collections/rda/data/d337000/prepnr/2021/prepbufr.gdas.2021021600.nr`). The converter built and completed. 

Using `ncdiff`, I compared the generated IODA files (`obs.aircraft_obs_2021021600.nc4, obs.ascat_obs_2021021600.nc4, obs.radiosonde_obs_2021021600.nc4, obs.satwind_obs_2021021600.nc4`) withe same files generated from the develop branch. The files are identical.

fcvdb (approved):
Thanks Ben!

src/ncar/obs2ioda/cmake/Obs2Ioda_CompilerFlags.cmake
```diff
@@ -0,0 +1,36 @@
...
+# -ffree-line-length-none: Remove the limit on the length of lines in the source file
+set(FORTRAN_COMPILER_GNU_FLAGS
+    $<$<COMPILE_LANGUAGE:Fortran>:-ffree-line-length-none>
+)
+
+# Set Fortran compiler flags specific to the GNU Compiler and Linux OS.
+# -mcmodel=medium: Allow for larger datasets in memory
+set(FORTRAN_COMPILER_GNU_LINUX_FLAGS
+        $<$<COMPILE_LANGUAGE:Fortran>:-mcmodel=medium>
+)
```
fabiolrdiniz, line 11: I'm seeing an error message related to this change: https://github.com/JCSDA-internal/ioda-converters/issues/1682

Issue #1669, opened 2025-07-07
bufr2nc_fortran.x crashes on certain prepbufr files


fcvdb:
## Current behavior (describe the bug)

While running Skylab-ingest, the `convertObservations_prepbufr` task sometimes segfaults with the message:

`forrtl: error (65): floating invalid`

This is for instance the case for cycles:
20240902T120000Z
20240923T120000Z
20240928T000000Z

## To Reproduce

> What computer are you running on?

   Hercules

> What compilers/modules are you using?

The issue occurs with both intel and GNU compilers.

> Steps to reproduce the behavior

1. Build the Skylab with either Intel of GNU
2. Edit the config file [jedi-bundle/skylab/experiments/ingest-observations.yaml](https://github.com/JCSDA-internal/skylab/blob/develop/experiments/ingest-observations.yaml)
    - Set the time to one of the failing cycles, eg:
       init_cycle: 2024-09-02T12:00:00Z
       last_cycle: 2024-09-02T12:00:00Z
    - Comment in all platforms, except the prepbuft, eg:
       - !INCLUDE ${JEDI_WORKFLOW}/skylab/obs/ingest/gdas_prepbufr_ingest.yaml
3. Run the ingest suite:
   ` create_experiment.py  jedi-bundle/skylab/experiments/ingest-observations.yaml`
4. Locate the `convertObservations_prepbufr.1.err` log file in the `ecflow/<jobid>/ingest/obs_prepbufr` directory.
...

## Expected behavior

The converter should generate a IODA file.

fcvdb:
Using the GNU compiler in a Debug mode build, the error message in the convertObservations_prepbufr.1.err indicates an error at line 1329 of routine [prepbufr_mod.f90](https://github.com/JCSDA-internal/ioda-converters/blob/d0b4f976a8d5004b4ac2cb771373fd0e7f9f4782/src/ncar-bufr2nc-fortran/prepbufr_mod.f90#L1310):


```
++ /work2/noaa/jcsda/fvandenb/skylab/jedi_2025070116_develop_gnu_hercules/build/bin/bufr2nc_fortran.x obs.prepbufr.20240902T120000Z

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x1479d8cfdd8f in ???
#1  0x45b334 in calc_qs
[...]

PR #1673, opened 2025-07-10
Fix bufr2nc_fortran.x crash on certain prepbufr files

fcvdb:
## Description

See complete description in issue #1669

### DIFF
--- a/src/ncar-bufr2nc-fortran/prepbufr_mod.f90
+++ b/src/ncar-bufr2nc-fortran/prepbufr_mod.f90
@@ -1307,7 +1307,12 @@ contains
 
       if (.not. ice) then   ! over water only
 
-         es = es_alpha*exp(es_beta*t_c/(t_c + es_gamma))
+        ! exponential cannot be numerically evaluated for t_c < -200C
+        if (t_c .gt. -200.) then
+            es = es_alpha*exp(es_beta*t_c/(t_c + es_gamma))
+        else ! use es(-200C) for all temperature below -200C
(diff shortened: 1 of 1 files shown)


amstokely (approved):

src/ncar-bufr2nc-fortran/prepbufr_mod.f90
```diff
@@ -1307,7 +1307,12 @@ subroutine calc_qs(t, p, qs, wrt_ice)
 
       if (.not. ice) then   ! over water only
 
-         es = es_alpha*exp(es_beta*t_c/(t_c + es_gamma))
+        ! exponential cannot be numerically evaluated for t_c < -200C
+        if (t_c .gt. -200.) then
+            es = es_alpha*exp(es_beta*t_c/(t_c + es_gamma))
+        else ! use es(-200C) for all temperature below -200C
+            es = 3.19E-33
```
BenjaminRuston, line 1314: wow still a pretty small value, but it's OK with this!

looks like this converter doesn't have any gross QC that just tosses unphysical points then

BenjaminRuston (approved):
thanks @fcvdb and thanks for walking me through this and it is noted that no matter the `t_c` everything is considered over water

I consider this a separate thing to be handled by someone willing to reconcile some of the pre-processing done in the creation of the prepBUFR files themselves as that is where the saturation vapor pressure tables also need to be consistent

in short it's too big of an issue to tackle here

amstokely (approved):

Issue #1677, opened 2025-07-28
Speed up the GHCN snow depth converter.


ClaraDraper-NOAA:
## Description

The GCHN snow depth converter is rather slow. For an example of 20191130, it takes about 90 seconds. 
For the offline work flow the converters are called in serial, so we pre-process the obs into IODA files. This has been a nuisance, as the IODA format occasionally changes. 

There is also a --maskout flag that is not really needed, since it is best to always set this to true. With false, the output file has data at all stations in the GHCN data set (130,000) - many of which never have snow obs. With true it includes only stations with data (<10,000). 

This change will make the new spin-up for GEFS land initial conditions with the offline workflow much simpler.

## Requirements

The GHCN ioda converter runs faster, ideally in under ~5-10 seconds. 

## Acceptance Criteria (Definition of Done)

The GHCN ioda converter runs in under ~5-10 seconds, with no change in the results. 

## Dependencies
None.

BenjaminRuston:
closed with this PR:
https://github.com/JCSDA-internal/ioda-converters/pull/1681

PR #1678, opened 2025-07-29
Feature/speedy ghcn

ClaraDraper-NOAA:
## Description

This PR speeds up the ghcn ioda converter. The original version takes about 90 seconds, and this version takes about 1 second.

It also includes some tidy up of the code, and removes the maskout option. 

Additionally, the previous code would exclude data from stations not included in the fix file (the fix files provides the location of the station). In the previous test case, this means just 78 of 2000 stations are saved. Missing stations from the fix file usually means that the wrong fix file has been used, so this PR also adds code to exit if there are missing stations. There is an optional argument that can be used to over-ride this, and simply warn on missing stations.

## Issue(s) addressed

Addresses issue [1677](https://github.com/JCSDA-internal/ioda-converters/issues/1677)
## Dependencies

None.

## Impact

None

## Checklist

- [ x] I have made corresponding changes to the documentation
- [x ] I have run the unit tests before creating the PR

NOTE: see comment on fix files above. The old test had a serious bug, so I have updated it. If I change the code in this PR to exclude stations missing from the fix file I can reproduce the old test.


ClaraDraper-NOAA:
Creating a draft for now as the test is failing. The original code generates an IODA file with only 78 obs, while the updated code has 2000. I think the updated code is correct, but I am still looking into it.

ClaraDraper-NOAA:
FYI: @BenjaminRuston . I've left this as a draft for now, but will convert it once I figure out what is going on with the test.

CoryMartin-NOAA (commented):
do you need to update the test reference, @ClaraDraper-NOAA, now that maskout is removed?

ClaraDraper-NOAA:
> do you need to update the test reference, @ClaraDraper-NOAA, now that maskout is removed?

No. The test used maskout=True, which I've now hardcoded as the only option. This update should have passed the existing test ( I think I know why it's not though).

ClaraDraper-NOAA:
The test was failing due to a bug in the old test. Specifically, the old test excludes data from stations not included in the fix file (the fix files provides the location of the station). In the old test case, this means just 78 of 2000 stations are saved. Missing stations from the fix file usually means that the wrong fix file has been. 

I have replaced the fix file and input file to be from the latest download of the data. Now, of the 2000 stations in the input file 1999 stations are saved (one is removed due to invalid snow depth). I've also added a check that will exit if there are stations missing from the fix files, since we do expect them all to be there.

ClaraDraper-NOAA:
@CoryMartin-NOAA @yuanxue2870 @YoulongXia-NOAA @BenjaminRuston 

This is ready to review (and is needed for the reanalysis prototype runs at PSL).

CoryMartin-NOAA (approved):

yuanxue2870 (approved):
Looks good to me. Thanks!

YoulongXia-NOAA (approved):
look good. I approved this PR.

BenjaminRuston:
@ClaraDraper-NOAA my bad... the branch need to be put into JCSDA-internal!

I can do this for you,, but our CI can't checkout code from NOAA github:
`ClaraDraper-NOAA:feature/speedy_ghcn`

BenjaminRuston:
close as duplicate of https://github.com/JCSDA-internal/ioda-converters/pull/1681

PR #1718, opened 2025-10-21
ahi obs2ioda remove dimension nstring

### LINKED ISSUE #1719: obs2ioda writing extra dimension and attributes
## Description

the NCAR obs2ioda converter when being used for AHI is writing an additional dimension `nstring` which is not used downstream and can cause errors and confusion

this dimension should be removed 

additionally there are attributes `min_datetime` and `max_datetime` that also are unused and can be removed

## Requirements

latest NCAR obs2ioda converter

conversion of all datatypes using obs2ioda remain unchanged and the additional dimension and attributes are removed for all datatypes except conventional

BenjaminRuston:
## Description

the nstring dimension is non-standard and is removed

## Issue(s) addressed

Resolves #1719 

## Impact

IODA file without dimension nstring

## Checklist


src/ncar/obs2ioda/src/ncio_mod.f90
```diff
@@ -148,14 +148,17 @@ subroutine write_obs (filedate, write_opt, outdir, itim, fileExt)
...
+             status = netcdfPutAtt(netcdfID, trim(name_ncdim(i)), val_ncdim(i))
+             status = netcdfAddVar(netcdfID, trim(name_ncdim(i)), NF90_INT, 1, [trim(name_ncdim(i))])
+         end if
       end do
 
       ! define global attributes
-      status = netcdfPutAtt(netcdfID, "min_datetime", xdata(ityp, itim)%min_datetime)
-      status = netcdfPutAtt(netcdfID, "max_datetime", xdata(ityp, itim)%max_datetime)
+!     status = netcdfPutAtt(netcdfID, "min_datetime", xdata(ityp, itim)%min_datetime)
+!     status = netcdfPutAtt(netcdfID, "max_datetime", xdata(ityp, itim)%max_datetime)
```
BenjaminRuston: are these used for anything?  would like to remove these entirely

@amstokely or @ibanos90 have you ever seen these Attributes be used these were new to me

@fabiolrdiniz same question?

src/ncar/obs2ioda/src/ncio_mod.f90
```diff
@@ -148,14 +148,17 @@ subroutine write_obs (filedate, write_opt, outdir, itim, fileExt)
       status = netcdfAddVar(netcdfID, trim(ncname), NF90_INT, 1, [trim(ncname)])
 
       do i = 2, n_ncdim
-         status = netcdfAddDim(netcdfID, trim(name_ncdim(i)), val_ncdim(i), ncid_ncdim(i))
-         status = netcdfPutAtt(netcdfID, trim(name_ncdim(i)), val_ncdim(i))
-         status = netcdfAddVar(netcdfID, trim(name_ncdim(i)), NF90_INT, 1, [trim(name_ncdim(i))])
+         if ( .not. trim(name_ncdim(i)) == 'nstring' ) then
```
BenjaminRuston:
```suggestion
         if ( (write_opt == write_nc_conv) .or. &
       ((write_opt /= write_nc_conv) .and. (trim(name_ncdim(i)) /= 'nstring')) ) then
```

BenjaminRuston:
@ibanos90 do you know why the `nstring` dimension is included?   Do not believe we use it, and causes issues if added for radiance data.  Similarly the attributes `max_datetime` and `min_datetime` we don't use and will remove this can be determined from the `MetaData/dateTime` themselves

@amstokely if you remember anything please let us know but may remove the `nstring` dimension and attributes `max_datetime` and `min_datetime`

ibanos90:
> @ibanos90 do you know why the `nstring` dimension is included? Do not believe we use it, and causes issues if added for radiance data. Similarly the attributes `max_datetime` and `min_datetime` we don't use and will remove this can be determined from the `MetaData/dateTime` themselves
> 
> @amstokely if you remember anything please let us know but may remove the `nstring` dimension and attributes `max_datetime` and `min_datetime`

Hi @BenjaminRuston, I believe the `nstring` dimension and `max_datetime` and `min_datetime` variables may be inherited from older IODA versions and kept in the files when we used the IODA upgraders, but we do not use in any of our filters. I believe they can be safely removed.

amstokely:
@BenjaminRuston @ibanos90 I believe this is a legacy attribute from when fixed-length character arrays were used for strings. Since we now use the NetCDF string type, `nstring` is no longer needed for data representation.

That said, much of the legacy `obs2ioda` code relies on the number of attributes, variables, and dimensions for indexing (unfortunately). Removing `nstring` would reduce the dimension count and could break parts of the code that depend on it. I ran into a similar issue before when removing unused variables caused indexing errors.

I'm not sure what level of testing `ioda-converters` currently has for `obs2ioda`, but the test suite I wrote back when NCAR maintained it caught these issues quickly. It may also affect analysis scripts that use index-based loops dependent on the number of dimensions.

BenjaminRuston:
@amstokely my preference is to try to keep the attribute in memory where possible and just not write to the final file

I spoke with @ibanos90 and we believe it can be removed from the output, and we'll try to monitor the output in our applications.  My preference is to move forward with this shiny new version and make changes as needed.

From everything I've seen so far it is safe to remove `nstring` from the files written, will also remove the min/max dateTime attributes

fcvdb (approved):
Thanks Ben!

BenjaminRuston:
@ibanos90 think this is ready, would be good to have confirmation on your side that it is a lateral move

@amstokely you were right there is something buried in there somewhere appears to be conventional data related so only skipping the `nstring` dimension for non-conventional data types

fabiolrdiniz (approved):
Thanks, @BenjaminRuston!

test/testoutput/amsua_n18_obs_2020110112.nc4
```diff
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:34ba505d99286afae2d3e073cd1c98a3abed83cfdc36ac5dab2f61d57165c6b3
-size 129269
+oid sha256:f4f9d437da9fa0597e3fcef1890cf168d6d2b6782a3e2b8548c7f47476948b39
+size 203243
```
rajichidamb, line 3: This file size seem to be bigger than original.

BenjaminRuston: I can run the `nccopy -u -d 9 -s` on it to reduce it... I did run this on a few of them

you are correct it should be smaller but not sure how they were generated in the first place ;0)

PR #1737, opened 2026-01-30
Add "tier2" label to "test_iodaconv_mrms" test

eap:
Tier 2 tests will not be run on integration tests due to their long runtime. They will be run on the project's unit tests.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1004,6 +1004,7 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_mrms
                   TYPE    SCRIPT
                   ENVIRONMENT "PYTHONPATH=${IODACONV_PYTHONPATH}"
                   COMMAND bash
+				  LABELS tier2
                   ARGS    ${CMAKE_BINARY_DIR}/bin/iodaconv_comp.sh
                           netcdf
                           "${Python3_EXECUTABLE} ${CMAKE_BINARY_DIR}/bin/mrms_grib2ioda.py


test/CMakeLists.txt
```diff
@@ -1004,6 +1004,7 @@ ecbuild_add_test( TARGET  test_${PROJECT_NAME}_mrms
                   TYPE    SCRIPT
                   ENVIRONMENT "PYTHONPATH=${IODACONV_PYTHONPATH}"
                   COMMAND bash
+				  LABELS tier2
```
rajichidamb, line 1007: Does ioda-converters ctest have tier2 or is this the only test under it?

eap: only one so far. I'm going to be using this in several repositories; anywhere that we have tests taking over 60 seconds. Those shouldn't run during integration testing time but can run during tests on the target project.

eap: also note this is distinct from UFO tier2 tests that are managed at build time, although I also plan to add the label there for consistency.

svahl991: Maybe, instead of sending it to tier 2 all by itself, we should disable it and enter an issue, like we did for other failing tests. Whoever cares about this test may want to modify it so that it doesn't take so long and can be reactivated in tier 1, rather than sending it to the purgatory of tier 2. Or do both, send it to tier 2 AND enter an issue for it.

svahl991 (approved):

test/CMakeLists.txt
fmahebert: I favor @svahl991 's suggestion of asking the "owner" of this converter to think about setting up a quicker test. And to disable it until then.

But I'd also be ok merging this as-is.

eap:
@svahl991 - as @fmahebert noted in https://github.com/JCSDA-internal/jedi-ci/pull/52, our approach to tier2 is fragmented. Of the many approaches available to us for tier2 testing, unfortunately we are not using labels anywhere yet. This PR is hopefully a move in the direction of [standardizing](https://xkcd.com/927/) tier2 testing.

> we should disable it and enter an issue ... Whoever cares about this test may want to modify it so that it doesn't take so long and can be reactivated

I disagree with the approach of disabling this test since it does work right now, slowly. It is my hope that moving it to tier2 should be the appropriate motivation for its maintainer to address the speed.

That said, filing an issue is probably justified.

fmahebert (approved):
> I disagree with the approach of disabling this test since it does work right now, slowly. It is my hope that moving it to tier2 should be the appropriate motivation for its maintainer to address the speed.

Fair enough, agreed.

svahl991:
Issue #1738 has been entered.

svahl991:
And #1739, apparently.

eap:
@svahl991 - you're fast. I've closed mine as duplicate.

svahl991:
> @svahl991 - you're fast. I've closed mine as duplicate.

I was about to close mine as duplicate just as you closed yours! Mine is still open though. I referenced yours in mine.

eap:
Attaching to this CI bug to document history on the decision:
 https://github.com/JCSDA-internal/jedi-ci/issues/57

PR #1748, opened 2026-02-24
Merging shuffle and deflate code from NCAR

### LINKED ISSUE #1720: NCAR obs2ioda should investigate deflation and shuffle
## Description

similar to what is done in the GNSS-RO fortran converter:
https://github.com/JCSDA-internal/ioda-converters/blob/f794abc237046964bc9316f45698af9a0eced121/src/gnssro/gnssro_bufr2ioda.f90#L311

the obs2ioda may benefit from deflation and shuffle options

lateral move in IODA files information but smaller sizes!

rajichidamb:
## Description

jedi-ci-test-select=gcc

Adding Shuffle and Deflate option to compress files just like in GNSS_RO.
Andy Stokely checked in code in https://github.com/NCAR/obs2ioda/ 
I am merging those code in Jedi and testing the code .

Also adding coding norms test for the C++ files only, so most of the coding norm changes are in C++ files. Fortran code has the shuffle and deflate code changes.

## Issue(s) addressed

Resolves https://github.com/JCSDA-internal/ioda-converters/issues/1720

## Dependencies

- [ ] ...

## Impact

None

If you would like your reviewers to manually build and test the change, please include
instructions on how the change should be built and tested. Also include a short
justification on why manual testing is necessary for this change.

## Checklist

### DIFF
--- a/src/ncar/obs2ioda/src/cxx/netcdf_variable.cc
+++ b/src/ncar/obs2ioda/src/cxx/netcdf_variable.cc
@@ -1,3 +1,10 @@
+/*
+ * (C) Copyright 2026 UCAR
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
 #include "netcdf_variable.h"
 #include "netcdf_file.h"
 #include "netcdf_error.h"
@@ -36,8 +43,8 @@ namespace Obs2Ioda {
             std::vector<netCDF::NcDim> dims;
             dims.reserve(numDims);
             for (int i = 0; i < numDims; i++) {
-                dims.push_back(file->getDim(iodaSchema.getDimension(
-                        dimNames[i])->getValidName()));;
+                dims.push_back(file->getDim(
+                    iodaSchema.getDimension(dimNames[i])->getValidName()));;
             }
             auto iodaVarName = iodaSchema.getVariable(varName)->getValidName();
             auto var = group->addVar(iodaVarName,
@@ -50,7 +57,7 @@ namespace Obs2Ioda {
                 std::vector<size_t> chunks;
                 chunks.reserve(dims.size());
 
-                for (const auto &d: dims) {
+                for (const auto &d : dims) {
                     size_t n = d.getSize();
 
                     // avoid huge chunks (important for performance)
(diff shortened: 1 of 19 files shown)


src/ncar/obs2ioda/src/cxx/netcdf_variable.cc
```diff
@@ -29,34 +34,47 @@ namespace Obs2Ioda {
...
+                std::vector<size_t> chunks;
+                chunks.reserve(dims.size());
+
+                for (const auto &d : dims) {
+                    size_t n = d.getSize();
+
+                    // avoid huge chunks (important for performance)
+                    if (n > 256) n = 256;
+                    if (n == 0) n = 1;
+                    chunks.push_back(n);
```
srherbener, line 66: This chunking scheme isn't great for runtime performance as well as compression efficiency. The issue is that the chunk sizes are small and they don't necessary fit well with the access patterns in the data. Plus shuffle doesn't really make sense with floating point values (it tends to be much more effective with integer values).

However, the compression algorithm being used is ZLIB (deflate) which is known to have slow runtime. An effort is in place to investigate newer compressions schemes (LZ4, BLOSC) and we will likely move away from ZLIB.

I think in this context, it would be okay to go with this code for now since it isn't disastrous, and we will likely replace it later on.

What do others think?

srherbener (commented):
Thanks for doing this work - especially with the coding norms! That's a nice addition.

src/ncar/obs2ioda/src/cxx/netcdf_variable.cc
BenjaminRuston: @srherbener thanks for the review, there's no large need for rapid performance of this component

it generally runs offline and even in a near real time run this will not be the slowest component in the ingest, so going for maximum disk space saving over the fastest run time

hope that addresses the question

srherbener: Thanks @BenjaminRuston, your feedback is helpful! I'm still thinking it makes sense to go with what we have here, and wait for the outcome of the investigation into modern compression algorithms before considering any optimization here.

srherbener (approved):

hailingz (approved):
Looks good to me. Thank you for adding the deflate and shuffle!

BenjaminRuston (approved):
@rajichidamb this has been working well for me as well,,, note @srherbener comments and in a perfect world we don't have to use this much longer but it just keeps on going

BenjaminRuston:
this is so strange @rajichidamb I will dig into this some more, but these should be unaffected

these tests look very old, @CoryMartin-NOAA and/or @ADCollard should we remove these tests from iodaconverters?

think you've abstracted much and some of these early unit tests are not meaningful anymore, let me look into `fv3jedi_test_tier1_3dvar_gfs_sondes` again `ioda-converter` test to see what it's using.  This may be a good opportunity to streamline the repository a bit


image

rajichidamb:
There is a PR https://github.com/JCSDA-internal/fv3-jedi/pull/1470 to comment out this test. 2 weeks ago this test was added to CI https://github.com/JCSDA-internal/fv3-jedi/pull/1461, since this is failing only in Intel compiler run CI this test is again planned to be commented out.

PR #1756, opened 2026-03-17
Correct tangent point geolocation for space-based GNSS TEC observations

### LINKED ISSUE #1757: Correct tangent point geolocation for space-based GNSS TEC observations
## Current behavior 
The original converter uses LEO satellite locations as the observed TEC locations, which is incorrect. The observed TEC is slant path TEC, which corresponds to the signal path between the GNSS transmitter and the LEO receiver, not the LEO position itself.

For each observation, the converter should computes the closest point along the straight-line ray between the GNSS satellite and the LEO satellite. The closet point is defined tangent point for that sTEC measurement. Please note the straight line ray is an approximation to the real bended ray path. It is used only for ionosphere since the drift of the straight line from the bended ray path is much smaller compared to the scale of the ionosphere height range.

The closest point to the Earth's center along the line-of-sight is given by
**r**t=**r**LEO-(**r**LEO ⋅ **d**)**d**
where,
**𝑟**𝐿𝐸𝑂:  be the position vector of the LEO satellite
**d**: be the unit vector from the LEO satellite toward the GNSS transmitter

This point represents the minimum-distance point between the GNSS–LEO ray and the Earth's center.

The geodetic latitude, longitude, and altitude are then derived from this position vector and stored as the observation geolocation.

Notes:
For elevation < 0, this closest point corresponds to the tangent point of the ray path, which is physically meaningful for occultation geometry. For this reason, please use elevation < 0 as a criteria to uses height as a vertical coordinate for sTEC.
For elevation ≥ 0, the point represents the geometric closest approach of the ray to Earth, which is still a valid geometric reference for the slant TEC observation. But it is not the tangent point defined in occultation geometry since the GNSS is higher than LEO and therefore no occultation occurs. This lowest point is no longer between GNSS and LEO location. 

## Expected behavior
This change:
Corrects the geolocation associated with space-based GNSS TEC observations.
Ensures the observation location reflects the ray path geometry rather than the receiver position.
Improves the physical consistency of these observations when used in downstream analysis or data assimilation.

huishao-r:
## Description

This PR corrects the geolocation used in the GNSS space-based TEC converter.

Previously, the converter incorrectly used the LEO satellite latitude, longitude, and height as the location of the tangent point associated with the slant TEC measurement. It was done only for placeholder purposes, since the original data files from UCAR/CDAAC (podTEC) don't contain the tagent point locations of the TEC observation. Instead, the locations of the GNSS transmitter and the LEO receiver are provided.

The updated implementation computes the closest approach point of the GNSS–LEO line-of-sight to the Earth, which provides the correct geometric location associated with the TEC measurement.

## Issue(s) addressed

Resolves #1757 

## Impact

Add tangent point locations, which are computed based on provided geometry information

## Checklist


src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
```diff
@@ -254,20 +279,38 @@ def get_GNSS_mission(ds):
...
     transformer = pyproj.Transformer.from_crs({"proj": 'geocent', "ellps": 'WGS84', "datum": 'WGS84'},
                                               {"proj": 'latlong', "ellps": 'WGS84', "datum": 'WGS84'})
     # handling of km to meters should be automated
-    for i, x in enumerate(obs_data[("xECEFPosition", "MetaData")]):
-        lon, lat, height = transformer.transform(1000.*x,
-                                                 1000.*obs_data[("yECEFPosition", "MetaData")][i],
-                                                 1000.*obs_data[("zECEFPosition", "MetaData")][i],
-                                                 radians=False)
+    for i in range(nxleo):
+        # For elevation angle < 0, the point cloest to earth is in between GNSS and LEO and can be
```
jhaiduce:
```suggestion
        # For elevation angle < 0, the point closest to earth is in between GNSS and LEO and can be
```

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
```diff
@@ -254,20 +279,38 @@ def get_GNSS_mission(ds):
...
+        # considered as tangent point since straight line is a sufficient approximation in ionosphere.
+        # For elevation angle > 0, the point cloest to earty is below LEO height. So this is
```
jhaiduce:
```suggestion
        # For elevation angle > 0, the point closest to earth is below LEO height. So this is
```

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
```diff
@@ -254,20 +279,38 @@ def get_GNSS_mission(ds):
...
+        # not a physical tangent point definition. We'll keep this for now.
```
jhaiduce: Can you clarify what you mean by "not a physical tangent point definition?"

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
```diff
@@ -254,20 +279,38 @@ def get_GNSS_mission(ds):
 
 
 def get_geolocation(obs_data):
-    # wrapper to compute a reasonably accurate latitude and longitude
+    #  wrapper to compute latitude, longitude and height
```
jhaiduce:
```suggestion
    # wrapper to compute latitude, longitude and height
```

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
```diff
@@ -113,6 +113,30 @@ def main(args):
     writer.BuildIoda(obs_data, VarDims, VarAttrs, GlobalAttrs)
 
 
+def compute_tangent_point(xleo, yleo, zleo, xgps, ygps, zgps):
```
jhaiduce, line 116: Curious whether you considered writing this function to take two numpy arrays as opposed to six scalars. Operating on arrays would shorten the function by about six lines and it might be more performant as well.

jhaiduce (commented):
Overall this looks good. Just a few minor edits and suggestions.

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
huishao-r: As mentioned in our meeting, the tangent point is only validly defined when there is an occultation. But when there is no occultation, GNSS elevation > 0, the lowest point to the earth which is computed from the added function might be lower than LEO. It doesn't represent the ray path between GNSS and LEO. Therefore, though the computation is still valid, the altitude itself might not be as useful. We'll need to find what is the appropriate altitude definition for such a scenario. It will be included in a separate issue and PR later.

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
huishao-r: True. I'll come back later to implement that. For now, it works ok. But I can foresee the implementation with arrays would help with a large data set.

BenjaminRuston:
was able to run this for `podTc2_C2E3.2020.275.08.18.0013.G10.01_0001.0001_nc`

produced output with these ranges:
```
range(/MetaData/dateTime): [2020-10-01 08:18:31, 2020-10-01 08:31:02]
range(/MetaData/elevationAngleGNSS): [21.347209930419922, 47.77872848510742]
range(/MetaData/height): [-1731259.875, 60520.34765625]
range(/MetaData/latitude): [-24.946813583374023, -8.633904457092285]
range(/MetaData/longitude): [86.27073669433594, 137.4101104736328]
range(/MetaData/satelliteConstellationRO): [401, 401]
range(/MetaData/satelliteIdentifier): [752, 752]
range(/MetaData/satelliteSubIdentifier): [3, 3]
range(/MetaData/satelliteTransmitterId): [10, 10]
range(/MetaData/sequenceNumber): [1, 1]
range(/MetaData/xECEFPosition): [-4105.20947265625, 837.8147583007812]
range(/MetaData/xECEFPositionGNSS): [1279.3040771484375, 3160.76904296875]
range(/MetaData/yECEFPosition): [5426.45849609375, 6438.6181640625]
range(/MetaData/yECEFPositionGNSS): [14975.291015625, 15777.4443359375]
range(/MetaData/zECEFPosition): [1217.577392578125, 2687.346435546875]
range(/MetaData/zECEFPositionGNSS): [21377.58984375, 21775.896484375]
range(/ObsValue/totalElectronContent): [6.47951602935791, 11.782017707824707]
```

BenjaminRuston (approved):
this is working and the new tangent point is being calculated correctly thanks

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
jhaiduce: Makes sense to leave it as-is unless we see a need for improved performance.

jhaiduce (approved):

PR #1763, opened 2026-03-26
update mtg irs converter to include a variety of more end users use cases.

### LINKED ISSUE #1762: updates needed to mtg irs converter
## Description
After using some data in an hofx test for MTG-IRS I noticed a few changed and missing variables. One where I'm not sure what the correct answer would be is sensoScanPosition. Right now, I just filled that in with dwell number. 
@hongli-wang filled this with a truncated integer value of sensorZenithAngle as this roughly matched up with similar metadata associated with the GOES ABI implementation. 
## Requirements
Can run an hofx test using CRTM.
End users are happy with available metadata

Can run an hofx test using CRTM
End users are happy with available metadata

## Dependencies
None

karpob:
## Description

Resolves #1762. In addition to fixing variables that are needed for CRTM.  I added the option to apodize and output radiance and Tb. I added radiance (in addition to Tb) as @hongli-wang uses it in his implementation.  I also added some metadata that NOAA-EMC included in their implementation.  To reiterate the question in the issue: what should be placed in sensorScanPosition? 

## Issue(s) addressed

Resolves #1762 

## Dependencies
None

## Impact
Makes it possible to run and hofx test using mtg-irs along with satisfying more end users metadata use cases.

## Checklist

### DIFF
--- a/src/satpy/irs_2ioda.py
+++ b/src/satpy/irs_2ioda.py
@@ -15,6 +15,7 @@ import argparse
 from datetime import datetime, timezone, timedelta
 import os.path
 import h5py
+import netcdf4 as nc
 import numpy as np
 import pyiodaconv.ioda_conv_engines as iconv
 from pyiodaconv.orddicts import DefaultOrderedDict
--- a/src/satpy/irs_2ioda.py
+++ b/src/satpy/irs_2ioda.py
@@ -74,26 +74,28 @@ def main(args):
     output_file = task_params.pop('output')
     dtg = task_params.pop('date')
     threads = task_params.pop('threads')
-    func_with_args = partial(get_data_from_files, **task_params)
-    with ProcessPoolExecutor(max_workers=threads) as executor:
-        for file_obs_data in executor.map(func_with_args, input_files):
-            if not file_obs_data:
-                print("INFO: non-nominal file skipping")
-                continue
+    if (threads > 1):
+        func_with_args = partial(get_data_from_files, **task_params)
+        with ProcessPoolExecutor(max_workers=threads) as executor:
+            for file_obs_data in executor.map(func_with_args, input_files):
+                if not file_obs_data:
+                    print("INFO: non-nominal file skipping")
+                    continue
+                if obs_data:
+                    concat_obs_dict(obs_data, file_obs_data[1])
+                else:
+                    obs_data = file_obs_data[1]
+                    wavenumber = file_obs_data[0]
+
+    # serial option
+    else:
+        for afile in input_files:
+            file_obs_data = get_data_from_files(afile, **task_params)
             if obs_data:
                 concat_obs_dict(obs_data, file_obs_data[1])
             else:
                 obs_data = file_obs_data[1]
                 wavenumber = file_obs_data[0]
-
-    # serial option
-#   for afile in input_files:
-#       file_obs_data = get_data_from_files(afile, **task_params)
-#       if obs_data:
-#           concat_obs_dict(obs_data, file_obs_data[1])
-#       else:
-#           obs_data = file_obs_data[1]
-#           wavenumber = file_obs_data[0]
 
     obs_data[('sensorCentralWavenumber', metaDataName)] = np.array(wavenumber, dtype='float32')
     obs_data[('sensorChannelNumber', metaDataName)] = np.arange(1, len(wavenumber)+1, dtype='int32')
@@ -114,6 +116,7 @@ def main(args):
     # pass parameters to the IODA writer
     VarDims = {
         'brightnessTemperature': ['Location', 'Channel'],
+        'radiance': ['Location', 'Channel'],
         'sensorChannelNumber': ['Channel'],
         'sensorCentralWavenumber': ['Channel'],
     }
@@ -134,23 +137,49 @@ def main(args):
     VarAttrs[(k, 'ObsValue')]['units'] = 'K'
     VarAttrs[(k, 'ObsError')]['units'] = 'K'
 
+    k = 'radiance'
+    VarAttrs[(k, 'ObsValue')]['_FillValue'] = float_missing_value
+    VarAttrs[(k, 'ObsError')]['_FillValue'] = float_missing_value
+    VarAttrs[(k, 'PreQC')]['_FillValue'] = int_missing_value
+    VarAttrs[(k, 'ObsValue')]['units'] = 'W m-2 sr-1 m'
+    VarAttrs[(k, 'ObsError')]['units'] = 'W m-2 sr-1 m'
+
     obs_data[('longitude', metaDataName)] = obs_data[('longitude', metaDataName)] % 360
     # final write to IODA file
     writer.BuildIoda(obs_data, VarDims, VarAttrs, GlobalAttrs)
 
 
-def readMatrix(f, band):
+def apodizeHammingMatmul(spectrum):
+    A = np.zeros([max(spectrum.shape), max(spectrum.shape)])
+    np.fill_diagonal(A, 0.54)
+    msk_upper = np.eye(A.shape[0], k=1, dtype=bool)
+    msk_lower = np.eye(A.shape[0], k=-1, dtype=bool)
+    A[msk_upper] = 0.23
+    A[msk_lower] = 0.23
+    # Apply Boundary condition for 2 point vs 3 point window on edges
+    A[0, 0] = A[0, 0]/0.77
+    A[0, 1] = A[0, 1]/0.77
+    A[-1, -2] = A[-1, -2]/0.77
+    A[-1, -1] = A[-1, -1]/0.77
+    spectrum_out = A@spectrum.T
+    return spectrum_out.T
+
+
+def readMatrix(f, band, apodize=False):
     h5 = h5py.File(f, 'r')
     ReconstructionOperator = np.asarray(h5[band + '/reconstruction_operator'])
     Mean = np.asarray(h5[band + '/mean_spectrum'])
     h5.close()
+    if (apodize):
+        ReconstructionOperator = apodizeHammingMatmul(ReconstructionOperator)
+        Mean = apodizeHammingMatmul(Mean)
     # flip reconstruction operator to be consistent with what is done for
     # local reconstruction operator
     return ReconstructionOperator.T, Mean
 
 
-def applyPc(f, scores, operator_local, scores_local, band):
-    R, meanz = readMatrix(f, band)
+def applyPc(f, scores, operator_local, scores_local, band, apodize):
+    R, meanz = readMatrix(f, band, apodize)
     Rl = operator_local
     # reshape like this to preserve dwells and columns
     s = np.asarray(scores).transpose([2, 0, 1])
@@ -190,7 +219,15 @@ def r2tb(Radiance, nu, c1=1.191042972e-16, c2=1.4387769e-2):
     return Temperature
 
 
-def get_data_from_files(afile, resolution=160, scan_shape=(160, 160), include_reconstructed_radiance=False, baseEV=None):
+def get_data_from_files(
+    afile,
+    resolution=160,
+    scan_shape=(160, 160),
+    include_reconstructed_radiance=False,
+    include_reconstructed_tb=False,
+    apodize_reconstructed=False,
+    baseEV=None
+):
     print('processing file', afile)
     f = nc.Dataset(afile)
     obs_data = {}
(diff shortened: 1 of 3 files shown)


BenjaminRuston:
@karpob regarding this:
> To reiterate the question in the issue: what should be placed in sensorScanPosition?

I think what I've done in the past for Geostationary data, was set the `sensorScanPosition` to the x-index from the particular array of the scan segment

as someone has likely noticed it needs to be defined though it actually is never used downstream in the UFO interface call to CRTM (or RTTOV)

BenjaminRuston:
@karpob here is what was done for SEVIRI data:
https://github.com/JCSDA-internal/ioda-converters/blob/54ab4bd3eb43608d5a7ed3729273d970f8216323/src/satpy/seviri_l1b_to_ioda.py#L622

HaidaoLin-NOAA:
> @karpob here is what was done for SEVIRI data:
> 
> https://github.com/JCSDA-internal/ioda-converters/blob/54ab4bd3eb43608d5a7ed3729273d970f8216323/src/satpy/seviri_l1b_to_ioda.py#L622

@BenjaminRuston  Thanks!  It looks like this is how gsi handles seviri/abi data for sensor scan position:
iscan = nint(hdr(ilzah))+1.001_r_kind 
ilzah is the satellite zenith angle

BenjaminRuston:
thanks @HaidaoLin-NOAA yeah, it needs to have an integer value of some kind currently for some sanity checks

@karpob mentioned that the rounded zenith was used in other code,, think the use of `dwell_number` is a relevant one and may be the best.  The `zenith angle` is already stored so adding the dwell is new information

to be honest happy with whatever is decided, for my vote would stick with Bryan's choice of the `dwellNumber`

hongli-wang:
Will sensorScanPosition be used in VarBC? If yes, the way like GSI might be more effective to remove bias. @karpob @BenjaminRuston @HaidaoLin-NOAA

karpob:
I'm not sure the GSI approach would make the most sense here for a bias predictor. I would think you could use sensorZenithAngle directly? 

Alternatively, if having a discrete quantity is desired, I would think the best approach would be a to have a quantity which uniquely defines the ifov. I would think you could do this with a combination of Local Area Coverage number (1-4), dwell number (1-72 in the test data, on an aside this looks like it may end up being 1-73), and a number to identify the ifov within the 160x160 dwell.

hongli-wang:
Regarding the bias correction, I think we can specify the predictors (for example, sensorZenithAngle) in JEDI.

karpob:
The last push requires more explanation which I'll try to document today. It may make sense to split this up into a few PRs, or since this is a draft just put it all in. Basically I added the following:
1. A new value for sensorScanPosition which gives a unique number based on the pixels in the dwell, number of dwells, and local area coverage.
2. The ability to set a the next level lower thinning so higher zenith angles can be thinned more similarly to center FOV based on the middle of the dwell zenith angle.
3. Add the ability to set a cutoff zenith angle such that data will not be passed on past a selected zenith angle.
4. Mimic the behavior currently in NOAA's spoc converter which gives preference to the largest 1st pc score indicating warmest fov, or optionally the cloudFraction.

karpob:
So for an example of 1 on my list. This is just the new value given which will be unique value ranging from 0-7,475,200 for a local area coverage containing 73 dwells (what it currently looks like), or 0-7,372,800 for a coverage area conaining 72 dwells (what is in the publicly  available data). Here's a plot of what that looks like:

image

For an example of 2 on my list. This is an option which allows you to select two thinning levels and have this switch at a specified zenith angle for the center dwell pixel.  You will notice that in the previous plot, sampling is rather sparse over Europe vs over Africa (thinning resolution = 160). If one runs the following command:
```
./irs_2ioda.py -i tmp/*.nc -j10 -o $PWD/${DTG}_irs_dump.nc4 --resolution 160 --step_resolution --resolution_step 55
```
You will get the following:

image

You will notice a denser sampling pattern for zenith angles above 55 degrees. The resolution is specified in the "resolution_pairs" dictionary at the start of the script. For a given resolution of 160, the code will use a resolution of 160  from 0-55 degrees zenith angle, then from 55-90 use a resolution of 128. 

An example of feature 3 on my list, say you would rather not include zenith angles greater than 80, because the radiative transfer model is incapable of performing radiative transfer calculations beyond 80 degrees. The user can then use the following command:

```
./irs_2ioda.py -i tmp2/*.nc -j10 -o $PWD/${DTG}_irs_dump.nc4 --resolution 160 --max_zenith 80
```
You will get the following removing any pixel above 80 degrees zenith angle:

image

Next, for an example of 4 on my list, I have plotted two curves. The blue curve is the zonal mean from observations using the nominal 160 resolution with an 80 degree cutoff (command above), the red curve is using the following command:
```
./irs_2ioda.py -i tmp2/*.nc -j10 -o $PWD/${DTG}_irs_dump.nc4 --resolution 160 --max_zenith 80 --get_largest_pc --window 3
```

image

As designed, a 3x3 selection is made based on the largest value of PC score 1. The option --get_cloud_fraction will do the same process, but for the clearest pixel in-lieu of largest PC score 1. 

Finally, just for fun plotting the previous map on a Mollweide projection provides a festive colored egg pattern:
image

jack-woollen:
@karpob Is there an example of ioda output from your recent version updates so we can pick up the template? Also I'd be curious to see what execution times this code produces for various thinning levels. Thanks.

karpob:
@jack-woollen I can run this over some of the publicly available data. In the mean time, here's the output from an ncdump of a file I've been looking at. 
[template.txt](https://github.com/user-attachments/files/26637540/template.txt)

jack-woollen:
@karpob Your code was helpful to understand the IODA file contents required for IRS data. Thanks! We have converted the TD528 dataset and were wondering what JEDI version is needed to test the data in.

karpob:
> @karpob Your code was helpful to understand the IODA file contents required for IRS data. Thanks! We have converted the TD528 dataset and were wondering what JEDI version is needed to test the data in.

@jack-woollen I think this would depend upon what you're looking to test. If you're looking to test the pc score to radiance transform, you would need at least a version which includes this pull request merged:
https://github.com/JCSDA-internal/ufo/pull/3895

BenjaminRuston:
@karpob should we take this out of draft, think it's ready to go no?

karpob:
Yeah, I think so. If there's anything else, I can put in another PR. Quite a bit here anyway.

src/satpy/irs_2ioda.py
```diff
@@ -11,12 +11,9 @@
...
 Python code to ingest netCDF4 IRS data
 """
-from matplotlib import pyplot as plt
 import argparse
 from datetime import datetime, timezone, timedelta
 import os.path
-import sys
-import glob
 import h5py
 import netCDF4 as nc
```
BenjaminRuston, line 18:
```suggestion
```

BenjaminRuston:
@karpob here's the error:

```
  File "/workdir/build/bin/irs_2ioda.py", line 692, in <module>
    main(args)
  File "/workdir/build/bin/irs_2ioda.py", line 105, in main
    file_obs_data = get_data_from_files(afile, **task_params)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workdir/build/bin/irs_2ioda.py", line 251, in get_data_from_files
    f = nc.Dataset(afile)
        ^^
NameError: name 'nc' is not defined. Did you mean: 'np'?
```

karpob:
@BenjaminRuston, You do need the line back that you removed after applying the suggestion:
```python
import netCDF4 as nc
```

jack-woollen (commented):
@karpob Looking at
    # --- sensorScanPosition: unique identifier across all coverages and pixels ---
    # Encoding: coverage_idx * (N_DWELLS * n_rows * n_cols)
    #         + dwell_number  * (n_rows * n_cols)
    #         + dwell_row     * n_cols
    #         + dwell_col
    #
    # coverage attribute is one of Q1, Q2, Q3, Q4 (N dwells each default 73).
    # dwell_number is converted 0-based within the coverage (0–N).
    #    dwell_row and dwell_col are the 2D pixel indices within the dwell (scan_shape).

From the data it looks like the dwell number is absolute (0-302), not relative to the LAC it's in. So maybe no need for coverage_idx in the position calculation.

karpob:
@BenjaminRuston seems it's currently failing because the test data doesn't have zenith angle in it. If I recall correctly you did a fair amount of massaging of the test data to minimize file size. I can take a stab at making a new test dataset, if you don't have the original data handy.

BenjaminRuston:
@karpob oh,, I should still have the incantation somewhere that reduced the file size

it's a matter of copying **only** the items you need into a new file and then reduce that one so likely missed that metaData field

BenjaminRuston:
starting with this:

```
nccopy -V latitude,longitude,dwell_type,dwell_number,satellite_azimuth_angle,satellite_zenith_angle,solar_azimuth_angle,solar_zenith_angle,time,wavenumber,instrument_quality_warning,radiometric_quality_warning,global_pc_scores \
-G /data,/data/lwir,/data/mwir,/data/lwir/compressed,/data/lwir/quality_band,/data/mwir/compressed,/data/mwir/quality_band \
${afile} ${output_file}

nccopy -d 9 -u -s ${output_file} ${output_file}.temp
mv ${output_file}.temp ${output_file}
```

BenjaminRuston:
ncdump of the header shows satellite_zenith_angle...
`ncdump -h ../../test/testinput/EUMT_IDPFS_DEV_20210624080230_20210624080240_N__T_0001_0015.nc`:

```
...
  	short satellite_zenith_angle(dwell_row, dwell_column) ;
  		satellite_zenith_angle:_FillValue = -32767s ;
  		satellite_zenith_angle:long_name = "Satellite zenith angle" ;
  		satellite_zenith_angle:units = "degrees" ;
  		satellite_zenith_angle:add_offset = 38.1218195645793 ;
  		satellite_zenith_angle:scale_factor = 0.0505088774702499 ;
```

digging into it

BenjaminRuston:
think I've got it because of a `ncks` failure and couldn't find my workaround last time (think at least did the striding with python)

but had to add `state/platform/platform_altitude` in the variable and group copy and then the default `160 x 160` scene is sub-sampled with a stride of 20 for a resulting `8 x 8` which is used in the unit test



Did the whole thing in python, do not think this is worth adding to the repository but will attach here to this PR:
`TD528 input file: EUMT_IDPFS_DEV_20210624080230_20210624080240_N__T_0001_0015.nc`

[make_irs_ctest_file.py](https://github.com/user-attachments/files/29312734/make_irs_ctest_file.py)

BenjaminRuston (approved):
thanks @karpob these are helpful additions

BenjaminRuston:
@jack-woollen does this seem ready from your point of view

@karpob think this is now ready for merge any hesitations let me know

BenjaminRuston:
@HaidaoLin-NOAA or @keenaneure have you tried this?

the MTG-IRS data is being released in late July 2026

keenaneure:
I have not personally. It is possible @hongli-wang has tried this.

rajichidamb (approved):

karpob:
> @jack-woollen does this seem ready from your point of view
> 
> @karpob think this is now ready for merge any hesitations let me know

Sorry, just got back from break. Looks OK from my view. I'll take Jack's suggestion under advisement, and make note of it. I'm sure there will be minor updates along the way at some point to add the suggestion as well.

jack-woollen (approved):
I think the formula for sensor scan position is not right. The dwell number I see in IRS data is absolute (IE 1-305), not relative within a LAC. So if LAC number is represented in the scan position then max_dwells should be 305 or dwell number should be adjusted to be a relative position in the LAC, no bigger than max_dwells. I haven't seen any of the restricted data so maybe the dwell number there is LAC relative position?

karpob:
@jack-woollen I've been taking a look at dwell_number a little bit, and it seems things could get complicated. Reading https://user.eumetsat.int/resources/user-guides/mtg-irs-level-1-data-guide, it looks like it flips at each equinox, so these numbers would be seasonally dependent. In the simulated data, I'm seeing dwell_numbers that go to 308, and something slightly smaller in the pre-release. 

Any thoughts on what formula makes sense for scan position?

jack-woollen:
@karpob The dwell number is a unique reference to generate a scan position from. The fact that the reported dwell number layout changes periodically might require some mapping to an absolute dwell numbering which would always refer to the same geographical location, if that's important. TBD.

BenjaminRuston:
@jack-woollen and @karpob moving this forward for now and we will update as needed

PR #1765, opened 2026-03-31
 dynamically determine if Refractivity is present in wmo bufr

### LINKED ISSUE #1764: option change for GNSS-RO converter for WMO bufr when no refractivity
## Description

the `gnssro_bufr2ioda.py` has an option `--onlybendingangle` that is not really needed

right now it is required if you have an input file that does **not** contain refractivity profiles for the converter to run and not crash.  In this case the `ObsValue/Refractivity` is still populated with missing value regardless.

This option can be removed and the converter create the `ObsValue/Refractivity` populated with missing value on-the-fly as needed

## Requirements

test with a file with bending angle only and no refractivity profile, this is common for some real time EUMETSAT WMO BUFR data from the GRAS sensors

converter continues to function for files with and without refractivity profiles

BenjaminRuston:
## Description

remove the `--onlybendingangle` option and dynamically determine if Refractivity is present in the file

## Issue(s) addressed

Resolves #1764 

## Impact

user does **not** need to specify the `--onlybendingangle` option if the input file is missing the `Refractivity` profiles.  This continues with the practice to populate an `ObsValue/Refractivity` with missing values

## Checklist

### DIFF
--- a/src/gnssro/gnssro_bufr2ioda.py
+++ b/src/gnssro/gnssro_bufr2ioda.py
@@ -64,7 +64,6 @@ def main(args):
     dtg = datetime.strptime(args.date, '%Y%m%d%H')
     qc = args.qualitycontrol
     addLSW = args.localspectralwidth
-    only_bang = args.onlybendingangle
     use_average_tangent_point = args.use_average_tangent_point
 
     # read / process files in parallel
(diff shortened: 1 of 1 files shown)


BenjaminRuston:
@fcvdb and @haydenlj found this when processing the `GRAS-2` files from EUMETSAT and was required to provide the `--onlybendingangle` switch or option

this now dynamically will determine if `Refractivity` profiles are present and fill them with missing values, the same behavior as when the `--onlybendingangle` option was present

haydenlj (approved):
Thanks @BenjaminRuston

src/gnssro/gnssro_bufr2ioda.py
hailingz, line 301: Talking about the bending angle to refractivity processing chain (such as from EUMETSAT to ROMSAF), the summary bit of quality flag could be non-nominal when bending angle is good but refractivity is bad/missing/not processed yet. For this reason, some centers ignore the summary bit when assimilating bending angle in operation. Users who assimilate and care about refractivity can simply filter it in yaml by using the qualityFlag parameter. This is exactly the same reason @BenR asked me to remove such quality controls in the Fortran converter a couple years ago.  I prefer to remove this summary bit (i_non_nominal) here.

hailingz (approved):

PR #1774, opened 2026-04-27
Add prescribed ob error option to COSMIC2 to ioda converter

### LINKED ISSUE #1773: Observation error missing from gnss_tec_spacebased_cdaac2ioda converter
This issue is concerning the converter:

src/space_weather/gnss_tec_spacebased_cdaac2ioda.py 

The input observation files for this converter do not have observation error variables, however the converted file  needs to have an observation error.  Should add a prescribed observation error to the converted output file.

michael-a-herrera:
run-ci-on-draft=true

## Description

For the COSMIC2 to ioda converter, the output observation files were missing observation error, so this fix adds the option to prescribe an observation error.  The input observation files do not have observation errors so the variable "ObsError" is missing from the final converted netcdf file, which is required for assimilation.  

Added an optional argument --obserror or -e which expects a float with a single observation error that is applied to all observations in the file.  Default value is set to 3.0 TECU.   

## Issue(s) addressed

Resolves #1773

## Checklist

### DIFF
--- a/src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
+++ b/src/space_weather/gnss_tec_spacebased_cdaac2ioda.py
@@ -210,7 +210,7 @@ def get_obs_data(ifile, get_obs_data_args):
     obs_data = get_geolocation(obs_data)
     # the observation value
     obs_data[("totalElectronContent", "ObsValue")] = np.array(ds['TEC'][:], dtype=ioda_float_type)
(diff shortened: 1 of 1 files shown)


jhaiduce (approved):
Thanks @michael-a-herrera !

huishao-r (approved):
Thanks, @michael-a-herrera. This fix is necessary for current runs. Thanks for doing this. @BenjaminRuston, it seems necessary to have a ctest for this now. We talked about this. I'll work on getting a ctest set up and a data file.

haydenlj (approved):
Thanks @michael-a-herrera !

Issue #1775, opened 2026-04-28
Question on compute_scan_angle


chengdang:
@BenjaminRuston @BenjaminTJohnson 

Looking at the compute_scan_angle https://github.com/JCSDA-internal/ioda-converters/blob/ddba91077c286891906ed75e9cd616ba5bb2d80c/src/pyiodaconv/def_jedi_utils.py#L77

I have a question: with 
https://github.com/JCSDA-internal/ioda-converters/blob/ddba91077c286891906ed75e9cd616ba5bb2d80c/src/pyiodaconv/def_jedi_utils.py#L90

When `sum(good) <= 0`, or for non-good indices,`ratio` never gets assigned, which would bring a random value to `scanang` calculation.

Is this by design since it does not matter if QC is bad?

chengdang:
Another question is with condition:

https://github.com/JCSDA-internal/ioda-converters/blob/ddba91077c286891906ed75e9cd616ba5bb2d80c/src/pyiodaconv/def_jedi_utils.py#L94

If user do pass a valid qc_flag array to this function, for example, [0,1,1,1,1]. I assume it means it's bad QC except for the 1st element. But, with the current setup, qc_flag will be re-assigned as all good QC [0,0,0,0,0]. Is this also by design?

Still new to this QC flag so I'm wondering why it's set up like this.

BenjaminRuston:
thanks @chengdang yes there is likely something to fix here

the desired behavior was for qcFlag to be optional and **if** passed it would only compute for Locations with qcFlag == 0; and the others should be populated with the float missing value

PR #1780, opened 2026-04-30
Fix function compute_scan_angle (QC flags), remove unused argument

chengdang:
## Description

This PR fixes `compute_scan_angle` by:

- removing the unused argument
- fixing QC-flag handling
- updating all instrument converter scripts that call `compute_scan_angle`

## Issue(s) addressed

Resolves issue https://github.com/JCSDA-internal/ioda-converters/issues/1775

## Dependencies
None

## Impact
None

## Checklist


chengdang:
Have not tested it, needs to rebuild ioda

BenjaminRuston:
> needs to rebuild ioda

@chengdang actually for these you don't technically need to rebuild, just copy the new def_jedi_utils.py into the $JEDI_BUILD directory,, the path is in I think lib/python/....  

thanks for getting this in and for giving it the attention it needed

chengdang:
@BenjaminRuston 
Thanks! That's good to know. I was wondering if I need to re-build. Let me run some tests then.

BenjaminRuston:
ok yes the **hack** is to obviously get on the branch and then you have to copy the `def_jedi_utils.py` file to:

```
${JEDI_BUILD}/lib/python3.11/pyiodaconv
```

so that's a shortcut to do a quick test

chengdang:
Thank you @BenjaminRuston 
I did a test with the aws file, here is the post-ioda view angle:

image

I think it is working as expected, but more tests with other sensors are probably needed.

BenjaminTJohnson (approved):

src/pyiodaconv/def_jedi_utils.py
```diff
@@ -87,18 +87,23 @@ def compute_scan_angle(instr_scan_ang, sensor_altitude, sensor_zenith, qc_flag=[
     d2r = np.pi/180.
     r2d = 180./np.pi
 
-    # do we need a missing here
-    ratio = np.empty_like(sensor_altitude)
+    # Initialize output with missing values
+    scanang = np.full_like(sensor_altitude, np.nan, dtype=float)
```
BenjaminRuston:
```suggestion
    scanang = np.full_like(sensor_altitude, float_missing_value, dtype=float)
```

BenjaminRuston: @chengdang just one small change, this utility also tried to standardize the `int`, `float` and `long` missing values put into the files... they are *globals* up at the top

src/pyiodaconv/def_jedi_utils.py
```diff
@@ -86,18 +86,23 @@ def compute_scan_angle(instr_scan_ang, sensor_altitude, sensor_zenith, qc_flag=[
...
-    if not qc_flag[0]:
-        qc_flag = np.zeros_like(sensor_altitude)
-    good = qc_flag[:] == 0
-    if sum(good) > 0:
-        ratio[good] = earth_mean_radius_km/(earth_mean_radius_km + sensor_altitude[good]/1000.)
-
-    # γ = arcsin(R / (R + h) * sin(theta)),h: sat alt; theta: sat zenith angle
-    scanang = np.arcsin(ratio*np.sin(abs(sensor_zenith)*d2r))*r2d
+    # Handle qc_flag
+    if qc_flag is None:
```
rajichidamb: Coding norm failures
./def_jedi_utils.py:93:24: W291 trailing whitespace
./def_jedi_utils.py:100:21: W291 trailing whitespace

rajichidamb:
```suggestion
    if qc_flag is None:
```

src/pyiodaconv/def_jedi_utils.py
```diff
@@ -86,18 +86,23 @@ def compute_scan_angle(instr_scan_ang, sensor_altitude, sensor_zenith, qc_flag=[
...
-    scanang = np.arcsin(ratio*np.sin(abs(sensor_zenith)*d2r))*r2d
+    # Handle qc_flag
+    if qc_flag is None: 
+        good = np.ones_like(sensor_altitude, dtype=bool)
+    else:
+        qc_flag = np.asarray(qc_flag)
+        good = (qc_flag == 0)
+
+    # compute scan angle only for good observations
+    if np.any(good):
```
rajichidamb:
```suggestion
    if np.any(good):
```

src/pyiodaconv/def_jedi_utils.py
```diff
@@ -86,18 +86,23 @@ def compute_scan_angle(instr_scan_ang, sensor_altitude, sensor_zenith, qc_flag=[
     d2r = np.pi/180.
     r2d = 180./np.pi
 
-    # do we need a missing here
-    ratio = np.empty_like(sensor_altitude)
+    # Initialize output with missing values
+    scanang = np.full_like(sensor_altitude, float_missing_value, dtype=float)
```
rajichidamb: Test are failing, so is the value supposed to be double missing value?
```
DIFFER : VARIABLE : sensorViewAngle : TYPE : DOUBLE <> FLOAT
DIFFER : TYPES : ATTRIBUTE : _FillValue : VARIABLE : sensorViewAngle : DOUBLE <> FLOAT
```

BenjaminRuston:
ok @rajichidamb picked up your other changes and yes got some failures, switching to double to see if the failures change as this may fix those and cause other errors now elsewhere

think we have a situation where some converters used float and other double for the missing value

src/pyiodaconv/def_jedi_utils.py
```diff
@@ -86,18 +87,23 @@ def compute_scan_angle(instr_scan_ang, sensor_altitude, sensor_zenith, qc_flag=[
     d2r = np.pi/180.
     r2d = 180./np.pi
 
-    # do we need a missing here
-    ratio = np.empty_like(sensor_altitude)
+    # Initialize output with missing values
+    scanang = np.full_like(sensor_altitude, double_missing_value, dtype=float)
```
rajichidamb: Setting dtype=float but using double_missing_value? Other python scripts fill the array with 'float32'
```
 sensor_altitude,
        np.array(f['GeolocationAndFlags']['earth_inc_ang'], dtype='float32'),
```

BenjaminTJohnson:
The `np.full_like(..., float_missing_value)` initialization resolves the uninitialized-memory problem in the old `np.empty_like` path, and the new QC handling fixes the old if not `qc_flag[0]` bug that silently discarded the QC array whenever the first ob was good. A few things before this merges:

1. Two call sites still use the old signature (silent breakage). `src/hdf5/aws_2ioda.py` (~L69) and `src/satpy/irs_2ioda.py` (~L310) still call with the old 3-positional form. Since the removed argument was first, Python won't error, the args just shift (zenith → altitude, altitude → zenith, zenith → qc_flag), so nearly all obs get flagged bad and `sensorViewAngle` comes out missing. These need updating in this PR (if there's not corresponding PRs?) Worth also double-checking `cloudsat_netcdf_to_ioda.py`, `ozone_product_2ioda.py`, `wsfm_2ioda.py,` and `seviri_l1b_to_ioda.py`, which import the function.

2. Altitude units are inconsistent across callers. The function divides `sensor_altitude` by 1000 (expects meters), but amsr2 (strip('km')), gmi (407.0), and tropics (550.) pass km (maybe others??):  for those, the computed scan angle collapses to ~ the zenith angle. Pre-existing, but since this PR fixes the function, could we document the expected unit in the docstring and audit callers (or handle it here)?

3. Minor:
- `dtype=float` makes the output float64; suggest `np.float32` for consistency with `ioda_float_type` (or just be rigorous about using the correct float type that's already defined)?
- double_missing_value is added but unused, was there a plan here?
- Reference/CI data may need regenerating for converters that pass QC flags (`cowvr` especially), since outputs legitimately change.

src/pyiodaconv/def_jedi_utils.py
```diff
@@ -86,18 +87,23 @@ def compute_scan_angle(instr_scan_ang, sensor_altitude, sensor_zenith, qc_flag=[
     d2r = np.pi/180.
     r2d = 180./np.pi
 
-    # do we need a missing here
-    ratio = np.empty_like(sensor_altitude)
+    # Initialize output with missing values
+    scanang = np.full_like(sensor_altitude, float_missing_value, dtype=float)
```
BenjaminRuston:
```suggestion
    scanang = np.full_like(sensor_altitude, float_missing_value, dtype=np.float32)
```

BenjaminRuston:
thanks for the re-review @BenjaminTJohnson , have made the changes and confirmed satellite altitudes for those calling this function at least are in meters

rajichidamb:
Reference files need update to match the new calculated `sensorViewAngle`

BenjaminRuston (approved):
thanks for this clean up @chengdang it definitely helps and caught some inconsistencies with definition of satellite altitude. In short it was very useful

PR #1784, opened 2026-05-15
Cleaning tempo converter

mer-a-o:
## Description

This PR updates the TEMPO converter and files in this repo with V04 of TEMPO data. A test for total column NO2 is added and the uncertainty is now better handled for both total and troposphere columns. Some additional information is added to the `global attributes` such as `processing_version` and `apriori_source`

## Issue(s) addressed

None

## Dependencies

None

## Impact
None


weiwilliam (commented):
Thank you, @mer-a-o 
Updated code runs fine for me.
Only a few minor things.

- `nitrogendioxideTotal` was removed from share/ioda/yaml/validation/ObsSpace.yaml.
There is `nitrogendioxideProfile` but with mol mol-1 as units.
Do we want to bring `nitrogendioxideTotal` back or update the variable name in the converter?
- Do we need to keep the print out message of before/after compression?

mer-a-o:
Thanks @weiwilliam. I didn't know that! I've been using `nitrogendioxideTotal` and `carbondioxideTotal` for some time now. It looks like it has been gone for a few years. If we decide to use `nitrogendioxideProfile` we need to fix the unit. @jeromebarre what do you think?

Yes, we want to keep the before/after compression messages.

jeromebarre:
> * share/ioda/yaml/validation/ObsSpace.yaml

A tropospheric column and total column are just columns. So I think it is a good thing that the suffix total was removed it was confusing anyways.

jeromebarre:
We should not use profile for the tempo data...

jeromebarre:
The converter produces two different files for tropospheric and total column anyways. in each file the obs variable name should have the suffix column for both. introducting a metadata variable should help differenciate.

src/compo/tempo_nc2ioda.py
rajichidamb:
```suggestion
                if self.varname == 'hcho':
```

rajichidamb, line 9: Can you please add a description of what the file does, would be helpful for people like me. And may be the usage format too. Thanks.

mer-a-o: I added a description to the code. Let me know if it addresses your comments.

mer-a-o:
@jeromebarre @weiwilliam, I added a `column_type` key to the global attributes and use `nitrogendioxideColumn` for both tropospheric and total now. 
There may be an impact on skylab/simobs.

rajichidamb (approved):
I approve from my level of understanding. Thanks @mer-a-o

jeromebarre (approved):

BenjaminRuston (approved):
thanks @mer-a-o for the input and your patience

PR #1791, opened 2026-06-11
Write out the station ID in adpsfc and sfcshp IODA files

### LINKED ISSUE #1790: station ID is missing in station ID in adpsfc and sfcshp IODA files
## Description 

The station ID is missing from in `adpsfc` and `sfcshp` IODA files. This prevents the use of certain obs filters.

fcvdb:
## Description

The station ID is missing from the adpsfc and sfcshp IODA files. This prevents the use of certain obs filters.

Some BUFR files contain station IDs with non-UTF-8 bytes (e.g. Latin-1 accented characters). [iodaconv/src/ncepbufr/prepbufr_adpsfc_api.py](https://github.com/JCSDA-internal/ioda-converters/blob/4edd5731a0f80f51b14af5abf5639c9eb53d88f7/src/ncepbufr/prepbufr_adpsfc_api.py#L75)  and [iodaconv/src/ncepbufr/prepbufr_adpshp_api.py ](https://github.com/JCSDA-internal/ioda-converters/blob/4edd5731a0f80f51b14af5abf5639c9eb53d88f7/src/ncepbufr/prepbufr_sfcshp_api.py#L78) decodes strings as strict UTF-8 inside the C++ pybind11  layer (DataObject.h: py::str(data_[i])), so the UnicodeDecodeError cannot be caught mid-call and re-decoded. The proper fix is to change that line in DataObject.h to use Latin-1 decoding (e.g. py::bytes(data_[i]).attr("decode")("latin-1")) and rebuild.  Until then, fall back to empty strings so the conversion can complete.

## Issue(s) addressed

Resolves #1790 

## Dependencies

None

## Impact

No downstream impact

Run Skylab ingest suite with the  INCLUDE ${JEDI_WORKFLOW}/skylab/obs/ingest/gdas_prepbufr48h_ingest.yaml obs type.

## Checklist


BenjaminRuston:
thanks for the update. @fcvdb did you put in this comment?

Is `DataObject.h` inside IODA?

```
Some BUFR files contain station IDs with non-UTF-8 bytes (e.g. Latin-1 accented
characters). pyiodaconv.bufr decodes strings as strict UTF-8 inside the C++ pybind11
layer (DataObject.h: py::str(data_[i])), so the UnicodeDecodeError cannot be caught
mid-call and re-decoded. The proper fix is to change that line in DataObject.h to use
Latin-1 decoding (e.g. py::bytes(data_[i]).attr("decode")("latin-1")) and rebuild.
Until then, fall back to empty strings so the conversion can complete.
```

BenjaminRuston:
the `DataObject.h` lives here:
https://github.com/JCSDA-internal/ioda-converters/blob/e08244daa1020c02eee1d94eaed2b3d969916d2d/src/bufr/DataObject.h

is this the best way, we should likely open an issue on this tech debt if we do proceed with this merge

BenjaminRuston:
questions on this one... @dantyndall and @fcvdb 

does the unit test file produce this error when trying to get `SID` or do we need another sample file

should we address the change to the C++ [DataObject.h](https://github.com/JCSDA-internal/ioda-converters/blob/e08244daa1020c02eee1d94eaed2b3d969916d2d/src/bufr/DataObject.h) directly in this PR or in a subsequent one

any maybe most importantly what is the priority of this?

BenjaminRuston:
ok I ran both of these for the unit test file in `test/testinput` and do not encounter this error

so may try grabbing some **full** files to see if I can reproduce it will be good to have one and can attach it to an issue about doing a proper fix to the [DataObject.h](https://github.com/JCSDA-internal/ioda-converters/blob/e08244daa1020c02eee1d94eaed2b3d969916d2d/src/bufr/DataObject.h)

BenjaminRuston:
went ahead and fixed the testoutput so this passes.... again will try to get a file so can see if can't encounter this error,, attaching it would be helpful to a subsequent developer to fix this problem back at the source

BenjaminRuston:
tried a few dates and couldn't find this error, guess since they are BUFR we can just insert some `Latin-1 accented
characters` manually which may be easiest.

BenjaminRuston (approved):
this works as expected, however have not been able to adequately test cases with the Latin-1 accented characters

BenjaminRuston:
@haydenlj can you please have a look at this PR,, thanks

haydenlj (approved):
Thanks @fcvdb looks good to me

PR #1792, opened 2026-06-22
Add min error of 0.1 for airnow

huiying-noaa:
Add min error of 0.1 for AirNOW. Please let me know if there's any comments.

Issues: current version has 0 error in our development test for PM2.5 DA for RRFS-SD https://github.com/NOAA-EMC/rrfs-workflow/tree/rrfs-mpas-jedi. The RDASApp used for RRFS DA directly points to this repo. 

Simple PR, no dependency. 

I have performed a self-review of my own code and tested results.
Original PM2.5 error: _, 0.05, 0.47, 0, 0.12
New PM2.5 error: _, 0.1, 0.47, 0.1, 0.12.

PR #1794, opened 2026-06-30
Allow forks to use presubmit CI by using "pull_request_target" workflow.

eap:
Change the CI trigger to use `on: pull_request_target:...` instead of "on: pull_request:..." this means that all pull requests, including pull requests from forks, will run CI using the checked-in config from the target branch.

This change has been made to enable fork pull requests to test. The previous strategy only allowed tests from branches of this repository forks.

__Issues:__

This change unblocks https://github.com/JCSDA-internal/ioda-converters/pull/1792

### DIFF
--- a/.github/workflows/start-jedi-ci.yaml
+++ b/.github/workflows/start-jedi-ci.yaml
@@ -1,7 +1,7 @@
 name: start-jedi-ci
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
      - 'master'
      - 'main'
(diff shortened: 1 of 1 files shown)


.github/workflows/start-jedi-ci.yaml
eap, line 1: I removed "master" and "main" from this workflow since the strategy requires the referenced workflow to exist in the target branch. Since both are deprecated they do not have a copy of the workflow and the workflow will fail.

Our existing branch rules will prevent accidental submission to these branches.

ashley314 (approved):

weiwilliam (approved):
Thank you! @eap

PR #1799, opened 2026-07-08
Bugfix/tempo

mer-a-o:
## Description

This PR fixes a bug in the tempo converter regarding the ak calculation, cleans up the mask, and adds scattering weight (w) to the tempo ioda file. w can later be used for assimilation of Slant Column Density (SCD).


src/compo/tempo_nc2ioda.py
```diff
@@ -224,13 +225,15 @@ def _read(self):
                         ncd.groups['product'].variables['vertical_column_troposphere'][:].ravel()
                         + ncd.groups['product'].variables['vertical_column_stratosphere'][:].ravel()
                     ) * conv
-                    col_amf = tot_amf
```
rajichidamb, line 227: Why is this line removed here? `col_amf` is assigned in the following else statement, so wondering if this will affect the value.

mer-a-o: `col_amf` is now assigned outside of this if/else statement in line 236

rajichidamb (approved):
Thanks @mer-a-o

jeromebarre (approved):

PR #1802, opened 2026-07-28
NSIDC icec converters for NASA team and Bootstrap algorithms

Dooruk:
## Description

This converter handles data from the NSIDC NASA Team and Bootstrap algorithms. It extracts valid observations and applies necessary masking, assigns a constant obs. error (0.1).

Note: These algorithms have different data types and scaling factors. The NASA Team data are in "ubyte" and have a scaling factor of 0.004, while the BT's in "short" format and its scale factor is "0.001". They also use different filling values for land and pole hole. But NetCDF4 package "should" handle this automatically.

More details:
https://nsidc.org/data/user-resources/help-center/descriptions-and-differences-between-nasa-team-and-bootstrap-algorithms

The grid file(s) can be found here:
https://daacdata.apps.nsidc.org/pub/DATASETS/nsidc0771_polarstereo_anc_grid_info/

Tested grid file for Arctic: NSIDC0771_LatLon_PS_N25km_v1.1.nc
Tested grid file for Antarctic: NSIDC0771_LatLon_PS_S25km_v1.1.nc


Dooruk:
FYI @DavidSRussell and @vrx- for an example PR

mjkagnes123:
I tested this converter with actual NSIDC NASA Team and Bootstrap Arctic files for 2024-12-01, (NSIDC0051_SEAICE_PS_N25km_20241201_v2.0.nc and NSIDC0079_SEAICE_PS_N25km_20241201_v4.0.nc) and noticed something in the masking logic that may be worth checking. 

Since the converter is intended to keep only valid data points, I expected the masked land/coast/pole-hole locations to be excluded from the IODA output.  When F17_ICECON is read with netCDF4, the scale factor is already applied, so the valid sea-ice concentrations are returned as values between 0 and 1. netCDF4 also returns the special NSIDC flag locations as masked values (251-254 for NASA Team and 1100 and 1200 for Bootstrap). 

The behavior seems to come from this section: 
land_mask = (cice_2d ==1200)
pole_hole_mask = (cice_2d == 1100)
invalid_mask = land_mask | pole_hole_mask | np.isnan(cice_2d)
valid = ~ invalid_mask

Since cice_2d is already a MaskedArray, these comparisons also produce masked arrays and the existing masked locations are therefore not converted into Boolean invalid locations, and cice_2d[valid] appears to retain ALL grid cells. 

For the 2024-12-01 Arctic files, the NASA Team input has 67,884 valid SIC cells and Bootstrap has 67,904, while the current converter writes 136,192 IODA locations for both. The masked flag values also appear in ObsValue/seaIceFraction, all with ObsError = 0.1  and PreQC = 0.

Would it make sense to use the existing NetCDF mask directly, for example:

cice_data = np.ma.getdata(cice_2d)
invalid_mask = np.ma.getmaskarray(cice_2d)
invalid_mask  |= ~np.isfinite(cice_data)

valid = ~invalid_mask

cice_1d = cice_data[valid].astype(np.float32)
lon_1d = np.ma.getdata(lon_2d)[valid].astype(np.float32)
lat_1d=np.ma.getdata(lat_2d)[valid].astype(np.float32)

With this change, I get 67,884 NASA Team locations and 67,904 Bootstrap locations, and all output seaIceFraction values are within [0,1].   

My input files, original converter outputs, revised converter outputs, and test code are available here for reference:  /discover/nobackup/mkim1/soca/ioda_cov_NASA_BootStrap_Data

BenjaminRuston:
thanks for the PR and breadcrumbs for data access

was able to view and download a couple of files will try and verify on our end, looks like there may be a comment to address first?

Dooruk:
Yes! Converting it to draft until @mjkagnes123's comments are resolved

Dooruk:
I addressed @mjkagnes123's concern (thank you!). The only difference is with this part of the code:

```
        # Convert fraction/percentage to scale [0, 1] if required
        if np.max(cice_1d) > 1.0:
            cice_1d /= 1000.0
```
Rather than to produce an error it does scaling. Other han that it is ready for review.

mjkagnes123 (approved):
Thanks for addressing my masking comment. I tested the revised converter with both NASA Team and Bootstrap products for the Arctic and Antarctic cases. In all four tests, the number of IODA locations now matches the number of valid input cells, and all output sea-ice fractions are within [0,1] with no masked or out-of-range values. 

Everything looks good from my testing. Approving.

Dooruk:
We need more approvals for this I believe?

src/marine/gmao_nsidc_cice2ioda.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Keep only valid data points to reduce memory/disk footprint
+        valid = ~invalid_mask
+
+        cice_1d = cice_data[valid].astype(np.float32)
+        lon_1d = np.ma.getdata(lon_2d)[valid].astype(np.float32)
+        lat_1d = np.ma.getdata(lat_2d)[valid].astype(np.float32)
+
+        # Convert fraction/percentage to scale [0, 1] if required
+        if np.max(cice_1d) > 1.0:
```
rajichidamb: Do we need to also check for the longitude and latitude values - if they are in radians?

PR #1812, opened 2026-08-25
Update the converter script to save values to one variable

### LINKED ISSUE #1806: Caliop leve2 extinction coefficient converter fixes
## Description

The operator requires few changes - 1) Replace the channel for layer 2) maybe remove the `atmosphereLayerThicknessZ` 3) Remove variable `heightVertice`

## Requirements

Passing ctest

rajichidamb:
## Description

caliop_l2_extioda.py script was creating 2 variables of extinctionCoefficient for each wavelength, updating the script to match other variable naming convention. Now one variable `extinctionCoefficient` but for 2 channels will be created and saved with data per ['Location', 'Channel']. Variable `heightVertice` is not read and saved in nc4 file.

## Issue(s) addressed

Resolves https://github.com/JCSDA-internal/ioda-converters/issues/1806

## Dependencies

none this produces output consistent with the current UFO, and adds additional capability for PR4325 listed under impact

## Impact

- [ ] https://github.com/JCSDA-internal/ufo/pull/4325

ctest -R test_iodaconv_caliop_l2_extcoeff

## Checklist


src/compo/caliop_l2_ext2ioda.py
```diff
@@ -47,21 +46,24 @@
...
+# location = layer*n_profiles + profile (0-indexed). "height" is each layer's
+# midpoint altitude; combined with atmosphereLayerThicknessZ (top = height +
+# thickness/2, bottom = height - thickness/2).
+# Channel is the CRTM/wavelength channel (532nm=1, 1064nm=2).
 VarDims = {
-    'extinctionCoefficient_532nm': ['Location', 'Channel'],
-    'extinctionCoefficient_1064nm': ['Location', 'Channel'],
-    'pressure': ['Location', 'Channel'],
-    'height': ['Channel'],
-    'heightVertice': ['Vertice'],
```
jeromebarre, line 55: We need to keep this because `obs_iheight` need to be used directly instead of using `atmosphereLayerThicknessZ` in UFO. 
The profile layer operator is a partial integration and to keep it generic all we need is the bounds of the dimension of the integrated interval. We should require the converters to provide such bounds and not observation type specifics such as `atmosphereLayerThicknessZ`. Please revert this. see https://github.com/JCSDA-internal/ufo/pull/4325/changes#r3863625165

src/compo/caliop_l2_ext2ioda.py
```diff
@@ -47,21 +46,24 @@
...
-    'atmosphereLayerThicknessZ': ['Channel'],
-    'cloudAerosolDiscriminationHigher': ['Location', 'Channel'],
-    'cloudAerosolDiscriminationLower': ['Location', 'Channel'],
+    'extinctionCoefficient': ['Location', 'Channel'],
+    'pressure': ['Location'],
+    'height': ['Location'],
+    'atmosphereLayerThicknessZ': ['Location'],
```
jeromebarre: we don't need this but we need this
`'heightVertice': ['Vertice']`

src/compo/caliop_l2_ext2ioda.py
```diff
@@ -181,16 +170,15 @@ def _read(self):
             thickness[tmpidx] = thickness[tmpidx + 1]
             thickness[tmpidx - 1] = thickness[tmpidx - 1] + np.abs(thickness[tmpidx] - oldthick)
 
-        # Height at interface
```
jeromebarre, line 184: we need to keep that calculation here and not in UFO, see:
https://github.com/JCSDA-internal/ufo/pull/4325/changes#r3863625165

src/compo/caliop_l2_ext2ioda.py
```diff
@@ -238,44 +224,75 @@ def _read(self):
...
+
+        self.outdata[('latitude', metaDataName)] = tile_per_profile(lats_all)
+        self.outdata[('longitude', metaDataName)] = tile_per_profile(lons_all)
+        self.outdata[('dateTime', metaDataName)] = tile_per_profile(time_all)
+        self.outdata[('sequenceNumber', metaDataName)] = tile_per_profile(seq_all)
+        self.outdata[('pressure', metaDataName)] = flatten_profile_layer(pres_all)
+        self.outdata[('cloudAerosolDiscriminationHigher', metaDataName)] = flatten_profile_layer(cad1_all)
+        self.outdata[('cloudAerosolDiscriminationLower', metaDataName)] = flatten_profile_layer(cad2_all)
+        self.outdata[('height', metaDataName)] = repeat_per_layer(np.array(height, dtype=np.float32))
+        self.outdata[('atmosphereLayerThicknessZ', metaDataName)] = repeat_per_layer(np.array(thickness, dtype=np.float32))
```
jeromebarre: we don't need `atmosphereLayerThicknessZ` but we need 
`self.outdata[('heightVertice', metaDataName)] = np.array(iheight, dtype=np.float32)`

jeromebarre (dismissed):
Please change accordingly with: 
https://github.com/JCSDA-internal/ufo/pull/4325#pullrequestreview-5031475361

src/compo/caliop_l2_ext2ioda.py
rajichidamb: https://github.com/JCSDA-internal/ioda/blob/3f64153094bb1153febb4cf8b6068c8d16e0a063/share/ioda/yaml/validation/ObsSpace.yaml does not have `heightVertice` as an approved variable name. So how do we proceed, to add this variable?

jeromebarre: just add it there with a PR
we already have it in pressure
https://github.com/JCSDA-internal/ioda/blob/3f64153094bb1153febb4cf8b6068c8d16e0a063/share/ioda/yaml/validation/ObsSpace.yaml#L562

BenjaminTJohnson:
@rajichidamb @jeromebarre I have an independent CALIOP APro to IODA converter (conservative regrid of the 399-bin extinction profiles onto target height or pressure layers, written for CRTM / operator comparisons) and compared it against `caliop_l2_ext2ioda.py` at this branch head on a full V5 granule (`CAL_LID_L2_05kmAPro-Standard-V5-00.2006-06-12T00-53-43ZN.hdf`, 3728 records). Operator-side points are on ufo #4325; these are converter-side.

**Corroborated**

- Thickness / interface geometry: identical to 0.001 m between the two implementations (179.87 m above the 20.2 km transition, 59.96 m below, sums 30398.5 m). The transition fix-up in this script gives the same result as deriving the two regime spacings directly.
- Time (middle-shot `Profile_UTC_Time`), geolocation (middle shot) and per-bin MERRA-2 `Pressure` handling match.
- Integrating the raw valid extinction bins over these thicknesses reproduces the file's own `Column_Optical_Depth_Tropospheric_Aerosols + Stratospheric` to 3e-5 rms per record, so this bin geometry is exactly what the L2 processor used.

**Suggested changes**

1. Special values leak into ObsValue/ObsError. Only `-9999` is mapped to the IODA fill (`obs == caliop_missing_value`). CALIPSO also uses `-111` (low laser energy), `-333` (failed retrieval), `-444` (improper single-shot cloud clearing) in the extinction fields and `-29` (opaque water cloud) in the uncertainty field; all of these currently pass through as data. This granule happens to contain none of them, so the ctest would not catch it, but they occur in the record.

2. Clear-air semantics. `-9999` means "no feature detected" in most bins (about 94% on this granule), which is a zero-aerosol observation, not a missing one. Mapping it to fill makes the obs set one-sided (aerosol can only be added, never removed by the operator). I would suggest an option to write those bins as zero extinction, using `Atmospheric_Volume_Description` feature type bits 1 to 3 to separate genuinely clear bins (type 1) from cloud (2), surface / subsurface (5, 6) and totally attenuated (7) bins, which are not retrievable for aerosol and should stay fill. Note that bins carrying a retrieved value should be kept regardless of the other half-bin's feature type; excluding them breaks the column-AOD check above.

3. Extinction QC flag choice. `{0, 1, 2, 16, 18}` follows Young et al. (2018); the L3 aerosol profile product (Tackett et al. 2018) filters on `{0, 1, 16, 18}`. On this granule flag 2 covers 6247 retrieved bins and flag 6 another 12095, so the choice is not cosmetic. A command-line option would let users pick. Also, when the two half-bin flags differ, taking the flag from the half that holds the aerosol feature (AVD feature type) is more faithful than `np.maximum`.

4. Layer bounds (the `heightVertice` thread). Per-Location `heightOfTopOfLayer` / `heightOfBottomOfLayer` in MetaData would give the operator exactly the integration bounds it needs without a `Vertice` dimension, without a shared-grid assumption, and without the layer-major ordering the operator currently relies on. It also allows dropping empty bins and writing coarser layers, which is the direct fix for the 399x GeoVaLs cost.

5. If coarser layers are ever written, coverage metadata becomes necessary: fraction of the layer that was retrievable, cloud fraction, attenuated fraction, and a PreQC bitmask (partial coverage, cloud, attenuated, failed bins, low energy, QC flag outside the accepted set, weak CAD, surface intersect). I have this implemented and can contribute it.

6. Minor: `caliop_time2dt` and the per-profile tiling are fine; `AttrData['datetimeRange']` is useful. `sequenceNumber` is reset to start at 0 across files, so it is no longer the granule record index; worth documenting.

I am glad to open a PR against `feature/extinctionCoeff` with items 1 to 5 if you would rather review code than a comment.

_Comparison prepared with Claude Code against the branch heads as of 2026-08-26; the numbers are reproducible checks on the granule named above._

srherbener:
Is this PR in sync with jcsda-internal/ufo/pull/4325? If so, it seems that this is in good shape for approval (or near approval) since we are converging on an acceptable solution in jcsda-internal/ufo/pull/4325. What do others think?

BenjaminRuston (approved):
thanks @rajichidamb this has both backward compatibility and allows work to advance for space lidar

BenjaminTJohnson (approved):

PR #1816, opened 2026-09-14
Add compatibility changes for numpy 2.0, these should work with recen…

eap:
Fix numpy 2.0 compatibility.

Compat changes included here:
- [NEP 50 - scalar propmotion](https://numpy.org/neps/nep-0050-scalar-promotion.html) - When combining numpy types and python scalars, additional type coercion is necessary.
- I could find no link for this, but numpy 2.0+ does not offer `NaN` or `NAN`, luckily `nan` is available in 1.0.

This partially fixes the issues described here, addressing only the numpy 2.0+ issues and not the pandas string problems https://github.com/JCSDA-internal/ioda-converters/issues/1815


srherbener (approved):
I think these changes look good. Thanks for this update!

srherbener:
It would be good for someone on the OBS team to review this. I'm going to add @fcvdb to the reviewer list.

svahl991 (commented):
I'm curious about the testing we have for ioda-converters. Do we have ctests with reference files? If not, I'd be nervous that we might unintentionally be changing the data type being written to file somewhere with one of these changes.

svahl991 (approved):
I dug into the testing in ioda-converters a little and it looks like we do compare everything to a reference file, so I'm hitting Approve. This is not to override @srherbener's desire to have @fcvdb take a look too, though.

climbfuji:
> I dug into the testing in ioda-converters a little and it looks like we do compare everything to a reference file, so I'm hitting Approve. This is not to override @srherbener's desire to have @fcvdb take a look too, though.

In fact, most of the changes in this PR are only because of the comparison to reference files. The ctests failed because the data types changed implicitly due to the numpy nep50 bug.

src/compo/mopitt_co_nc2ioda.py
```diff
@@ -124,10 +124,10 @@ def _read(self):
...
             # convert all concentrations and column to correct units to avoid single precision issues
             u_conv = avogadro / scm2sm
-            xa_gd = xa_gd * vmr2col / u_conv
-            xa_tc = xa_tc / u_conv
-            xr_tc = xr_tc / u_conv
-            er_tc = er_tc / u_conv
+            xa_gd = xa_gd * xa_gd.dtype.type(vmr2col) / xa_gd.dtype.type(u_conv)
+            xa_tc = xa_tc / xa_tc.dtype.type(u_conv)
+            xr_tc = xr_tc / xr_tc.dtype.type(u_conv)
+            er_tc = er_tc / er_tc.dtype.type(u_conv)
```
BenjaminRuston, line 130: @climbfuji , @jeromebarre ,  @mer-a-o this is being updated for numpy 2, 

from my quick glance what was here was to handle the very large/small values now this is updated..

believe though everything out of this converter and down the line in UFO ObsSpace will carry single precision values from whatever is output from here

so we want to add some explicit place where this single precision-ifying takes place

BenjaminRuston:
@eap what is the urgency to merge this?

we definitely could move it along, have quick questions that could be moved to another issue if we would like to proceed

src/compo/mopitt_co_nc2ioda.py
climbfuji: The reason for these changes is explained in https://github.com/JCSDA-internal/ioda-converters/issues/1815, see also @svahl991's question and my reply here.

eap:
@BenjaminRuston - there's no immediate urgency on this, I'm just trying to get ahead of this fairly large task by migrating as much old code as possible so that our final migration for spack-stack 2.2 can be limited to addressing compatibility issues arrising from recent work.

climbfuji (approved):
I reviewed your changes. They are surprisingly similar if not identical to the two patches that I created a few weeks ago with the help of Claude, see https://github.com/JCSDA/spack-stack/blob/develop/repos/spack_stack/spack_repo/spack_stack/packages/ioda_converters/ioda-converters-numpy2-ma-scalar-dtype.patch and https://github.com/JCSDA/spack-stack/blob/develop/repos/spack_stack/spack_repo/spack_stack/packages/ioda_converters/ioda-converters-numpy2-api-and-precision.patch !

PR #1817, opened 2026-09-16
Another update to tempo converter

mer-a-o:
## Description

Writing the scattering weights (`w`, or box AMFs) can significantly increase the output file size. This PR makes writing `w` optional; it is omitted by default and can be enabled with --save_w.

This PR also adds the native TEMPO L2 grid indices (xtrack and mirror_step) to the IODA output as `xtrackIndex` and `mirrorStepIndex`. These indices allow filtered observations to be mapped back onto the native two-dimensional scan grid, preserving pixel adjacency needed by emissions-inversion and superobbing workflows.

## Dependencies
None

## Impact
None

## Checklist

### DIFF
--- a/src/compo/tempo_nc2ioda.py
+++ b/src/compo/tempo_nc2ioda.py
@@ -23,6 +23,12 @@
 #        -c: column type, one of [total, troposphere]
 #        -q: optional maximum QA value to keep before QC, default 0
 #        -t: optional random thinning fraction from 0.0 to 1.0, default 0.0
+#        -w: optional flag to also write the scattering weights (box AMFs) to
+#            RetrievalAncillaryData/w. Off by default.
+#
+#        The native L2 grid indices are always written as MetaData/mirrorStepIndex
+#        and MetaData/xtrackIndex so that observations can be scattered back onto
+#        the (xtrack, mirror_step) lattice for native-grid finite differences.
 
 import argparse
 import netCDF4 as nc
@@ -59,13 +65,15 @@ molarmass = {"no2": 46.0055, "hcho": 30.031, "o3": 48.0}
 
 
 class tempo(object):
-    def __init__(self, filenames, varname, columnType, qa_flg, thin, obsVar):
+    def __init__(self, filenames, varname, columnType, qa_flg, thin, obsVar,
+                 save_w=False):
         self.filenames = filenames
         self.varname = varname
         self.columnType = columnType
         self.qa_flg = qa_flg
         self.thin = thin
         self.obsVar = obsVar
+        self.save_w = save_w
         self.varDict = defaultdict(lambda: defaultdict(dict))
         self.outdata = defaultdict(lambda: DefaultOrderedDict(OrderedDict))
         self.varAttrs = DefaultOrderedDict(lambda: DefaultOrderedDict(dict))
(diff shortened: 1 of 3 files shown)


rajichidamb (approved):

src/compo/tempo_nc2ioda.py
```diff
@@ -465,6 +529,13 @@ def main():
         help="percentage of random thinning from 0.0 to 1.0. Zero indicates"
         " no thinning is performed. (default: %(default)s)",
         type=float, default=0.0)
+    optional.add_argument(
+        '-w', '--save_w',
```
BenjaminRuston: such a small point but I am not a fan of short options `-w`  or `-l` and prefer to only have the long verbose names such as `--save_weights` and leave the short ones for primarily `-i` and `-o` 

that said have no strong opinion to change as see there are a number of **short** options already present here 😺

mer-a-o: Agreed. I removed the short options and only kept them for `-i` and `-o`.

src/compo/tempo_nc2ioda.py
```diff
@@ -367,7 +407,10 @@ def _read(self):
                     self.outdata[('viewingZenithAngle', 'MetaData')] = vza[flg]
                     self.outdata[('albedo', 'MetaData')] = albedo[flg]
                     self.outdata[('averagingKernel', 'RetrievalAncillaryData')] = avg_kernel[flg]
-                    self.outdata[('w', 'RetrievalAncillaryData')] = w[flg]
+                    self.outdata[('mirrorStepIndex', 'MetaData')] = mirror_idx[flg]
```
jeromebarre: This needs to be optional too we don't need this for actual data assimilation.

src/compo/tempo_nc2ioda.py
```diff
@@ -367,7 +407,10 @@ def _read(self):
...
+                    self.outdata[('xtrackIndex', 'MetaData')] = xtrack_idx[flg]
```
jeromebarre: This needs to be optional too we don't need this for actual data assimilation.

src/compo/tempo_nc2ioda.py
```diff
@@ -266,6 +274,28 @@ def _read(self):
...
+            # granules are concatenated into one IODA file.
+            if 'mirror_step' in ncd.variables:
+                ms_vals = np.asarray(ncd.variables['mirror_step'][:], dtype='int32')
+            else:
+                ms_vals = np.arange(mirror, dtype='int32')
+            if 'xtrack' in ncd.variables:
+                xt_vals = np.asarray(ncd.variables['xtrack'][:], dtype='int32')
+            else:
+                xt_vals = np.arange(xtrack, dtype='int32')
+            mirror_idx = np.ma.array(np.repeat(ms_vals, xtrack), mask=mask)
```
jeromebarre: if 'mirror_step' in ncd.variables or if 'xtrack' in ncd.variables is false the converter will break, no?

src/compo/tempo_nc2ioda.py
```diff
@@ -310,7 +340,10 @@ def _read(self):
             print('err: ', np.shape(err))
             print('preslev: ', np.shape(preslev))
             print('avg_kernel: ', np.shape(avg_kernel))
-            print('w: ', np.shape(w))
+            print('mirror_step: ', np.shape(mirror_idx))
```
jeromebarre: make mirror step and xtrack optional. they are not something we need for JEDI data assimilation

jeromebarre:
As a general comment I would put  IODA output as xtrackIndex and mirrorStepIndex as optional. It is actually not needed in JEDI based inversion and also not needed for superobservation. This somewhat a scope creep from another non-JEDI project that I am willing to close my eyes on if this is made optional.

